PR #16, opened 2020-01-15
11 read from methods

pjcon:
Changes to implement new api XData(seedname) -> XData.from_castep(seedname). 
- Updated documentation
- Working tests
- Extendable to new interfaces
- Interpreter autocomplete

### DIFF
--- a/euphonic/data/bands.py
+++ b/euphonic/data/bands.py
@@ -40,7 +40,7 @@ class BandsData(Data):
         Default units eV
     """
 
-    def __init__(self, seedname, model='CASTEP', path=''):
+    def __init__(self, data, **kwargs):
         """"
         Calls functions to read the correct file(s) and sets BandsData
         attributes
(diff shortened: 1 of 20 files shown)


euphonic/data/bands.py
```diff
@@ -122,6 +125,20 @@ def _get_data(self, seedname, model, path):
...
+                try:
+                    self.seedname = data['seedname']
+                    self.model = data['model']
+                    self.path = data['path']
+                except: #TODO warn
+                    self.seedname = None
+                    self.model = None
+                    self.path = ''
+        except: #TODO warn
+            pass
```
rebeccafair: Could we change this part to:

```
try:
    self.model = data['model']
    if data['model'].lower() == 'castep':
        self.seedname = data['seedname']
        self.model = data['model']
        self.path = data['path']
except KeyError:
    pass
```

So we'll except on a KeyError and just carry on if the model key isn't there, and for now we'll just assume that if model == 'CASTEP' the seedname, model and path keys are also there, as they should be set by the `_castep._read_interpolation_data` function.

And the same thing for the `_get_data` function in `InterpolationData` and `PhononData`

euphonic/data/phonon.py
```diff
@@ -143,19 +143,14 @@ def _set_data(self, data):
...
-                    self.model = data['model']
-                    self.path = data['path']
-                except: #TODO warn
-                    self.seedname = None
-                    self.model = None
-                    self.path = ''
-        except: #TODO warn
+                self.seedname = data['seedname']
+                self.model = data['model']
+            self.path = data['path']
```
rebeccafair: `self.path` indentation?

PR #34, opened 2020-03-23
95 unit test remove false positives

JamesKingWork:
Ensure unit test failure on not using c extension

As a maintainer of Euphonic, I would like the C extension to be tested in my python unit tests.

Acceptance criteria:

- When use_c is True and the c extension is not used in the unit tests then the test fails

Notes:

- I have added a fall_back_on_python keyword argument that specifies whether or not the user wishes to accept not using the C extension if not possible (this defaults to True)

### DIFF
--- a/test/test_interpolation_data.py
+++ b/test/test_interpolation_data.py
@@ -348,17 +348,13 @@ class TestInterpolatePhononsLZO(unittest.TestCase):
                             atol=1e-10)
 
     def test_calculate_fine_phonons_no_asr_c(self):
-        with warnings.catch_warnings(record=True) as warning_messages:
-            warnings.simplefilter("always")
-            self.data.calculate_fine_phonons(self.qpts, use_c=True)
-            for warning_message in warning_messages:
-                self.assertTrue("use_c=True" not in str(warning_message.message))
-            npt.assert_allclose(self.data.freqs.to('hartree').magnitude,
-                                self.expctd_freqs_no_asr.to('hartree').magnitude,
-                                atol=1e-10)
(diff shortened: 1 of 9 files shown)


test/test_interpolation_data.py
```diff
@@ -346,14 +347,16 @@ def test_calculate_fine_phonons_no_asr(self):
                             self.expctd_freqs_no_asr.to('hartree').magnitude,
                             atol=1e-10)
 
-    def test_calculate_fine_phonons_no_asr_c(self):
-        self.data.calculate_fine_phonons(self.qpts, use_c=True)
+    @parameterized.expand([(True, ), (False, )])
+    def test_calculate_fine_phonons_no_asr_c(self, test_c: bool):
+        self.data.calculate_fine_phonons(self.qpts, use_c=test_c, fall_back_on_python=not test_c)
```
rebeccafair: Does `@parameterized.expand([(True, ), (False, )])` in this case mean it will test with both `use_c=True` and `False`? A few lines above we have `test_calculate_fine_phonons_no_asr` which tests with `use_c=True` so we're duplicating tests.

Maybe just use `@parameterized.expand([(True, )])` for now and we can combine the `<test>`, `<test>_c` and `<test>_c_2threads` into one test in another issue?

The same goes for all the other `@parameterized`

JamesKingWork: My mistake, I misunderstood the purpose of the other tests! But yes it does mean what you have said. I think for now I will remove parameterized completely, because pytest has another method for parameterisation so, for now, it makes sense not to use it when needed.

PR #37, opened 2020-03-24
91 pylint in ci

JamesKingWork:
As a maintainer of Euphonic, I would like my code to be linted on the CI to ensure any code changes do not contain logical errors and are not contrary to style guidelines.

Acceptance criteria:

- Codebase contains pylint configuration matching Euphonic style guidelines
- Jenkinsfile has "Static Code Analysis" stage where pylint is run against python files in the codebase using the pylint configuration
- Outcome of linting is displayed in an appropriate place on Jenkins and/or GitHub


static_code_analysis/.pylintrc
```diff
@@ -0,0 +1,581 @@
+[MASTER]
```
rebeccafair: Just a small comment, could we move this file into the test directory? I don't like having too many top-level directories

PR #55, opened 2020-04-14
98 script testing

JamesKingWork:
Ticket: https://github.com/pace-neutrons/Pace-Project-Plan/issues/98

## Work completed

Refactored scripts to make them testable. Added unit, integration and regression tests for the 3 scripts. Refactored utils files to make them match the testing structure. Written rationale [here](https://github.com/pace-neutrons/pace-developers/blob/master/euphonic/testing/script_testing.md).

## Story:

As a maintainer of the Euphonic package, I would like the following scripts adequately tested:
- dispersion.py
- dos.py
- optimise_eta.py

## Acceptance criteria:

- Testing strategy is defined and documented
- Testing tools, techniques and processes are put in place to support strategy
- An initial set of tests have been created to meet the strategy
- Relevant further testing is spun out into new tickets

## Notes:

- All testing has been done as part of this ticket, so no new tickets have been created for further testing


scripts/dispersion.py
```diff
@@ -7,33 +7,28 @@
...
-import os
-from euphonic.data.bands import BandsData
-from euphonic.data.phonon import PhononData
+import warnings
+
+from typing import List
+
 from euphonic.plot.dispersion import output_grace, plot_dispersion
 
+from .utils import load_data_from_file, get_args_and_set_up_and_down, matplotlib_save_or_show
```
rebeccafair, line 16: Can we move `.utils` to the main Euphonic package (e.g. `euphonic/script_utils.py`)? I can imagine users moving the scripts out of this directory and then wondering why they don't work

tests_and_analysis/test/script_tests/utils.py
```diff
@@ -0,0 +1,125 @@
...
+     to separate csv files using the filepath_prefix parameter as a prefix and the
+      index of the line and .csv to complete the filename.
+
+    Parameters
+    ----------
+    filepath_prefix : str
+        A prefix to use to write the files e.g. C:/data
+    """
+    for index, line in enumerate(matplotlib.pyplot.gcf().axes[0].lines):
+        np.savetxt(filepath_prefix + str(index) + ".csv", line.get_xydata().T, delimiter=",")
```
rebeccafair, line 125: Is there a reason why each line is written to a separate .csv? I feel like it would be easier to keep track of the data if it was all in the same file, but maybe there's a good reason for doing it this way?

tests_and_analysis/test/script_tests/test_dos.py
```diff
@@ -0,0 +1,120 @@
...
+    def test_when_called_with_w_and_or_b_then_ito_is_called_correctly(self, inject_mocks, w_b):
+        num_of_args = len(w_b)
+        scripts.dos.main([get_phonon_file()] + w_b)
+        assert scripts.dos.ureg.Quantity.ito.call_count == 2-num_of_args, \
+            "ito should not be called for each time -w and -b is specified"
+
+    def test_when_called_then_np_arange_called_correctly(self, inject_mocks):
+        phonon_mock = inject_mocks["phonon_mock"]
+        scripts.dos.main([get_phonon_file()])
+        assert scripts.dos.np.arange.call_count == 1
```
rebeccafair, line 84: This test might be going too deep into implementation details, if the script somehow gets changed so `np.arange` is called for a different reason too, then this test will fail. Unit testing scripts is hard!

scripts/dispersion.py
JamesKingWork: Yes, I have moved it

tests_and_analysis/test/script_tests/utils.py
JamesKingWork: Though this would make sense, we cannot save a 3D array in human readable form using numpy, so it is best to stick with this for now

tests_and_analysis/test/script_tests/test_dos.py
JamesKingWork: This seems to be the case for all the unit tests, we have instead decided to focus on improving the regression testing using various parameters

PR #59, opened 2020-04-29
122 performance benchmarking

JamesKingWork:
https://github.com/pace-neutrons/Pace-Project-Plan/issues/122

Develop scripts to generate code for performance regression testing and test scripts to run performance regression testing.


tests_and_analysis/performance_benchmarking/Jenkinsfile
```diff
@@ -0,0 +1,110 @@
...
+                        )"
+                    """,
+                    returnStdout: true
+                )
+            } else {
+                error("Cannot get commit author in Windows")
+            }
+        }
+    }
+}
```
rebeccafair, line 62: If we're not failing the tests we probably don't need to report the status back to Github either

PR #68, opened 2020-05-20
Performance benchmark storage

JamesKingWork:
Issue: https://github.com/pace-neutrons/Pace-Project-Plan/issues/128

Work completed:

- Script to pull performance benchmark data from Jenkins and store it in the SAN storage
- Modification of visualisation and speedups calculation to easily read from SAN


tests_and_analysis/performance_benchmarking/artifact_handling/copy_benchmark_artifacts.py
```diff
@@ -0,0 +1,228 @@
...
+        # When there is no range parsed copy all the builds
+        return largest_possible_range
+
+
+def write_to_file(directory: str, filename: str,
+                  performance_benchmarks_json: str):
+    # Check access
+    test_file = os.path.join(directory, "test.txt")
+    try:
+        open(test_file, "w+")
```
rebeccafair, line 125: Rather than creating a specific test file it might be better just to try and write the data that we want and let it fail if we don't have permissions - hopefully that should give a helpful error message anyway

tests_and_analysis/performance_benchmarking/utils.py
```diff
@@ -42,3 +42,16 @@ def get_qpts() -> np.ndarray:
...
+    """
+    Returns
+    -------
+    str
+        The server location of the SAN storage.
+    """
+    return (
+        r'\\isis.cclrc.ac.uk\Shares\PACE_Project_Tool_Source'
+        r'\euphonic_performance_benchmarking'
+    )
```
rebeccafair, line 57: In Horace they've tried to keep the SAN storage location out of Github - we should probably do the same. It's convenient to have it hardcoded in but not essential - I have to manually put in my Jenkins user/token anyway so putting in the storage location too isn't too much trouble

tests_and_analysis/performance_benchmarking/artifact_handling/copy_benchmark_artifacts.py
```diff
@@ -0,0 +1,228 @@
...
+        # We have access therefore write
+        # No need to copy if the file already exists
+        if not os.path.exists(directory):
+            os.mkdir(directory)
+            filepath = os.path.join(directory, filename)
+            json.dump(performance_benchmarks_json, open(filepath, "w+"))
+    except FileNotFoundError:
+        print("Cannot access {}. Exiting.".format(directory))
```
rebeccafair, line 133: Also, because checking if the directory exists and creating it doesn't happen until after the test file is attempted to be created, this fails if the directory doesn't exist, even if we have the permissions. But not creating the test file should fix this anyway.

tests_and_analysis/performance_benchmarking/utils.py
JamesKingWork: My bad, removed

tests_and_analysis/performance_benchmarking/artifact_handling/copy_benchmark_artifacts.py
JamesKingWork: Yes, this was what I thought would be logical as well. However, when copying to network locations (at least on Windows) if we do not have permissions to access it the script hangs on os.path.exists and won't complete or fail. Hence the creating of a test file.

JamesKingWork: Not that I'm sure it's the best way around it, but I tried a few different methods and this was the one I found that worked most reliably.

rebeccafair: Do we need to check if the directory exists before creating it? If `os.mkdir` is called and the directory already exists a `FileExistsError` is raised, which we can catch. Or does `os.mkdir` have the potential to hang also?

PR #73, opened 2020-06-02
Spherical averaging tools

ajjackson:
In order to compute coherent S(q,w) of powders, we would like to perform numerical powder averaging.

The convergence properties and optimal approaches to this problem are not fully-explored, so before reaching a final "recommended" implementation, we can implement a range of options and use these to investigate the problem more thoroughly. Ultimately we would like to provide a good compromise between robustness and efficiency with minimal user intervention.

Approaches include regular grids, quasi-random arrays, fully-random sampling, and random sampling with constrained or re-scaled properties.

This pull request implements spherical sampling functions and relevant analysis tools.
- The primary outcome is a range of sampling functions for spherical surfaces, in `euphonic.sampling`
- In addition it we have an interface to streamline the computation of spherically-averaged S(q,w) spectra from force constants data, in `euphonic.powder`
- The script `scripts/show_sampling.py` allows easy visualisation of the sampling schemes
- The StructureFactor class is extended with a `calculate_1d_average()` method that generates an sqw map and then averages the output over Q to obtain a 1D DOS-like spectrum.


ajjackson:
This is now able to sample S(q,w) as well as DOS. The behaviour is reassuringly similar, so we can use this in convergence testing/analysis.

Figure_1

I'm not sure where is best to put the demo scripts I am using for these plots and initial tests. They aren't intended for usage long-term, they are just a development aid.

I'm finding it a little tricky to correctly convert from a StructureFactor object to an appropriately normalled average 1D spectrum, so that should probably be built into the module.

ajjackson:
- I have now implemented a method `StructureFactore.calculate_1d_average()` that can return an averaged 1D spectrum.
- The workflow for generating a sphere of q-points, evaluating structure factors then summing with this method has been wrapped in the `euphonic.powder.sample_sphere_structure_factor()` function. This is mostly for convenience and has a similar interface to `euphonic.powder.sample_sphere_dos()`.
- A script encapsulating these workflows has been put in a pull request for the `pace-developers` repository, as it was agreed this is a good home for development debugging/prototyping/demonstration scripts.
- I'd like to keep the 3D sampling visualisation script in Euphonic if possible; it should be low-maintenance and is likely to enlighten/entertain some users. The installed entry-point name should be made more specific than `show-sampling`, though, so that it is ok to add to PATH. Maybe `euphonic-show-sphere`?

ajjackson:
I think a bit more test coverage is needed but otherwise the PR is achieving its goals.

ajjackson:
I make that 100% code coverage for the new modules from my local run of `coverage`, but I can't find a way to check if Jenkins saw the same thing.

rebeccafair:
> I make that 100% unit test coverage for the new modules from my local run of `coverage`, but I can't find a way to check if Jenkins saw the same thing.

It's a little convoluted, you have to click on 'Code Coverage', then click:
image

Then scroll to the bottom of the next page then click on 'project' on the right even though it doesn't look like you can:
image

Then scroll to the bottom again and click on the '.' on the right which is the euphonic package (it took me a long time to notice it was even there):
image

Unfortunately the plugin doesn't seem to be able to find the source code, so that should probably be looked into at some point. At least you can see the overall coverage of each file...

euphonic/powder.py
```diff
@@ -0,0 +1,279 @@
...
+        energy_bins = _get_default_bins(phonons)
+
+    s = phonons.calculate_structure_factor(
+        scattering_lengths=scattering_lengths, dw=dw)
+
+    return s.calculate_1d_average(energy_bins)
+
+
+def _get_default_bins(phonons: QpointPhononModes,
+                      nbins: int = 1000) -> np.ndarray:
```
rebeccafair, line 208: Should return `Quantity?`

euphonic/powder.py
```diff
@@ -0,0 +1,279 @@
...
+        euphonic.util.get_reference_data() function.
+
+    Returns
+    -------
+    Spectrum1D
+
+    """
+
+    if isinstance(scattering_lengths, str):
+        import euphonic.util
```
rebeccafair, line 179: Is there a reason behind importing here instead of at the top of the file?

euphonic/sampling.py
```diff
@@ -0,0 +1,352 @@
...
+    Returns
+    -------
+    Iterator[Tuple[float, float, float]]
+
+        Sequence of (x, y, z) coordinates (if cartesian=True) or
+        (r, phi, theta) spherical coordinates.
+
+    """
+    if npts < 6:
+        raise ValueError("This sampling scheme has a minimum of 6 points")
```
rebeccafair, line 282: Add `Raises` section to docstring

euphonic/sampling.py
```diff
@@ -0,0 +1,352 @@
+"""Functions for sampling properties in high-dimensional space.
+
+These are implemented in a generic way for application in euphonic.powder
+"""
+
+from itertools import product
+import numpy as np
+from typing import Iterator, Tuple
```
rebeccafair, line 8: PEP8 imports:

```
from itertools import product
from typing import Iterator, Tuple

import numpy as np
```

I know not every Euphonic module sticks to this, but this is a chance to start fixing that...

euphonic/powder.py
```diff
@@ -0,0 +1,279 @@
...
+from euphonic import (Crystal, DebyeWaller, ForceConstants, QpointPhononModes,
+                      Spectrum1D)
+from euphonic import ureg
+from euphonic.util import mp_grid
+
+sampling_choices = {'golden', 'sphere-projected-grid', 'spherical-polar-grid',
+                    'spherical-polar-improved', 'random-sphere'}
+
+
+def sample_sphere_dos(fc: ForceConstants,
```
rebeccafair, line 16: What's the use case for this function? Would this form part of creating a 2D (|Q|, E) dos spectrum by calling this multiple times? Is that something people would want to do?

euphonic/structure_factor.py
```diff
@@ -106,6 +108,41 @@ def __setattr__(self, name, value):
                 ureg(getattr(self, name)).to(value)
         super(StructureFactor, self).__setattr__(name, value)
 
+    def calculate_1d_average(self,
```
rebeccafair, line 111: How would users use this? Again, would they have to call this for different `StructureFactor` objects with sampling at different |Q| to build their own 2D |Q|,E spectrum?

scripts/show_sampling.py
```diff
@@ -0,0 +1,90 @@
+#! /usr/bin/env python3
```
rebeccafair, line 1: This script is very cool! We should probably test it somehow, even if all the testing does is check that the script runs without errors. There are some existing script tests in `tests_and_analysis/test/script_tests`. They're a little complicated, but have good a example of how to mock matplotlib.

tests_and_analysis/test/euphonic_test/test_powder.py
```diff
@@ -0,0 +1,266 @@
...
+
+    if 'jitter' in sampling_kwargs:
+        sampling_kwargs['jitter'] = jitter
+
+    npt.assert_almost_equal(
+        _get_qpts_sphere(npts, sampling=sampling, jitter=jitter),
+        random_qpts_array)
+
+    mocked_sampling_function.assert_called_with(*sampling_args,
+                                                **sampling_kwargs)
```
rebeccafair, line 57: I'm wondering if we should test private functions and implementation details such as testing that the sampling functions have been called, but at the same time just testing the output of functions such as `sample_sphere_dos` wouldn't tell us much about why its failing...

This could provide useful information and has already been written so can be left in, but something to think about

rebeccafair (changes requested):
So if I understand correctly this PR is mostly to provide useful tools to allow us to start investigating powder averaging? Rather than providing a completely polished way to calculate S(|Q|,w)?

euphonic/powder.py
```diff
@@ -0,0 +1,279 @@
...
+        1000 bins (1001 bin edges) from 0 to 1.05 * [max energy]
+
+    Returns
+    -------
+    Spectrum1D
+
+    """
+
+    qpts_cart = _get_qpts_sphere(npts, sampling=sampling, jitter=jitter
+                                 ) * mod_q
```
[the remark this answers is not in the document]
ajjackson: Most of the example code in PEP8 has lone operators with whitespace, e.g. `i = i + 1`.

It does appear to recommend mixing spaced and unspaced operators to give clarity over priority, e.g. `hypot2 = x*x + y*y` (https://www.python.org/dev/peps/pep-0008/#other-recommendations) . The only other example I can find in PEP8 is within slices, again giving a clear priority distinction from the slice operator `:`.

In general I strongly prefer to space operators for readability; `2*x` is quite nice but it is hard to see that `some_var/the_divisor` includes an operation.

Flake8 specifies space before and after *all* operators https://www.flake8rules.com/rules/E225.html
Black also chooses to format this line with spaces around the `*`.

euphonic/powder.py
```diff
@@ -0,0 +1,279 @@
...
+        energy_bins = _get_default_bins(phonons)
+
+    return phonons.calculate_dos(energy_bins)
+
+
+def sample_sphere_structure_factor(
+    fc: ForceConstants,
+    mod_q: float,
+    dw: DebyeWaller = None,
+    temperature: Quantity = 273. * ureg['K'],
```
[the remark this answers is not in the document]
ajjackson: More agreeable in this case, but see previous comment

euphonic/powder.py
```diff
@@ -0,0 +1,279 @@
...
+
+            - 'golden':
+                Fibonnaci-like sampling that steps regularly along one
+                spherical coordinate while making irrational steps in the
+                other
+
+            - 'sphere-projected-grid':
+                Regular 2-D square mesh projected onto
+                sphere. npts will be distributed as evenly as possible
+                (i.e. using twice as many 'longitude' as 'lattitude' lines),
```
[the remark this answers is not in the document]
ajjackson: Looks like this is recommended for entire docstrings if possible. Bit of a pain, but not too much trouble to follow here.

euphonic/powder.py
ajjackson: There's a good story here: I've been running into import loops as I've added more type hints. Mostly this has hit me on the spectrum plot branch, but clearly that developed a habit that sneaked in here. Basically the problem is

- top-level `euphonic.__init__` imports classes from nearly every module
- a lot of these classes import functionality from `euphonic.util`
- `euphonic.util` tries to import classes for its type hints
- but we're already trying to import that class...
- BOOM

Deferring imports from `euphonic.util` breaks this loop. It's pretty ugly though, I think a better approach will be to wrap class names in strings in `euphonic.util` to defer/prevent their import there.
In hindsight this is really a consequence of keeping a miscellaneous `util` module rather than establishing those features within a clearer heirarchy!

In any case I don't think it's still a problem on _this_ branch, so will see about taking it out.

euphonic/sampling.py
ajjackson: Good spot. Should I also add it to the powder methods that call this function?

euphonic/powder.py
ajjackson: Yes, this could be called repeatedly to generate a 2D DOS.

Is there any value in mod(q)-constrained/resolved DOS data? Well, I guess it's more of a general theoretical tool than something for simulating experiments. If you are interested in the general properties of phonon dispersion it's nice to separate this from the neutron-weighted structure factor data. We'll probably use it as part of the work on convergence.

euphonic/structure_factor.py
ajjackson: It's already called by `euphonic.powder.sample_sphere_structure_factor`, and could be used by other workflow-specific sampling functions we develop.

As a fundamental building-block for anyone who wants to play around with obtaining 1D spectra from different q-sampling schemes, it makes sense to provide public access on the StructureFactor object. It is really _the_ transformation from StructureFactor to Spectrum1D, just as `calculate_sqw_map` is the transformation from StructureFactor to Spectrum2D. Perhaps they should have more similar names.

scripts/show_sampling.py
ajjackson: I think we still need to give this a command name and add it to setup.py, actually.

ajjackson: Some of this code is a bit redundant with a private function in `euphonic.powder`, but as I see it the script also functions as a bit of an `example` so it is good that it only uses the public API.

tests_and_analysis/test/euphonic_test/test_powder.py
ajjackson: Yeah, the tests are a bit abstract unfortunately. The trouble is that these objects interact with nearly every class in Euphonic, so without extensive mocking the test will fail whenever something goes wrong anywhere else in Euphonic.

We should have some kind of integration/regression test on some actual numbers as well, to catch numerical changes in future. But developing such a test wouldn't really have told me whether the module is working as it is supposed to!

ajjackson:
> So if I understand correctly this PR is mostly to provide useful tools to allow us to start investigating powder averaging? Rather than providing a completely polished way to calculate S(|Q|,w)?

This PR provides tools for spherical averaging; you can now get a 1D S(|q|, w) for a particular |q| with `euphonic.powder.sample_sphere_structure_factor()`. This is a starting-point for further sampling work, including various approaches to 2D S(|q|,w) spectra.

euphonic/powder.py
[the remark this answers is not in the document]
rebeccafair: Fair enough, I guess it's a matter of opinion! I've at least generally not put spaces around *,/ in the rest of Euphonic

euphonic/powder.py
rebeccafair: Yeah I've had trouble in the past with this... I was also thinking about separating out util at some point because its just starting to become a very long list of functions. Maybe `__init__` could do with a rethink too.

euphonic/sampling.py
rebeccafair: That's a good question. I would say probably not as it should be fairly uncommon. Also in this case if we add raises documentation further up the stack that might set a precedent for following other exceptions through elsewhere...

euphonic/structure_factor.py
rebeccafair: `calculate_sw_spectrum` would probably be a similar name, but without the formatting of S(ω) it looks pretty strange...

scripts/show_sampling.py
rebeccafair: Good points, I like `euphonic_sphere_sampling`, we should write it with underscores instead of hyphens though in case I want to wrap/import any scripts later (e.g. to include .bands file reading in another project). I should also change the name of the existing scripts, to `euphonic_dispersion` etc. for consistency, I'll make an issue for that.

It would be good to have something to indicate its only for visualisation rather than actually calculating anything, but I'm not sure we can fit another verb in there (`euphonic_view_sphere_sampling` is pretty long).

rebeccafair: Actually, nevermind, thinking about it more I might just refactor the scripts instead, and for scripts hyphens do look nicer and follows everyone else's conventions. Lets not use underscores.

ajjackson: Ok, extensive regression tests are now in place for the script! Naming/entry-point changes to be deferred to a separate PR.

tests_and_analysis/test/euphonic_test/test_powder.py
ajjackson: Marking as "resolved" as there is no action required

euphonic/structure_factor.py
ajjackson: I guess it could be something like `calculate_energy_spectrum` but I don't really find that less confusing!

euphonic/powder.py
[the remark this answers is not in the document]
ajjackson: Will mark as resolved for now as no further action needed. I suspect we'll come back to it one day, though!

euphonic/structure_factor.py
rebeccafair: True, I'd be happy to leave this for now, we can always deprecate and rename in the future, maybe it'll become clearer what it should be called as we actually start to use it and add more powdery things

euphonic/powder.py
ajjackson: Issue opened #90

euphonic/structure_factor.py
ajjackson: I suppose it could be given a "private" name if we don't want to commit this name to the public API yet?

rebeccafair: It's probably ok to leave it for now, if we don't make it public now we might never get round to it!

PR #79, opened 2020-07-01
Reference data system

ajjackson:
An initial implementation of a reference data system:

- Reference data is sorted by label and property
  - This allows trusted collections of multiple properties to be
    accessed systematically
  - Currently the labels and properties are related to files in a
    hard-coded dict stored inside the lookup function
  - It would be good to move this lookup table to somewhere more
    convenient for editing/updating/reference
  - It would also be nice to provide a facility for direct access to
    files outside the Euphonic package (e.g. a 'file' argument that overrides the 'label' and 'prop' options)

I'm not especially happy about abbreviating `property` to `prop`, but it seems dangerous to conflict with the Python inbuilt `property()` function. "attribute" is hardly better. Any suggestions for short words that mean "physical property" and are not ambiguous with programming elements?

- A minor JSON extension is implemented to include complex numbers and
  unit metadata.
  - I have created two data files in this format from reference data. It's not clear if the cross-section data will be immediately useful, but it gives something to work with while establishing a flexible framework.


ajjackson:
On other projects I have used the inbuilt `pkg_resources` module for managing packaged data files. I understand that the new(ish) `importlib.resources` system is supposed to be cleaner and offer better performance. As this was introduced in Python 3.7 and we support 3.6, I have used the `importlib_resources` package. This is a lightweight dependency that backports importlib features from Python 3.9 to older versions (including 3.6).

ajjackson:
One option would be to eliminate the hard-coded table entirely and build it on-the-fly by reading metadata from the files. This would completely eliminate the possibility of them de-syncing in the code source, at a marginally increased risk of funny things happening at run time. When/where should that happen if so? During `import euphonic.util`? `import euphonic`?

ajjackson:
That's interesting, it looks like the DOS tests are failing by a numerical difference of about 1e-17. Shouldn't they have a slightly higher tolerance? I can't imagine how this would be related to the PR changes.

mducle:
> Any suggestions for short words that mean "physical property" and are not ambiguous with programming elements?

Since the method is `get_reference_data` maybe you can use `datum` instead of `prop`? Or `field_name` or `data_type`?

ajjackson:
I would expect `data_type` to be _float_, _complex_ or _pint.Quantity_ . `field_name` would make a lot of sense if the data was stored in tables with multiple properties. At the moment it is not, but that could change.

rebeccafair:
> I'm not especially happy about abbreviating `property` to `prop`, but it seems dangerous to conflict with the Python inbuilt `property()` function. "attribute" is hardly better. Any suggestions for short words that mean "physical property" and are not ambiguous with programming elements?

Would just literally using `physical_property` be too long?

rebeccafair:
How do we intend users to use this? Do they call `get_reference_data` directly, then pass the result to the function e.g. `calculate_structure_factor`? Or do we want a shortcut e.g. `calculate_structure_factor(scattering_lengths='Sears1992')`.

I'm leaning towards the former as then we won't have to change the `calculate_structure_factor` API, and users can manually inspect/edit the returned dictionary if they want to. And in that case we'll have to write the docstring describing which labels are valid for which properties, and maybe add an example usage to the user .rst docs.

ajjackson:
From @mducle elsewhere

> Hmmm... if "label" is a source or collection, shouldn't it be source or collection instead?
>
> What about the database-ish term entity for "prop"?

Yes, I like "source" or "collection" much more! :smile:  "Entity" sounds pretty badass but it doesn't really imply "dict of physical data" to me. "Collection" might be a little bit weird when only one property is available, but the implication that it has come from a wider dataset in that case seems fairly accurate.

@rebeccafair 
> Would just literally using physical_property be too long?

Now that you've written it and I'm looking at it... no, not really. :facepalm: 

Is there a reasonable case where we would like to have _unphysical_ property data saved in the same system?

ajjackson:
Re: API I'd like to have type dispatch so if you pass a string it looks up the data and if you pass a dict it uses the dict you gave it. Then as you say customisation is available by grabbing a dict and tweaking before passing it, but we can also have good defaults that make people's lives easier.

Would be nice to also accept a file path so people can keep a local dataset without having to smuggle it into the Euphonic package.

ajjackson:
Hmm, there is an API problem with accepting filenames actually: it makes the `physical_property` argument meaningless. Perhaps it is better not to provide that option; it's only a couple of clean python lines to load a dict from JSON anyway.

rebeccafair:
> Re: API I'd like to have type dispatch so if you pass a string it looks up the data and if you pass a dict it uses the dict you gave it.

That's fair, I guess as long as it still accepts a dict it would just be adding to the API rather than changing it.

> Hmm, there is an API problem with accepting `collection` filenames actually: it makes the `physical_property` argument meaningless. Perhaps it is better not to provide that option at the `get_reference_data` level; it's only a couple of clean python lines to load a dict from JSON anyway, and that can be passed at the higher level (i.e. when calling `calculate_structure_factor`).

Having both `collection` and `physical_property` could be useful in the case where you have multiple physical properties from the same source (e.g. I think the Sears 1992 also has cross sections and incoherent scattering?). Then maybe the structure of the .json files could be something like:

```
"description": "desc",
"reference": "ref",
"physical_property": [
    "coherent_scattering_length": {"__units__": "fm", "H": -3.739...},
    "incoherent_scattering_length": {"__units__": "fm", "H": null...}
]
```

And inside `calculate_structure_factor` we would call `get_reference_data(collection=filename, property='coherent_scattering_length')` and users would just have to use the same .json format for their files. This would also avoid accidentally using incoherent scattering lengths in a coherent algorithm, for example (unless there's some strange reason people would want to do that?).

ajjackson:
Right, it would work if we store multiple properties in the same JSON file. Which the current implementation does not, but it could.

The downside is that this approach involves loading a bunch of redundant data when the file is read. But it's going to be a trivial amount of time/memory as long as we don't keep re-doing it. (Perhaps we should memoize the loader function?)

> This would also avoid accidentally using incoherent scattering lengths in a coherent algorithm, for example 

Right, this is the whole reason I wanted to separate the collection and property options: we give the calling function chance to exert some control over what is being looked up.

rebeccafair:
> The downside is that this approach involves loading a bunch of redundant data when the file is read. But it's going to be a trivial amount of time/memory as long as we don't keep re-doing it. (Perhaps we should memoize the loader function?)

That's true! There are ways around it, as you said memoizing, or obtaining the dictionary outside of `calculate_structure_factor` then passing it in. Maybe we shouldn't worry about this too much for now, just implement it (and maybe do some ballpark profiling?) and if it is/becomes a problem we can memoize it.

ajjackson:
I think we can do it easily with this decorator: https://docs.python.org/3/library/functools.html#functools.lru_cache

Haven't tried it yet though, last time I looked into this I needed to support Python 2 as well!

ajjackson:
If the user is crazy/malicious/unfortunate enough to name a custom data file in the current directory with one of the inbuilt collection names (and without a .json extension!), which should take priority?

i.e.
```python
pathlib.Path('BlueBook').touch()
dat = euphonic.util.get_reference_data(collection='BlueBook',
                                       physical_property='coherent_cross_section')
```

I'm inclined to have the reference data take priority, because the user can workaround by instead passing '/abs/path/to/BlueBook' or './BlueBook', which will fail the comparison to accepted names.

ajjackson:
Ok, the convenient access from `calculate_structure_factor` is in place. I see that the tests for this method are currently in *legacy_tests` and compare to a reference file. Any thoughts on how/where the new feature should be tested?

I'd be inclined to add tests that checks if the results agree between ref data and a direct weights dictionary, without any comparison to reference results (as this is already covered.)

rebeccafair:
Great! That sounds like a sensible way of testing it, you could add a test user data file in the `test/data` directory, and to minimise the amount of code you have to write for the tests, for now it might be best to just tack a test onto `legacy_tests/test_structure_factor.py` and I'll include it as a test case with the other structure factor tests once I've got around to refactoring them..

ajjackson:
I found some useful tools in the new tests for structure factor data, so popped something in there. No need to distribute a reference file, it's small enough to dump out JSON from a dict on-the-fly.

rebeccafair:
Looks good to me! I think now all that needs to be done is to update the docs to include the new functionality (probably somewhere [here](https://github.com/pace-neutrons/Euphonic/blob/6d308d1fecd09af50e88e2a48a3cf9f80d580070/doc/source/qpoint-phonon-modes.rst#calculating-the-coherent-neutron-structure-factor)) and add a note to `CHANGELOG.rst`

ajjackson:
Done! I also introduced the Sphinx type hints plugin.

It builds ok for me but I had to locally modify *disp/dos/eta-script.rst* to get the docs to build before I even changed anything; the paths to the CLI programs don't seem to be right? But if RTD is happy... :shrug:

euphonic/util.py
```diff
@@ -130,6 +138,101 @@ def get_qpoint_labels(crystal, qpts):
...
+                                             "Li": {"__complex__": true,
+                                                    "real": Li_property1_real,
+                                                    "imag": Li_property1_imag},
+                                             "Nh": None,
+                                             ...},
+                               "property2": ...}}
+
+    Parameters
+    ----------
+    collection:
```
rebeccafair, line 167: Sort of nitpicking, but but you don't need the colon if you've got type annotations (https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_numpy.html). The same goes for `physical_property`

doc/source/euphonic.util.rst
```diff
@@ -1,5 +1,22 @@
+Utilities
```
rebeccafair, line 1: It would be nice to link to this page directly from `tutorials.rst` too

doc/source/qpoint-phonon-modes.rst
```diff
@@ -71,14 +71,29 @@ Calculating The Coherent Neutron Structure Factor
...
+with the *coherent_scattering_length* property. (See :ref:`Reference Data <ref_data>` for details.)
+By default, :py:meth:`QpointPhononModes.calculate_structure_factor <euphonic.qpoint_phonon_modes.QpointPhononModes.calculate_structure_factor>` will use the ``"Sears1992"`` data set included in Euphonic.
+If you have a custom data file, this can be used instead.
+
+Debye-Waller factor
+^^^^^^^^^^^^^^^^^^^
+Inclusion of the
 Debye-Waller factor is optional, and can be provided in the ``dw`` keyword
 argument, see `Calculating The Debye-Waller Exponent`_. From this function a
 :ref:`StructureFactor<structure-factor>` object is returned.
```
rebeccafair, line 93: Because the sentence 'From this function a :ref:`StructureFactor<structure-factor>` object is returned' is now in the Debye-Waller subsection, it's a bit confusing. Can it be removed and put in the above section? (I'll add a comment to indicate)

doc/source/qpoint-phonon-modes.rst
```diff
@@ -71,14 +71,29 @@ Calculating The Coherent Neutron Structure Factor
 
 The neutron structure factor can be calculated for each branch and q-point
 using :py:meth:`QpointPhononModes.calculate_structure_factor <euphonic.qpoint_phonon_modes.QpointPhononModes.calculate_structure_factor>`
```
rebeccafair, line 73: Add 'From this function a :ref:`StructureFactor<structure-factor>` object is returned' here. Or maybe that's just superfluous and a user could figure that out from the docstring anyway...

euphonic/qpoint_phonon_modes.py
```diff
@@ -169,19 +172,37 @@ def reorder_frequencies(self, reorder_gamma=True):
...
         each q-point
 
         Parameters
         ----------
-        scattering_lengths : dictionary of float Quantity
-            Dictionary of spin and isotope averaged coherent scattering
-            length for each element in the structure in e.g.
-            {'O': 5.803*ureg('fm'), 'Zn': 5.680*ureg('fm')}
+        scattering_lengths : Dataset of coherent scattering length for each
+            element in the structure. This may be provided in 3 ways:
```
rebeccafair, line 186: The formatting of this line causes Sphinx to display it weirdly (see image). Since we've got type hints now maybe just put something like

```
scattering lengths
    This is a dataset of coherent scattering lengths for each element...
```

Then for consistency we'd also have to change the `dw` parameter to match. Perhaps I need to think about putting in type hints everywhere at some point...

image

ajjackson:
Cheers for review, I'll get on those points. A general observation about the sphinx / type-hinting interaction: it seems to broadly play well with Napoleon and picks up the argument types correctly. However, it is still required to give the return type explicitly (and redundantly) in the docstring because this is a hard requirement of Numpy-style docstrings. Annoyingly, this then displays without the same automatic formatting as the argument types.

rebeccafair:
> Done! I also introduced the Sphinx type hints plugin.
> 
> It builds ok for me but I had to locally modify _disp/dos/eta-script.rst_ to get the docs to build before I even changed anything; the paths to the CLI programs don't seem to be right? But if RTD is happy... 🤷

Strange! If I run `sphinx-build -b html . test_html` from the `doc/source` directory it works fine for me.

I'm having some trouble with the `sphinx-autodoc-typehints` plugin on readthedocs... We'll be moving away from RTD in the nearish future, but at the moment this would have to be fixed before I can merge (building the docs isn't actually part of the testing suite but maybe it should be). I can give you access to the RTD project so you can have a look, at the moment it looks sort of similar to this issue: https://github.com/readthedocs/readthedocs.org/issues/3734

ajjackson:
> If I run `sphinx-build -b html . test_html` from the `doc/source` directory it works fine for me.

Ah yes, that seems to work for me too. I was trying to use the Makefile in the *doc* directory rather than call sphinx directly. Is that Makefile actually doing anything useful then?

rebeccafair:
You're right, I don't think the Makefile is actually doing anything, it was autogenerated as part of `sphinx-quickstart` and I haven't touched it since, and I don't think readthedocs uses it either. We could get rid of it, but actually updating the docs to work with the Makefile would probably be better. But RTD seems to call sphinx-build directly so maybe we should leave that until we migrate to Github pages.

ajjackson:
That seems to be building succesfully now!

Issue #84, opened 2020-08-03
calculate_sqw_map doesn't handle discontinuities elegantly


ajjackson:
I was trying to plot some neutron-weighted band structures using euphonic. My script is here: https://gist.github.com/ajjackson/21362eb4db1b7dcdb91371c4b6fab9d7

The results look quite good until we hit the non-continuous "extra" branches at the right-hand side. With `calculate_sqw_map` labelling the special points, they get spread out:

![defaults](https://user-images.githubusercontent.com/2511383/89201906-18799800-d5aa-11ea-983a-730b8e7214bb.png)

With some custom code to override the labelling, we can bring the _labels_ together...

![custom_pts](https://user-images.githubusercontent.com/2511383/89201934-23342d00-d5aa-11ea-9f74-eb4689601f1a.png)

but on the plot we still have horizontal artefacts. The discontinuity interacts with a binning/pixellation scheme to gives an overly generous amount of x-axis space to these points. For reference, the phonon band structure plot: http://phonondb.mtl.kyoto-u.ac.jp/ph20180417/d000/mp-661.html

So, the question is what next? The existing setup makes a lot of sense for extended-zone-scheme cuts through the structure. Perhaps it should not change. But these Brillouin-zone plots are very useful. Perhaps a different `calculate_sqw_map` mode/implementation should return multiple Spectrum2D objects that have been separated at the appropriate points?

rebeccafair:
This is a fair point, I'd always intended to handle discontinuities similarly to in the bandstructures by putting them on a separate subplot, but never got round to it.

I do like the idea of `calculate_sqw_map` returning multiple `Spectrum2D` objects on discontinuities, maybe we can set a default 'break tolerance' argument like in `plot_dispersion` (if users still want everything in one object they can set a high tolerance), and then allow `plot_2d` to take a list of `Spectrum2D` objects.

ajjackson:
Detecting discontinuities is an operation that could always be a bit fragile. Given that the user already knows where the discontinuities are, maybe we should just require them to use that information!

The underlying trouble here is that conventional band structure plots actually have k-point data that is 3 levels deep: 
```python
[[[k1a, k1b, k1c], [k2a, k2b, k2c], ...], # branch 1
 [[k11a, k11b, k11c], [k12a, k12b, k12c]], # branch 2
 ...]
```

while our objects in Euphonic work with a 2-D array of k-points without branch divisions.

So the options I see are:
[...]

PR #85, opened 2020-08-04
Spectrum2D: rework plotting, provide a new user script

ajjackson:
Addressing Issue #84

- Refactor plot_2d to use NonUniformImage. This cleans up a lot of the plotting logic, taking care of the irregular spacing for us.
- Break actual axis plotting into separate function that can be re-used for custom/complex layouts, while plot_2d remains a "fire and forget" tool returning a Figure
- Allow Normalize object to be passed, in preparation for implementation of multi-part plot with common colour scale
- Dump script used for testing/prototyping into scripts directory; with a bit more work this should become a good tool for users.

Setting the plot aspect ratio is no longer a technical requirement as NonUniformImage manages this well. For now I will remove the feature; user-facing scripts would be able to set dimensions by manipulating the figure directly.

## To do
- [x] Agree decent names for everything
- [x] Accept a sequence of Spectrum2D in `plot_2d()`, to subplot into a nice band structure
- [x] Improve script to use band structure path and deliver data to `plot_2d()` in segments
- [x] Improve script to work with multiple input file types
- [x] Use Seekpath instead of ASE

## Future wishlist
- Allow `plot_2d` to use heuristics and break up a Spectrum2D automatically. (This should be optional, or at least bypassable.)
- Revisit name/structure/conventions of 1D plotting tools, get things consistent

### DIFF
--- a/euphonic/plot.py
+++ b/euphonic/plot.py
@@ -1,4 +1,4 @@
-from typing import List, Optional, Tuple, Union
+from typing import List, Optional, Sequence, Tuple, Union
 import warnings
 
 try:
@@ -187,7 +187,7 @@ def _plot_2d_core(spectrum: Spectrum2D, ax: Axes,
     return image
 
 
-def plot_2d(spectrum: Spectrum2D,
+def plot_2d(spectra: Union[Spectrum2D, Sequence[Spectrum2D]],
             vmin: Optional[float] = None,
             vmax: Optional[float] = None,
             cmap: Union[str, mpl.colors.Colormap] = 'viridis',
@@ -197,8 +197,11 @@ def plot_2d(spectrum: Spectrum2D,
 
     Parameters
     ----------
-    spectrum
-        Containing the 2D data to plot
+    spectra
+        Containing the 2D data to plot. If a sequence of Spectrum2D is given,
+        they will be plotted from right-to-left as separate subplots. This is
+        recommended for band structure/dispersion plots with discontinuous
+        regions.
     vmin
         Minimum of data range for colormap. See Matplotlib imshow docs
     vmax
@@ -218,13 +221,47 @@ def plot_2d(spectrum: Spectrum2D,
         The Figure instance
     """
 
-    fig, ax = plt.subplots(1, 1)
+    # Wrap a bare spectrum in list so treatment is consistent with sequences
+    if isinstance(spectra, Spectrum2D):
+        spectra = [spectra]
+
+    x_unit = spectra[0].x_data.units
+
+    def _get_q_range(data: Quantity) -> float:
+        dimensionless_data = data.to(x_unit).magnitude
+        return dimensionless_data[-1] - dimensionless_data[0]
+    widths = [_get_q_range(spectrum.x_data) for spectrum in spectra]
+
+    fig, axes = plt.subplots(ncols=len(spectra), nrows=1,
+                             gridspec_kw={'width_ratios': widths},
+                             sharey=True)
+
+    if not isinstance(axes, np.ndarray):  # Ensure axes are always iterable
+        axes = [axes]
+
+    intensity_unit = spectra[0].z_data.units
+
+    def _get_minmax_intensity(spectrum: Spectrum2D) -> float:
+        dimensionless_data = spectrum.z_data.to(intensity_unit).magnitude
+        assert isinstance(dimensionless_data, np.ndarray)
+        return np.min(dimensionless_data), np.max(dimensionless_data)
+    min_z_list, max_z_list = zip(*map(_get_minmax_intensity, spectra))
+    if vmin is None:
+        vmin = min(min_z_list)
+    if vmax is None:
+        vmax = max(max_z_list)
 
     norm = Normalize(vmin=vmin, vmax=vmax)
-    _plot_2d_core(spectrum, ax, cmap=cmap, norm=norm)
 
+    for spectrum, ax in zip(spectra, axes):
+        _plot_2d_core(spectrum, ax, cmap=cmap, norm=norm)
+
+    # Add an invisible large axis for common labels
+    ax = fig.add_subplot(111, frameon=False)
+    ax.tick_params(labelcolor="none", bottom=False, left=False)
     ax.set_xlabel(x_label)
     ax.set_ylabel(y_label)
+
     fig.suptitle(title)
     fig.tight_layout()
 
--- a/euphonic/plot.py
+++ b/euphonic/plot.py
@@ -1,4 +1,4 @@
-from typing import Optional, Union
+from typing import List, Optional, Tuple, Union
 import warnings
 
 try:
@@ -9,7 +9,6 @@ try:
     from matplotlib.colors import Normalize
     from matplotlib.image import NonUniformImage
 
-
 except ImportError:
     warnings.warn((
         'Cannot import Matplotliib for plotting (maybe Matplotlib is '
@@ -17,6 +16,7 @@ except ImportError:
         'dependency, try:\n\npip install euphonic[matplotlib]\n'))
     raise
 
+from pint import Quantity
 import numpy as np
 
 from euphonic import ureg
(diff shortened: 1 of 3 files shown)


ajjackson:
Script call: `./neutron-band-structure.py ../../euphonic-sampling/yaml-data/mp-2319-20180417.yaml --seekpath-labels --x-label 'Seekpath' --cmap pink`

Output:
pink_OsP2

ajjackson:
And here it is on a very fine grid with some broadening. Rather attractive!

pink_OsP2_broadened

ajjackson:
I'm sure Pint has pretty-printing capabilities built-in, anyone know how to get e.g. `millielectron_volt` to show as `meV`?

rebeccafair:
Looks very nice so far!

> I'm sure Pint has pretty-printing capabilities built-in, anyone know how to get e.g. `millielectron_volt` to show as `meV`?

You can do:

```
x = 1*ureg('meV')
print(f'{x.units:~P}')
```

rebeccafair:
I just thought, it might be nice if the script could do a plain (non-neutron weighted) dispersion too

ajjackson:
> `f'{x.units:~P}'`

That works perfectly, thanks!

> I just thought, it might be nice if the script could do a plain (non-neutron weighted) dispersion too

As a normal line plot? Or still as a spectrogram? It would be a good tool for demonstrating the effect of scattering cross-sections.

rebeccafair:
Maybe just as a normal line plot using the existing `plot_dispersion`? Unless you know of something that would be more useful?

ajjackson:
I guess there is some value in both options: spectrogram (i.e. q-dependent phonon DOS) is more meaningful for a busy spectrum with many branches, while the line plot is clearer for sparse data.

To implement a q-dependent DOS I'd basically have to reimplement `calculate_sqw_map`, operating on a QpointPhononModes. (Might as well implement that as a method really.)

To implement 1D plotting from the same script, I think the cleanest thing would be to add the multi-spectrum input option to _plot_1d()_. (i.e. give it a similar interface to the new _plot_2d()_.) Then only a few lines of logic would be needed to bypass the structure factor calculation and call the 1d plotter.

ajjackson:
Right, this work has hit a bit of a sidetrack in visiting the 1D stuff. I think the best way forwards will be to agree a general layout for the 1D work, use that to close out the main "To Do" list of this issue (i.e. name things) and then pursue 1D stuff and 2D splitting in a subsequent PR.

Would people be happy to take this into Master before the 1D stuff is done, or should we create some kind of longer-term feature branch?

rebeccafair:
I think a longer-term feature branch might be useful in this case, on the off chance that we decide to change the 2D plotting inferface based on how the 1D plotting works out - better to change it once than twice!

Another thought, is #86 blocking this at all? Would it be better to implement that first so you can use the public functions without having to change it later? To do this I'd probably also need to update the Spectrum1D/2D tests to the new testing framework, at the moment they're very thin.

ajjackson:
It's possible to do the work in either order, but there's certainly interaction there and it would be best not to do them in simultaneous open PRs!

rebeccafair:
> It's possible to do the work in either order, but there's certainly interaction there and it would be best not to do them in simultaneous open PRs!

Sure, I can try and get on to doing the `Spectrum1D/2D `test refactoring sooner rather than later, so you'll have some useful utilities for testing `Spectrum1DCollection`, `Sequence` etc.

ajjackson:
Updates:
- A feature branch has been created for this PR to target. This will allow us to wait until the next stage is finished/reviewed before any of it hits master. So we can relax a little about the exact names/API at this stage.
- Given the above, and that the plot-to-axis function is a _private_ method in this PR, we can declare that the current naming is good enough to proceed.
- *neutron_band_structure.py* now also works with CASTEP and JSON files. It's pretty neat!

We should really tidy up the naming of things in the scripts directory (or use the `console_scripts` entry point), but that's a separate issue.

scripts/neutron-band-structure.py
```diff
@@ -0,0 +1,209 @@
...
+import seekpath
+
+
+def get_parser() -> argparse.ArgumentParser:
+    parser = argparse.ArgumentParser()
+    parser.add_argument('file', type=str,
+                        help=('File with force constants data. Supported '
+                              'formats: .yaml (Phonopy); .castep_bin, .check '
+                              '(Castep); .json (Euphonic)'))
+    parser.add_argument('--ebins', type=int, default=200)
```
rebeccafair, line 21: There should be some help for `ebins`, (or maybe it should be `n-ebins`?) because from the CL initially I wasn't sure what it wanted

scripts/neutron-band-structure.py
```diff
@@ -0,0 +1,209 @@
...
+    parser.add_argument('--seekpath-labels', action='store_true',
+                        dest='seekpath_labels',
+                        help='Use the exact labels reported by Seekpath when '
+                             'constructing the band structure path. Otherwise,'
+                             ' use the labels detected by Euphonic.')
+    parser.add_argument('--q-distance', type=float, default=0.025,
+                        dest='q_distance',
+                        help='Target distance between q-point samples')
```
rebeccafair, line 29: Specify distance is in 1/angstrom

scripts/neutron-band-structure.py
```diff
@@ -0,0 +1,209 @@
...
+                        help='Matplotlib colormap')
+    parser.add_argument('--q-broadening', type=float, default=None,
+                        dest='gaussian_x',
+                        help='Width of Gaussian broadening on q axis in recip '
+                             'angstrom. (No broadening if unspecified.)')
+    parser.add_argument('--energy-broadening', type=float, default=None,
+                        dest='gaussian_y',
+                        help='Width of Gaussian broadening on energy axis in '
+                        'meV. (No broadening if unspecified.)')
+    return parser
```
rebeccafair, line 45: Should we support different units for energy via an extra argument (`--e-unit`)? The dispersion and dos scripts do. I can't imagine anyone wanting to use anything but `1/angstrom` for q though so that's probably ok as is.

scripts/neutron-band-structure.py
```diff
@@ -0,0 +1,209 @@
...
+        print(f"Found {len(break_points)} regions in q-point path")
+
+    regions = []
+    start_point = 0
+    for break_point in break_points:
+        regions.append((start_point, break_point + 1))
+        start_point = break_point + 1
+    regions.append((start_point, len(bandpath["explicit_kpoints_rel"])))
+
+    print("Computing phonon modes")
```
rebeccafair, line 159: It might be nice to tell the user how many q-points are being computed so they have a vague idea of how long it will take.

scripts/neutron-band-structure.py
```diff
@@ -0,0 +1,209 @@
...
+                        help='Width of Gaussian broadening on q axis in recip '
+                             'angstrom. (No broadening if unspecified.)')
+    parser.add_argument('--energy-broadening', type=float, default=None,
+                        dest='gaussian_y',
+                        help='Width of Gaussian broadening on energy axis in '
+                        'meV. (No broadening if unspecified.)')
+    return parser
+
+
+def get_seekpath_structure(crystal: euphonic.Crystal) -> Tuple[np.ndarray,
```
rebeccafair, line 48: This could be quite a useful function (maybe even as a method `Crystal.to_seekpath_structure`?) since there's already an example of doing this conversion in the docs and just calling this would be easier. Although it can probably be left here for now until we decide there's more uses for it

euphonic/plot.py
```diff
@@ -131,95 +139,144 @@ def plot_1d(spectra, title='', x_label='', y_label='', y_min=None, labels=[],
...
+    ax.images.append(image)
+    ax.set_xlim(min(x_bins), max(x_bins))
+    ax.set_ylim(min(y_bins), max(y_bins))
+
+    _set_x_tick_labels(ax, spectrum.x_tick_labels, spectrum.x_data)
+
+    return image
+
+
+def plot_2d(spectra: Union[Spectrum2D, Sequence[Spectrum2D]],
```
rebeccafair, line 192: The plot functions need to be tested and the docs need to be updated for the new interface (it now just returns a figure). This doesn't have to be done until we've finalised everything in the feature branch but just reminding myself that it needs to be done.

rebeccafair (changes requested):
Looks very nice! Just some comments on the user script

scripts/neutron-band-structure.py
```diff
@@ -0,0 +1,209 @@
+#! /usr/bin/env python3
```
rebeccafair, line 1: This script also needs to be tested

scripts/neutron-band-structure.py
ajjackson: Seems a nice courtesy. While crystallographers and chemists are _fairly_ well-agreed on Angstroms, some theoretical physicists like to work in Bohr.

scripts/neutron-band-structure.py
ajjackson: This is straightforward in the current implementation, but could get more complex as further sampling schemes are added. I guess each mode of operation can be responsible for the message here. It would be cute to have a little progress bar through mod(q) (but add a dependency?)

scripts/neutron-band-structure.py
ajjackson: I figured having it in the script is also an "example" of sorts. Maybe that's less true if we move scripts further into the module structure.

euphonic/plot.py
ajjackson: I really want to put this off but probably shouldn't!

scripts/neutron-band-structure.py
ajjackson: Easier to do if it's in the euphonic namespace, maybe until that is resolved?

rebeccafair: As long as we don't forget! Lots of PRs/issues on the go currently...

scripts/neutron-band-structure.py
rebeccafair: I guess for now we can just add the simple q-point calculation message and deal with it when things get more complicated as there aren't any compatibility issues with just changing how a message is printed.

I did look into a progress bar at one point (https://github.com/tqdm/tqdm) and adding it would be trivial for the Python case, but when calling down to C there's no way to report the progress back unless you periodically pass back to Python and I haven't done the research into whether a nice progress bar would be worth a performance hit (if any).

ajjackson: My understanding is that well-designed progress bars improve _perceived_ performance which can compensate for overhead. There are papers on this but I'm not having much luck accessing them :disappointed:

scripts/neutron-band-structure.py
ajjackson: Have allowed user to set q units; notation in help should look consistent with this option.

scripts/neutron-band-structure.py
ajjackson: Done, and have also allowed q units to be set for consistency

scripts/neutron-band-structure.py
ajjackson: You're right, that would need to be hooked in at a lower level. Have implemented a count for now, should at least help people understand why it freezes when they request an absurdly dense path.

PR #89, opened 2020-08-27
Feature branch: overhaul of spectral plotting

ajjackson:
This feature branch is intended to collect several stages of work on spectral plotting in Euphonic.

This way an overall consistent API / naming scheme can be wrangled here before anything gets back to the master branch.


ajjackson:
I believe the feature work and testing are now complete for this branch. The remaining tasks are to update the user documentation as necessary, and make some decisions about some public/private function/method names.

euphonic/plot.py
```diff
@@ -34,193 +42,288 @@ def plot_dispersion(phonons, btol=10.0, *args, **kwargs):
...
+
+            [Spectrum1DCollection, Spectrum1DCollection, ...]
+
+        Where each segment will be plotted on a separate subplot. (This
+        approach is mainly used to handle discontinuities in Brillouin-zone
+        band structures, so the subplot widths will be based on the x-axis
+        ranges.)
+
+        A singular Spectrum1D or Spectrum1DCollection will be automatically
+        split into segments if btol was set.
```
rebeccafair, line 130: I don't think this comment is true, btol doesn't seem to be passed through anywhere?

ajjackson: Indeed, btol isn't even accepted as an argument here.

We had discussed providing this as a convenience. But I'd be happy to leave it out; `plot_1d(my_spectrum.split())` is not such an unreasonable syntax, and it keeps the code cleaner.

rebeccafair: Sure, I'd be happy with that. Just need to make sure we mention the loss of the `btol` argument (and the `ratio` argument for `plot_2d`) in the changelog

ajjackson:
What were the outstanding things that need doing for this feature branch?

- [ ] Consider making _plot_x_core functions public (maybe e.g. `plot_1d_to_axis`?)
- [ ] Final check that docs are up-to-date

rebeccafair:
Sorry, was halfway through writing this!

- [ ] Decide whether to make `_get_bin_edges/centres` public. If so:
  - [ ]  More foolproofing w.r.t setting `bin_ax/data_ax`
  - [ ] Docstrings, in particular describing any limitations for differently sized bins. I don't know if we need to write anything specific in the docs, they should be self-explanatory...
  - [ ] Tests
- [ ] Decide whether to make public `_plot_1d/2d_core` and what to call them. I think we decided on `plot_1d/2d_to_axis` or similar? If so:
  - [ ] Use cases/examples in the plotting docs.
- [ ] Add some explanation of `Spectrum1DCollection` to the spectra docs.
- [ ] Final check of docs for other things I haven't thought of yet
- [ ] Changelog update!

ajjackson:
I'm in favour of making `get_bin_edges/centres` public, I wouldn't recommend people work without them.

PR #92, opened 2020-09-04
Add weights argument to calculate_qpoint_phonon_modes

rebeccafair:
This adds a q-point `weights` argument to `ForceConstants.calculate_qpoint_phonon_modes` which should make it easier in the future to calculate more efficient symmetry-weighted DOS


euphonic/qpoint_phonon_modes.py
```diff
@@ -53,8 +53,8 @@ def __init__(self, crystal, qpts, frequencies, eigenvectors, weights=None):
         eigenvectors: (n_qpts, 3*crystal.n_atoms, crystal.n_atoms, 3) complex ndarray
             Dynamical matrix eigenvectors
         weights : (n_qpts,) float ndarray, optional
-            The weight for each q-point. If None, equal weights are
-            assumed
+            The weight for each q-point, should sum to 1. If None, equal
+            weights are assumed
```
ajjackson, line 57: This requirement could be a pain-point when integrating with other codes: a lot of codes use integer weights and then normalise (or don't...). Is there a technical reason we'd prefer not relax this and do our own normalisation of weights?

rebeccafair: That's true, I guess provided we ensure weights are normalised it doesn't really matter which convention is used, I just chose summing to 1 because that's what CASTEP uses (and actually when reading from Phonopy to a Euphonic object it converts from the integer convention to summing-to-1 convention for consistency). I'll check that everything that uses weights is normalised and remove the comment about summing to 1.

ajjackson:
Ok, this looks good to me. I see the DOS weights are now normalised appropriately. The test case seems appropriately stressful, including splitting at Gamma.

I do wonder if we have covered all the cases for those options; what exactly happens when a user sets `splitting=True` and `insert_gamma=False`? Should that raise a useful error if the input q-points don't include duplicate GPs?

rebeccafair:
> I do wonder if we have covered all the cases for those options; what exactly happens when a user sets `splitting=True` and `insert_gamma=False`? Should that raise a useful error if the input q-points don't include duplicate GPs?

If there is splitting, but only 1 q-point it will just calculate the splitting in the 'forward' direction, but doesn't complain. I could get it to raise a warning?  That wouldn't be too hard to do, I don't think it's for this PR though!

ajjackson:
Is that a different result from `splitting=False, insert_gamma=False`? Maybe this option is redundant anyway. It's already a bit ugly that we quietly override it if `dipole=False`.

rebeccafair:
Yes, if `splitting=False` it doesn't calculate any correction at gamma at all. It does say in the docstring for `splitting` that it is only applied if `dipole=True` and the Born charges are present, and in `insert_gamma` it says it only inserts q-points if `splitting=True`, but maybe it could be made clearer that if no splitting is applied, no extra gamma points are inserted.

I think we settled on having an `insert_gamma` option because we couldn't decide on the convention for gamma points and whether people would expect to duplicate their own gammas or not...

ajjackson:
So if `dipole=True, splitting=False, insert_gamma=False`, it applies the non-analytic correction to all points _except_ Gamma? Is that useful to anyone?

rebeccafair:
Sorry, when I said 'no correction at all', I meant no non-analytical correction, the dipole correction is still applied! The non-analytic correction described by `splitting` is **only** ever applied at gamma, as its what calculates LO-TO splitting. The `dipole` correction is the correction that gets applied to all points. (see https://journals.aps.org/prb/abstract/10.1103/PhysRevB.55.10355 section III.B. describes the dipole correction, III.C. describes the gamma-only non-analytical correction).

ajjackson:
Right. I'm still not sure it ever makes sense to not do it if you are including the dipole corrections elsewhere and your q-point list includes Gamma.

Anyway, this can be discussed/tweaked separately if need be; the main thing is that it doesn't have bad interactions with the changes in this PR!

Happy for this to be merged; I don't have permission to do it.

Issue #90, opened 2020-08-28
Refactor util.py


rebeccafair:
`util.py` is quickly becoming a dumping ground for any new functions that don't have an obvious place anywhere else. This has 2 main issues (and probably more):
- Makes it difficult to find the function that you want (increasing the chance of you rewriting it yourself and duplicating code functionality)
- Increases the amount of dependencies, given that many classes are imported in `euphonic.__init__` (which depend on `util`, which has lots of dependencies...) this increases the chance of circular dependencies

Effort should be made to refactor `util.py` out into separate files and/or put the functions into appropriate classes before it gets out of hand.

ajjackson:
I'll add here that the specific practice which got me into dependency hell was importing classes to use in type annotations, e.g.

```python
from typing import List, Tuple
from euphonic.crystal import Crystal
def get_qpoint_labels(crystal: Crystal, qpts: np.ndarray) -> List[Tuple[int, str]]:
    ...
```

I think it would be a reasonable practice to avoid that specific problem by deferring the mypy imports with e.g.
```python
from typing import List, Tuple
import euphonic

def get_qpoint_labels(crystal: 'euphonic.Crystal', qpts: np.ndarray) -> List[Tuple[int, str]]:
    ...
```
It's not pretty but could at least be a consistent pattern within this file.

However we could easily hit this problem outside of type-hinting. It is not safe to import Euphonic classes at the top of this module, when so many of them use it. Deferred imports inside functions will sometimes help, but they make the problem harder to examine.

PR #96, opened 2020-10-02
90 refactor util

rebeccafair:
Addresses #90

Refactoring to reduce the size of util.py by moving functions with a similar purpose into separate modules, and putting functions into the relevant classes where possible. Commits detail specific changes, hopefully they all make sense.

This has made some changes to how some of the utility functions associated with labelling/plotting (`get_qpoint_labels`, `calc_abscissa`) are called to remove the dependence on `Crystal`, but hopefully it won't affect the plotting refactoring too much.


ajjackson:
This looks cleaner, easier to navigate and a touch more robust.

My only question is regarding the significance of the leading underscores on the functions accessed from validate. If we're routinely accessing them from other modules, shouldn't they be public?

ajjackson (approved):
As commented, this is good stuff and I don't see any harm done.

Would query use of leading underscores in the validate functions, but this is not a condition of approval.

ajjackson:
There is a bit of complexity introduced by adopting the spglib cell format alongside our own Crystal class. It's good for interoperability, though. Might also help with import loops, although that isn't the _best_ reason to add redundancy to a code...

rebeccafair:
> This looks cleaner, easier to navigate and a touch more robust.
> 
> My only question is regarding the significance of the leading underscores on the functions accessed from validate. If we're routinely accessing them from other modules, shouldn't they be public?

Yes, I've been wondering what to do about the leading underscores, because although other Euphonic classes/modules use them, I don't really intend for them to be used standalone and I wouldn't want external users or libraries to use them as I'd have to write standalone tests for them and it would take more effort to change their implementation later, I can't seem to find a good consensus on this anywhere....

I guess one solution could be to have all Euphonic object subclass a class which contains the validate functions, but maybe that's over-engineering

ajjackson:
> Yes, I've been wondering what to do about the leading underscores, because although other Euphonic classes/modules use them, I don't really intend for them to be used standalone and I wouldn't want external users or libraries to use them as I'd have to write standalone tests for them and it would take more effort to change their implementation later, I can't seem to find a good consensus on this anywhere....

If the intended meaning is "seriously, use with care" then I'm ok with that. Perhaps that should be documented somehow?

rebeccafair:
> Perhaps that should be documented somehow?

I guess I could put something in the docstring to the gist of `Warning: Intended for internal use only, may change at any time`, but maybe that's redundant as the underscore should imply that anyway. There are a few 'internal' Euphonic functions like this so we should probably come up with a convention and stick with it

PR #104, opened 2020-10-30
Use of Brille in Euphonic

rebeccafair:
I've been doing a lot more thinking than coding, so this PR is still at an early stage but I thought this would be a good time/place to review things so far to fully determine the scope of this work, review syntax and ensure there aren't any misunderstandings w.r.t how Brille works that could cause problems later.

So far I've written simple a `BrilleInterpolator` (name up for debate) class, that acts very much like `ForceConstants` in that it can calculate  `QpointPhononModes` at arbitrary q-points. This is the **bare minimum** that would be required for Horace, as we can now calculate `StructureFactor/DebyeWaller/whatever` from `QpointPhononModes`. However, this is probably not the most efficient way to do things, as we have to convert from Brille's eigenvector convention (basis) to Euphonic's (Cartesian), and there is no chunking of the calculation so all eigenvectors would have to be contained in memory. From this there are two questions:
 * Is the eigenvector conversion actually a bottleneck? (I guess I'd better do some profiling)
 * Should chunking of the structure factor calculation be implemented in Euphonic or the Horace interface? If it was to be implemented in Euphonic a `BrilleInterpolator.calculate_structure_factor` method would have to be added, in which case should there also be a `ForceConstants.calculate_structure_factor` method? The driving factor here should probably be would Euphonic users want to use this or only users of Horace?

The other issue with going with the 'bare minimum' implementation is that Euphonic users won't get the benefits of Brille's IBZ-finding for Debye Waller/DOS calculations. So far I've included an initial example of how `BrilleInterpolator.calculate_debye_waller` might work.

I guess in short, we need to balance development & maintenance time vs. benefit to users. The question is how far should this initial integration go? Should Euphonic implement the full set of `BrilleInterpolator/ForceConstants.calculate_xyz`, stick with the bare minimum, or somewhere inbetween? Sticking with the bare minimum for now would allow us to at least get something working, and add features/performance enhancements later, but could result in poor performance in the meantime for Horace users.

Also here's an example of what the syntax would currently look like:
```
import seekpath
from euphonic import ureg, ForceConstants
from euphonic.brille_interpolator import BrilleInterpolator

fc = ForceConstants.from_castep('quartz.castep_bin')
gkw = {'num_levels': 3, 'max_size': 0.01, 'max_points': -1}
ikw = {'use_c': True, 'n_threads': 2}
bri = BrilleInterpolator(fc, grid_type='mesh', grid_kwargs=gkw, interp_kwargs=ikw)

qpts = seekpath.get_explicit_k_path(fc.crystal.to_spglib_cell())['explicit_kpoints_rel']
brille_interp_kw = {'useparallel': True, 'threads': 2}
phon_brille = bri.calculate_qpoint_phonon_modes(qpts, **brille_interp_kw)
dw = bri.calculate_debye_waller(300*ureg('K'))

```


mducle:
I think that going with the "bare minimum" for now and then improving it later is the best option. Also, maybe you can reuse code from `brilleu`: https://github.com/brille/brilleu ? 

And there seems to be a `debye-waller` method (in C++) in Brille already ( https://brille.github.io/stable/api/classInterpolationData.html?highlight=debye#_CPPv4II0E0ENK17InterpolationData12debye_wallerE11ArrayVectorIdERK1AIdERKNSt6vectorIdEEKd ) although I'm not sure how it works - @g5t ?

rebeccafair:
> Also, maybe you can reuse code from `brilleu`: https://github.com/brille/brilleu ?

Most of this is based off Brilleu - that's how I figured out how to use Brille! The only issue with using code directly from Brilleu is that it duplicates some of the stuff that's in Euphonic already, and I wanted to add the minimum amount of code to Euphonic to make it easier to maintain

> And there seems to be a `debye-waller` method (in C++) in Brille already ( https://brille.github.io/stable/api/classInterpolationData.html?highlight=debye#_CPPv4II0E0ENK17InterpolationData12debye_wallerE11ArrayVectorIdERK1AIdERKNSt6vectorIdEEKd ) although I'm not sure how it works - @g5t ?

I think this was originally implemented in Brille because of being unable to share memory between C++/Python to avoid copying eigenvectors - but this is (or will be?) fixed in a newer version of Brille so it makes sense to use the implementation in Euphonic

rebeccafair:
Ok, before I go any further with this PR and write proper tests, documentation etc. its probably a good time to have a quick review of the syntax. I have added:
* A `euphonic.brille_interpolator.BrilleInterpolator` class
* This has a `calculate_qpoint_phonon_modes` method just like `ForceConstants`, except it performs linear interpolation on the grid included in the `BrilleInterpolator` object. Any extra kwargs are passed through to the Brille grid's `ir_interpolate_at` function.
*  The `BrilleInterpolator` constructor takes a `Crystal` object and a filled Brille grid object (`BZTrellisQdc`,`BZMeshQdc` or `BZNestQdc`). This fits with the constructors in the rest of Euphonic and would allow a user to create and fill their own grid if they wanted (although most users won't).
* There is a `BrilleInterpolator.from_force_constants` class method to create from an existing `ForceConstants` object. Most users will probably use this. Creates a Brille grid (type and approximate number of q-points can be set by the user) and fills it using frequencies/eigenvectors calculated by `ForceConstants`.

See ` euphonic/brille_interpolator.py` for more details. Any thoughts @mducle @ajjackson @g5t? If it looks ok I can add docs/tests and tie up the PR.

mducle:
@rebeccafair I think the syntax you've outlined is good. Is / will `euphonic.BrilleInterpolator` be an alias for `euphonic.brille_interpolator.BrilleInterpolator` like `euphonic.ForceConstants` is an alias for `euphonic.force_constants.ForceConstants`?

Also, I guess we should expose `BrilleInterpolator` to Matlab in the `horace-euphonic-interface` too?

rebeccafair:
> @rebeccafair I think the syntax you've outlined is good. Is / will `euphonic.BrilleInterpolator` be an alias for `euphonic.brille_interpolator.BrilleInterpolator` like `euphonic.ForceConstants` is an alias for `euphonic.force_constants.ForceConstants`?
> 
> Also, I guess we should expose `BrilleInterpolator` to Matlab in the `horace-euphonic-interface` too?

I would alias it, but the only issue is I currently have added brille as an optional dependency, so we can't import `BrilleInterpolator` in `__init__.py` like we have with the other classes. We could add brille as a required dependency and alias it in the future if we wanted, it depends how many users will need to use brille...

And yes, I have done some initial work on a horace-euphonic-interface branch https://github.com/pace-neutrons/horace-euphonic-interface/tree/1_brille. I'll create PRs for all the repos and make sure everything works together before merging anything.

mducle:
For the import in `euphonic/__init__.py` could we put it in a `try` block, e.g. `try: import brille; except ImportError: pass; else: from brille_interpolator import BrilleInterpolator`?

rebeccafair:
We could, but it sort of breaks how optional imports are handled in Euphonic... if a user tries to import `BrilleInterpolator` but brille hasn't been installed, we want to give them a helpful message saying they should've done pip install euphonic[brille]. If we add the try/except then they won't get that... But then again since we're also providing conda packages maybe that message isn't very helpful.

mducle:
Ah right... ok, well I guess a bit more typing never hurt anyone...

ajjackson:
This sort of thing is usually handled with plugin/factory systems where you poke some object that checks what's available. I'm not really a fan though.

Would it make sense to have a `brille` module which could contain others as helpful for organisation and expose the key stuff at the top? i.e.

```
from euphonic.brille.interpolator import BrilleInterpolator  # from euphonic/brille/interpolator.py

from euphonic.brille import BrilleInterpolator   # from euphonic/brille/__init__.py, which contains the above line
```

Then the user API does not have to be horribly deep/verbose and brille-related features can be organised together.

rebeccafair:
That seems like a nice idea, I definitely like `from euphonic.brille import BrilleInterpolator` more than `from euphonic.brille_interpolator import BrilleInterpolator`. I can't think of any other brille utilities that we might want to add at the moment but I guess there's no harm in having the submodule there even if it is just to make the imports nicer.

ajjackson:
For now you could just rename *brille_interpolator.py* to *brille.py*, and keep that option open for the future.

The only difference it would make to the user at this stage is that the import statements from `brille` that might go away into another namespace. But really if somebody's going to `from euphonic.brille import QPointPhononModes` then they'll get what they deserve when it breaks...

rebeccafair:
@ajjackson It would be nice to have your review on the Brille-related CL tool arguments at this point, and whether they make sense to someone who hasn't seen Brille too much before. I have only made them accessible from the `euphonic-powder-map` tool, as that's the only one that does enough computation to make it worth it.

Also, in terms of type-hinting, both `ForceConstants` and `BrilleInterpolator` provide  `calculate_qpoint_phonon_modes/frequencies` methods, so can be used as an argument for things such as `euphonic.powder.sample_sphere_dos`. There are 2 main options:
* Use `sample_sphere_dos(fc: Union[ForceConstants, 'BrilleInterpolator'] ...)`, bearing in mind `brille` is an optional dependency so `BrilleInterpolator` will probably have to be defined by a string.
* Both inherit from a superclass with abstract `calculate_qpoint_phonon_modes/frequencies` methods. What could we name this? `PhononCalculator` may lead to confusion with how Phonopy defines 'phonon calculators', but is something like `PhononInterpolator` misleading, could there be non-interpolating methods in the future?

ajjackson:
> Also, in terms of type-hinting, both `ForceConstants` and `BrilleInterpolator` provide `calculate_qpoint_phonon_modes/frequencies` methods, so can be used as an argument for things such as `euphonic.powder.sample_sphere_dos`. There are 2 main options:
> 
>     * Use `sample_sphere_dos(fc: Union[ForceConstants, 'BrilleInterpolator'] ...)`, bearing in mind `brille` is an optional dependency so `BrilleInterpolator` will probably have to be defined by a string.
> 
>     * Both inherit from a superclass with abstract `calculate_qpoint_phonon_modes/frequencies` methods. What could we name this? `PhononCalculator` may lead to confusion with how Phonopy defines 'phonon calculators', but is something like `PhononInterpolator` misleading, could there be non-interpolating methods in the future?

I don't have a name suggestion at this moment, but it occurs to me that option 2 would be backward-compatible with option 1, even with types enforced. So we could start with the Union and it shouldn't cause many problems if we come up with a more elaborate setup later.

ajjackson:
I'm getting some segfaults with the LZO data:

```
(science) localhost:~/src/euphonic (95_brille_integration)$ euphonic-powder-map tests_and_analysis/test/data/force_con
stants/LZO_force_constants.json --use-brille
ForceConstants data was loaded
Generating grid...
Grid generated with 8794 q-points. Calculating frequencies/eigenvectors...
Filling grid...
Setting up dimensions...
Sampling 120 |q| shells between 0.0 1/Å and 3.0 1/Å
  4%|███▍                                                                             | 5/120 [00:00<00:14,  7.78it/s]The node subscript  6 6 0 for the point [[ 0.0903,  0.1026,  0.0147]] is either invalid or points to a null node!
terminate called after throwing an instance of 'std::logic_error'
  what():  attempting to access null node!
Aborted (core dumped)
```

(I never hit this yesterday with phonopy data, maybe the LZO data has an unusual property?)

ajjackson:
The CaHgO2 data gives a different problem:

```
(science) localhost:~/src/euphonic (95_brille_integration)$ euphonic-powder-map tests_and_analysis/test/data/force_con
stants/CaHgO2_force_constants.json --use-brille
ForceConstants data was loaded
Generating grid...
Grid generated with 8813 q-points. Calculating frequencies/eigenvectors...
Filling grid...
Setting up dimensions...
2 : ( 0.894013 0.894013 0.894013 )
2 : ( 0.105987 0.105987 0.105987 )
0 : ( 0.500000 0.500000 0.500000 )
1 : ( 0.000000 1.000000 0.000000 )

does not have an equivalent atom to 0 for symmetry operation
  0 -1  0 -1  0  0  0  0 -1+  0.0000  0.0000  0.0000
  File "/home/adamjackson/anaconda3/envs/science/bin/euphonic-powder-map", line 33, in <module>
    sys.exit(load_entry_point('euphonic', 'console_scripts', 'euphonic-powder-map')())
  File "/home/adamjackson/src/euphonic/euphonic/cli/powder_map.py", line 95, in main
    modes = fc.calculate_qpoint_frequencies(
  File "/home/adamjackson/src/euphonic/euphonic/brille.py", line 108, in calculate_qpoint_frequencies
    vals, _ = self._br_grid_calculate_phonons(self._grid, qpts, **kwargs)
  File "/home/adamjackson/src/euphonic/euphonic/brille.py", line 118, in _br_grid_calculate_phonons
    return grid.ir_interpolate_at(qpts, **kwargs)
RuntimeError: All atoms in the basis *must* be mapped to an equivalent atom for *all* symmetry operations
```

ajjackson:
For the cases that work, I'm finding that the CLI arguments are nice; `--use-brille`, `--brille-grid-type` and `--brille-n-qpts` give some parameters to play with without exposing too many implementation details.

I'm finding the performance/reliability to be a bit machine/build-dependent still, will keep playing with that.

I think it would be useful to include a "Brille error estimator" and have some ideas about how it could work. Then, like the `euphonic-optimise-dipole-parameter`, we can offer some guidance with a quick calculation.

tests_and_analysis/test/euphonic_test/test_brille_interpolator.py
```diff
@@ -0,0 +1,294 @@
...
+    from importlib import reload
+    import euphonic.brille
+    real_import = builtins.__import__
+    def mocked_import(name, *args, **kwargs):
+        if name == 'brille':
+            raise ModuleNotFoundError
+        return real_import(name, *args, **kwargs)
+    mocker.patch('builtins.__import__', side_effect=mocked_import)
+    with pytest.raises(ModuleNotFoundError) as mnf_error:
+        reload(euphonic.brille)
```
mducle, line 52: Just curious why you have to import `euphonic.brille` correctly first before patching `__import__` to raise the module error and then check it on reload. Is this both a check that it can be imported correctly as well as that the correct error is raised if Brille cannot be imported?

tests_and_analysis/test/euphonic_test/test_brille_interpolator.py
```diff
@@ -0,0 +1,294 @@
...
+            self, faulty_crystal, grid_file, expected_exception):
+        grid = get_brille_grid(grid_file)
+        with pytest.raises(expected_exception):
+            BrilleInterpolator(faulty_crystal, grid)
+
+    # Deliberately create a grid incompatible with the quartz crystal
+    # and check that it raises an error
+    @pytest.mark.parametrize('expected_exception, grid_file',
+        [(ValueError, 'lzo_trellis_10.hdf5'),
+         (ValueError, 'quartz_mesh_10_unfilled.hdf5')])
```
mducle, line 169: So is the second case the error is raised because the grid is not filled, rather than because it is not the right crystal?

tests_and_analysis/test/euphonic_test/test_brille_interpolator.py
```diff
@@ -0,0 +1,294 @@
...
+        # Calculate structure factor 1D to test eigenvectors
+        ebins = np.arange(5, 190)*ureg('meV')
+        sf_fc = qpm_fc.calculate_structure_factor()
+        sf1d_fc = sf_fc.calculate_1d_average(ebins)
+        sf_brille = qpm_brille.calculate_structure_factor()
+        sf1d_brille = sf_brille.calculate_1d_average(ebins)
+        # Tolerances are quite generous, but that is required unless
+        # we have a very dense grid (expensive to test)
+        check_spectrum1d(sf1d_brille, sf1d_fc,
+                         y_rtol=0.01, y_atol=5e-3)
```
mducle, line 222: I think a 1% tolerance is fine, actually, considering this is meant to be a fast approximate method. It seems to be better that what was obtained in the LZO example in the documentation where at low energies the relative frequency residual seems to be around 2% (from the graph, a residual of 0.15meV at around 7meV).

mducle (approved):
Looks really nice!

I've just got a few comments, but I think it's fine as it is.

euphonic/brille.py
```diff
@@ -0,0 +1,257 @@
...
+            method. By default useparallel=True and
+            threads=multiprocessing.cpu_count() are passed.
+
+        Returns
+        -------
+        qpoint_frequencies
+            An object containing frequencies linearly interpolated
+            at each q-point
+        """
+        vals, _ = self._br_grid_calculate_phonons(qpts, **kwargs)
```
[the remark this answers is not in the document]
rebeccafair: It _should_ be faster, but I'd have to do the profiling first to be sure, and I'm not sure its worth the extra effort at the moment as I think `calculate_qpoint_phonon_modes` will be used more, frequencies-only is probably only useful for DOS calculations.

euphonic/cli/utils.py
```diff
@@ -421,13 +421,31 @@ def _plot_label_kwargs(args: Namespace, default_xlabel: str = '',
...
+    """
+    if args.n_threads is None:
+        # Nothing specified, allow defaults
+        return dict()
+    else:
+        if args.n_threads > 1:
+            parallel = True
+        else:
+            parallel = False
+        return dict(useparallel=parallel, threads=args.n_threads)
```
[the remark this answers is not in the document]
rebeccafair: Ah maybe, I was just following convention in the rest of the file that uses `dict`...

tests_and_analysis/test/euphonic_test/test_brille_interpolator.py
rebeccafair: I think I got this pattern from https://github.com/pace-neutrons/Euphonic/blob/master/tests_and_analysis/test/euphonic_test/test_plot.py#L41-L56 and didn't think too much about it...

The original test was added in #89 https://github.com/pace-neutrons/Euphonic/blame/7c8694618a978ebced0e9eba9a9098645025249b/tests_and_analysis/test/euphonic_test/test_plot.py, I think what you suggested is probably the case, there isn't anything obvious there to state otherwise.

tests_and_analysis/test/euphonic_test/test_brille_interpolator.py
rebeccafair: Yes

tests_and_analysis/test/euphonic_test/test_brille_interpolator.py
rebeccafair: Yeah, this test actually only creates a spectrum starting from 5 meV because I thought trying to reliably test those low frequencies would be a bit of a nightmare

euphonic/brille.py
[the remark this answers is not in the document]
mducle: Yeah, I think leave it for now, and if we find out it's needed later to add it then.

euphonic/cli/utils.py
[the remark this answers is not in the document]
mducle: To be honest I don't think the performance issue is worth worrying about here and we can just leave it as `dict(**kwargs)`, it just occurred to me and I googled some stuff and it seems it does make a (bit of a) difference.

mducle (approved):

PR #106, opened 2020-11-05
Add metadata to .json outputs and associated tests

rebeccafair:
Adds metadata to .json outputs describing the Euphonic version, and the class being written, to make it easier to automatically determine object types from files. This is to aid in reading files for command line tools, and came about as a result of https://github.com/pace-neutrons/Euphonic/pull/100#discussion_r517954830.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -3,8 +3,6 @@
 
 - Changes:
 
-  - ``yaml.CSafeLoader`` is now used instead of ``yaml.SafeLoader`` by
-    default, so Phonopy ``.yaml`` files should load faster
   - Some of Euphonic's dependency version requirements have been changed - but
     can now be relied on with more certainty due to better CI testing. This
     includes:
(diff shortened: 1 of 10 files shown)


rebeccafair:
Does this need a changelog entry? I don't really intend for users to make use of the labels, but it might be useful to know from when the extra keys were added to JSON output...

ajjackson:
Seems worth a one-liner

> Does this need a changelog entry? I don't really intend for users to make use of the labels, but it might be useful to know from when the extra keys were added to JSON output...

That's my thought as well. It will be useful to find when this happened from the CHANGELOG

ajjackson:
Am I right in thinking that this PR adds the metadata but does not yet require/use it for anything, so the test data does not need to be updated?

rebeccafair:
> Am I right in thinking that this PR adds the metadata but does not yet require/use it for anything, so the test data does not need to be updated?

Indeed! I just updated the .json tests to make sure the metadata is actually being written because we'd want to know if it suddenly stopped for some reason..

ajjackson (approved):
It works for me, and the implementation is simple enough. Approve!

Issue #108, opened 2020-11-19
Strange broadening with bright Bragg peaks


rebeccafair:
I've found that when broadening S(Q,w) data that contains data with very bright Bragg peaks (~10 orders of magnitude brighter than the rest of the data) the plots produced are very strange, seen below for LZO and Nb:

![image](https://user-images.githubusercontent.com/18187336/99651981-2cf57980-2a4f-11eb-9842-5462333be221.png)
![image](https://user-images.githubusercontent.com/18187336/99652020-367ee180-2a4f-11eb-8f56-949ac0c4c3fb.png)

Similar behavour is seen with both 'gauss' and 'lorentz' broadening. It should be investigated to see whether this is unavoidable or a result of the broadening method used.

ajjackson:
I think I understand what is happening. This is only seen when extreme amounts of clipping is applied to the data (~ 5 orders of magnitude or more!) by setting `vmax` in order to visualise the fainter plot lines. This raises the noise floor of the FFT convolution from imperceptible levels to something more dramatic. I suspect the problem will also be reduced by using more bins, as the precision of the FFT representation of the signal is directly related to the sampling frequency.

Solutions include:
- perform the broadening in realspace, so that precision depends on our float datatype. It's all addition and multiplication so should hold up well over a wide data range. (I think this is the approach you went for? `scipy.signal.oaconvolve` should be a drop-in replacement that works efficiently with these small broadening kernels.)
- apply clipping _before_ broadening so that the noise floor is consistent across the visible data. The problem is that we are then no longer simply passing `vmax` to Matplotlib, so there become two ways of doing things - more mess.
- somehow transform the space to e.g. logarithmic and apply the broadening there. Haven't really worked out if that's actually possible, may need to modify the kernels. Let's not do that.

ajjackson:
Have had a little play and the implementation in #109 seems to have a performance penalty compared to the existing implementation. `oaconvolve` doesn't make any difference, and `signal.convolve(method='direct')` has worse performance while also solving the problem.

![euphonic-convolution-options](https://user-images.githubusercontent.com/2511383/99959806-8fb68000-2d83-11eb-8a4c-46819ec7b8e2.png)

PR #109, opened 2020-11-19
108 improve broadening

rebeccafair:
#108 

Moving from `scipy.signal.fftconvolve` to `scipy.ndimage.gaussian_filter` seems to have fixed the issues, and I think is probably a more stable implementation as we are actually using something intended for images and we no longer have to generate our own Gaussian. There wasn't an implementation of Lorentzian filter so `scipy.ndimage.correlate` had to be used with the existing Lorentzian implementation, but that seems to have fixed the issues seen with `shape='lorentz'` too. See updated plots below:

image
image

This has also has minimal effect on the broadening elsewhere, the following are some examples of 1D broadening test data:
image
image

I've also compared the 2D test data by eye which looks the same, and made some 1D slices through the 2D data to check:
image
image

The peaks with the Gaussian broadening are slightly different on the 2D data, but I don't think it's enough to worry about. Even though the data is so similar, I still had to regenerate the test data to keep the tolerances low.


ajjackson:
Why use `ndimage.correlate` rather than `ndimage.convolve`? Is there a performance difference? I believe _correlate_ flips the kernel - not a problem for our symmetric kernel but could be very confusing if that ever changes! https://dsp.stackexchange.com/questions/64063/convolution-vs-correlation

rebeccafair:
In addition to avoiding the MemoryError, using successive 1D operations instead of 2D `correlate` has improved the performance and it is now similar to the existing broadening implementation on master (using `correlate` had a large performance penalty). See below for comparisons calculating broadening for ~800 x/y bins between the current master implementation (`Spectrum2DOld`), `correlate1d` (`convolve=False`) and `convolve1d` (`convolve=True`) for Gaussian and Lorentzian broadening, and for smaller and larger widths. For large widths and Lorentzian broadening master is ~2x faster, but that might be an unrealistically large broadening, and they are of the same order at least. There doesn't seem to be any performance difference between `convolve1d` and `correlate1d`

```
Line #      Hits         Time  Per Hit   % Time  Line Contents
==============================================================
     9                                           @profile
    10                                           def main():
    11         1    3645503.0 3645503.0     10.5      spec2d = Spectrum2D.from_json_file('spectrum-800.json')
    12         1    3623569.0 3623569.0     10.5      spec2d_old = Spectrum2DOld.from_json_file('spectrum-800.json')
    13                                           
    14         1         18.0     18.0      0.0      shape = 'gauss'
    15         1       5105.0   5105.0      0.0      xwidth = 0.05*ureg('1/angstrom')
    16         1       5579.0   5579.0      0.0      ywidth = 2*ureg('meV')
    17         1         14.0     14.0      0.0      bkwargs = {'x_width': xwidth,
    18         1          7.0      7.0      0.0                 'y_width': ywidth,
    19         1          9.0      9.0      0.0                 'shape': shape}
    20                                           
    21                                               # Gaussian broadening, small widths
    22         1    1157972.0 1157972.0      3.3      spec2db = spec2d.broaden(**bkwargs)
    23         1    1659371.0 1659371.0      4.8      spec2db_old = spec2d_old.broaden(**bkwargs)
    24                                           
    25                                               # Lorentz broadening, small widths
    26         1         37.0     37.0      0.0      bkwargs['shape'] = 'lorentz'
    27         1    1486169.0 1486169.0      4.3      spec2db = spec2d.broaden(**bkwargs, convolve=False)
    28         1    1444162.0 1444162.0      4.2      spec2db_cv = spec2d.broaden(**bkwargs, convolve=True)
    29         1    1608197.0 1608197.0      4.6      spec2db_old = spec2d_old.broaden(**bkwargs)
    30                                               
    31                                               # Lorentz broadening, large widths
    32         1       8281.0   8281.0      0.0      bkwargs['x_width'] = 0.5*ureg('1/angstrom')
    33         1      11666.0  11666.0      0.0      bkwargs['y_width'] = 10*ureg('meV')
    34         1    5304951.0 5304951.0     15.3      spec2db = spec2d.broaden(**bkwargs, convolve=False)
    35         1    5306393.0 5306393.0     15.3      spec2db_cv = spec2d.broaden(**bkwargs, convolve=True)
    36         1    3059099.0 3059099.0      8.8      spec2db_old = spec2d_old.broaden(**bkwargs)
    37                                           
    38                                               # Gaussian broadening, large widths
    39         1         44.0     44.0      0.0      bkwargs['shape'] = 'gauss'
    40         1    3359787.0 3359787.0      9.7      spec2db = spec2d.broaden(**bkwargs)
    41         1    2971846.0 2971846.0      8.6      spec2db_old = spec2d_old.broaden(**bkwargs)
```

ajjackson:
Here's a performance test against the old implementation. This is looking a lot better than the 2D convolution, with a performance gap of around x10 at the upper end of a "reasonable" plotting case (2000 x 2000). There does still seem to be some scaling issue though, especially compared with the scipy.signal implementation which is pretty linear with the number of pixels.

euphonic-conv-benchmark

ajjackson:
For Gaussian broadening it looks like the prefactor is gone while scaling is similar.

euphonic-conv-gauss-benchmark

ajjackson:
For Gaussian broadening it looks like the prefactor is gone while scaling is similar.

euphonic-conv-gauss-benchmark

rebeccafair:
It's been decided that this performance drop is acceptable and isn't likely to be a bottleneck in the immediate future as it is not likely to be performed in tight loops. If it does become a bottleneck the broadening method can be reviewed.

PR #113, opened 2020-12-17
Script label tests

rebeccafair:
Add checking of plot labels to script tests


euphonic/plot.py
```diff
@@ -170,17 +170,18 @@ def plot_1d(spectra: Union[Spectrum1D,
...
         if i == 0 and labels:
             ax.legend(labels)
 
-    #    if y_min is not None:
-    ax.set_ylim(bottom=y_min, top=y_max)
+    # Add an invisible large axis for common labels
+    ax = fig.add_subplot(111, frameon=False)
+    ax.tick_params(labelcolor="none", bottom=False, left=False)
+    ax.set_xlabel(x_label)
+    ax.set_ylabel(y_label)
```
ajjackson, line 184: Are there any consequences associated with this change to put the ylabel on the invisible axis, rather than the left-most? I think the coordinates determined by subplots_adjust will be the same so it should be harmless.

ajjackson (approved):
The code changes look fine, and cover a few more things that might go wrong.
Good to sort out the x-label placement at the same time.

Slightly annoying that the JSON diffs are so large, presumably this is to do with a change in indentation level? I wonder if a little hand-tuning can solve that.

euphonic/plot.py
rebeccafair: I actually copied this snippet from your 2D plot implementation for consistency haha, I checked all the figures by eye and the labels look fine, so I think we're good. Unfortunately I can't remember if there was an additional reason for doing this...

ajjackson: Simplicity/consistency I guess!

PR #115, opened 2021-01-14
Test plotting module

ajjackson:
Simple unit tests for euphonic.plot


ajjackson:
Ok, I'm getting 100% coverage on `euphonic.plot` now :smile: 

Not sure that _everything_ is being thoroughly checked for correctness, but a lot of it is.

tests_and_analysis/test/euphonic_test/test_plot.py
```diff
@@ -0,0 +1,332 @@
...
+
+        assert phonons.get_dispersion.called
+        if 'btol' in kwargs:
+            bands.split.assert_called_once_with(btol=kwargs['btol'])
+        else:
+            bands.split.assert_called_once_with(btol=10.)
+
+        euphonic.plot.plot_1d.assert_called_once_with(
+            [1, 2],
+            **{k: v for k, v in kwargs.items() if k != 'btol'})
```
rebeccafair, line 220: This has also made me realise that `get_dispersion` is actually not tested in `test_qpoint_phonon_modes`, but it really should be. And should `StructureFactor` also have a `get_dispersion` method? It contains all the information needed for it and might be something people would want to do

rebeccafair (commented):
Looks good! This is enough for our purposes, I think. Just another test on `plot_dispersion` might be helpful

euphonic/plot.py
```diff
@@ -166,11 +163,9 @@ def plot_1d(spectra: Union[Spectrum1D,
 
     gridspec_kw = _get_gridspec_kw(spectra)
     fig, subplots = plt.subplots(1, len(spectra), sharey=True,
-                                 gridspec_kw=gridspec_kw)
-    if not isinstance(subplots, np.ndarray):  # if only one subplot
-        subplots = np.array([subplots])
+                                 gridspec_kw=gridspec_kw, squeeze=False)
```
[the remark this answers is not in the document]
ajjackson: Yeah, I was finding it slightly frustrating to test the logic and then realised this would be _way_ cleaner 😄 

I might just get into the habit of always using `squeeze=False` now for non-interactive sessions.

tests_and_analysis/test/euphonic_test/test_plot.py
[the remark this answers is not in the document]
ajjackson: If we make changes to `get_dispersion()` or `plot_1d` then the output of `plot_dispersion` _should_ change.
The implementation of `plot_dispersion` is down to two lines now, we could just remove this function entirely.

ajjackson: Yes to both! But preferably not in this PR...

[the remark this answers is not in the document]
rebeccafair: Hmm, maybe you're right and we should just remove `plot_dispersion`. In that case we should explicitly mention the new syntax in the changelog (and the documentation, but that goes without saying).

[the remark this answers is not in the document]
ajjackson: Shall I move on with this, then? Removing `plot_dispersion` and giving hints in docs about how to plot a band structure?

ajjackson: Are we going to require `StructureFactor.get_dispersion` for this feature branch? Or open a separate Issue?

rebeccafair: Perhaps a separate issue? There might need to be some thought over whether we should just duplicate the `get_dispersion` code or have some kind of `QpointPhononFrequencies` superclass or something else entirely

ajjackson: Done! #117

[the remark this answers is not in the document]
ajjackson: Done! I also updated some of the docs for plot_1d and plot_2d.

ajjackson:
Ok very weird that the last changes have broken the Matplotlib ModuleNotFound test. Any ideas?

doc/source/plotting.rst
```diff
@@ -13,42 +13,64 @@ be tweaked, then viewed with ``matplotlib.figure.Figure.show()``
...
-  fig = plot_dispersion(phonons, y_label='Energy (meV)')
+  bands = phonons.get_dispersion()  # type: Spectrum1DCollection
+
+  fig = plot_1d(bands, y_label='Energy (meV)')
   fig.show()
 
-Docstring
----------
+Phonon dispersion plots often include large steps between
+discontinuous regions. In order to accommodate this, a single
```
rebeccafair, line 35: Maybe should be 'discontinuous regions in reciprocal space' or similar to be more specific

doc/source/plotting.rst
```diff
@@ -13,42 +13,64 @@ be tweaked, then viewed with ``matplotlib.figure.Figure.show()``
...
+A compact recipe to write a band-structure plot with discontinuities could be
+
+.. code-block:: py
+  from euphonic import QpointPhononModes
+  from euphonic.plot import plot_1d
+
+  phonons = QpointPhononModes.from_castep('quartz.phonon')
+
+  fig = plot_1d(phonons.get_dispersion().split())
+  fig.safefig('quartz-dispersion.pdf')
```
rebeccafair, line 51: savefig. I should probably think about adding doc code snippet tests at some point...

doc/source/plotting.rst
```diff
@@ -59,7 +81,8 @@ Docstring
 2D Plots
 ========
 
-2D spectra are plotted with :py:meth:`euphonic.plot.plot_2d`:
+2D spectra are arranged in a matplotlibFigure with
```
rebeccafair, line 84: matplotlib Figure

euphonic/plot.py
```diff
@@ -1,50 +1,22 @@
 from typing import List, Optional, Sequence, Tuple, Union
-import warnings
 
 try:
-    import matplotlib as mpl
```
rebeccafair, line 5: I think the removal of this line might be causing the test to fail, changing `name` in the mocked import. Maybe in `mocked_import` using `if 'matplotlib' in name` instead of `if name == 'matplotlib'` might fix it?

rebeccafair (changes requested):

doc/source/plotting.rst
ajjackson: done!

doc/source/plotting.rst
ajjackson: Fixed, good spot

doc/source/plotting.rst
ajjackson: Done

euphonic/plot.py
ajjackson: That's it! Have tweaked the mocker to function more robustly.

PR #119, opened 2021-02-05
Plotting spectra wrapup

rebeccafair:
Docs should be visible [here ](https://euphonic.readthedocs.io/en/plotting-spectra-wrapup/) for proofreading once they've built

Change summary, taken from https://github.com/pace-neutrons/Euphonic/pull/89#issuecomment-768241510
- [x] Decide whether to make `_get_bin_edges/centres` public. If so:
  - [x]  More foolproofing w.r.t setting `bin_ax/data_ax`
  - [x] Docstrings, in particular describing any limitations for differently sized bins. I don't know if we need to write anything specific in the docs, they should be self-explanatory...
  - [x] Tests
- [x] Decide whether to make public `_plot_1d/2d_core` and what to call them. I think we decided on `plot_1d/2d_to_axis` or similar? If so:
  - [x] Use cases/examples in the plotting docs.
- [x] Add some explanation of `Spectrum1DCollection` to the spectra docs.
- [x] Final check of docs for other things I haven't thought of yet
- [x] Changelog update!

Additional changes made:
* Euphonic Jupyter notebook tutorials removed. As Euphonic's functionality has increased, maintaining these has become more cumbersome, there are no tests for them, and there is no expectation or requirement for Euphonic to have tutorials in a Jupyter notebook format anyway. Rather than update them to use the new plotting API, they have been removed to reduce maintenance overhead.
* `doc/source/refactor.rst` has been removed. This document was to describe the major refactoring changes from Euphonic `0.2` -> `0.3`, and as with these major spectrum changes the next release will be >`0.3`, this file is no longer relevant and has been removed. It can still be accessed in the `0.3.x` documentation.


ajjackson (approved):
Good stuff, in general I see more API docs and less code :smile: 

A few small tweaks and comments but no show-stoppers.

Something not mentioned in the PR description is the removal of tutorials and refactor.rst. I'm not really bothered by this but maybe something should be on the record about how/why this is redundant/replaced?

rebeccafair:
> Something not mentioned in the PR description is the removal of tutorials and refactor.rst. I'm not really bothered by this but maybe something should be on the record about how/why this is redundant/replaced?

Good idea, added to the description

Issue #122, opened 2021-02-10
Create QpointFrequencies object


rebeccafair:
As a result of thinking about #117, its been realised that as both `QpointPhononModes` and `StructureFactor` contain `qpts` and `frequencies`, they should probably share a parent class, `QpointFrequencies` (note no `Phonon`, as there are no attributes relating this object to phonons specifically). This has several advantages:

- Both `QpointPhononModes` and `StructureFactor` can share functions that only use q-points and frequencies e.g. `get_dispersion` (addressing #117), `calculate_dos`
- This will allow more flexibility in reading output files that don't contain eigenvectors, as they are not required to calculate dispersion or dos.
- This will allow more flexibility in reading output files that only contain cell vectors, and not further structure information such as atom coordinates or species, as this detailed structure information is not required to calculate dispersion or DOS. This could be implemented with a `Crystal` object with zero-length `atom_r`, `atom_mass` and `atom_type`

PR #126, opened 2021-02-15
122 qpt frequencies

### LINKED ISSUE #117: Enhancement: provide get_dispersion() method on StructureFactor as well as QpointPhononModes
Arising from discussion on another PR, it would be handy if `get_dispersion()` was available with a somewhat consistent API on StructureFactor; currently it is only on QpointPhononModes and it could be a bit hacky to bring them together. https://github.com/pace-neutrons/Euphonic/pull/115#discussion_r561751027

rebeccafair:
Addresses #122, also will close #117

This ended up being a little more complicated than I expected, but hopefully this covers it. Summary of main changes:

- New `QpointFrequencies` object which doesn't contain eigenvectors
- Reading of CASTEP and Phonopy files without allocating/reading eigenvectors, reducing memory usage
- Calculation of `QpointFrequencies` from `ForceConstants` without allocating the eigenvector array, reducing memory usage
- `QpointPhononModes` and `StructureFactor` now inherit from `QpointFrequencies`. This has a number of effects:
  - ``StructureFactor`` now has a `weights` attribute
  - ``StructureFactor`` now has `get_dispersion` and `calculate_dos` functions
  - Less duplicated code
- Improved testing of `StructureFactor.calculate_1d_average` as previously the weights argument wasn't explicitly tested. Seemed a good idea to do this while adding tests for the `StructureFactor.weights` attribute
- `Crystal` can now be created from just cell vectors (`Crystal.from_cell_vectors`)
- Moved CASTEP/Phonopy test files to their own directories. Now that both `QpointPhononModes` and `QpointFrequencies` both read from `.phonon` files, for example, it didn't really make sense for them to still live in the `qpoint_phonon_modes` subdirectory.
- Added `get_dispersion` tests
- Updated/consistent docstrings with typing for `QpointFrequencies`, `QpointPhononModes` and `StructureFactor`. Other classes should probably be updated for consistency at some point...
- Tests for the above
- Updated docs for the above: [here](https://euphonic.readthedocs.io/en/122_qpt_frequencies/)


euphonic/qpoint_frequencies.py
```diff
@@ -0,0 +1,240 @@
...
+        Parameters
+        ----------
+        crystal
+            Lattice and atom information
+        qpts
+            Shape (n_qpts, 3) float ndarray. Q-point coordinates
+        frequencies
+            Shape (n_qpts, n_branches) float Quantity. Frequencies,
+            ordered according to increasing q-point number. Default
+            units meV
```
ajjackson, line 51: What does it mean for a Quantity input to have 'default units meV'? Won't it be whatever units are given?

euphonic/qpoint_frequencies.py
```diff
@@ -0,0 +1,240 @@
...
+        _obj_to_json_file(self, filename)
+
+    @classmethod
+    def from_dict(cls: T, d: Dict[str, Any]) -> T:
+        """
+        Convert a dictionary to a QpointFrequencies object
+
+        Parameters
+        ----------
+        d
```
ajjackson, line 169: Generally it's a good idea to avoid single-letter variable names because they aren't "greppable". It doesn't do much harm here though, as the function is only three lines long!

euphonic/qpoint_frequencies.py
```diff
@@ -0,0 +1,240 @@
...
+    def from_castep(cls: T, filename: str) -> T:
+        """
+        Reads precalculated phonon mode data from a CASTEP .phonon file
+
+        Parameters
+        ----------
+        filename
+            The path and name of the .phonon file to read
+        """
+        data = castep._read_phonon_data(filename, read_eigenvectors=False)
```
ajjackson, line 209: Why is this method private? Isn't it the main purpose of the submodule?

euphonic/qpoint_frequencies.py
```diff
@@ -0,0 +1,240 @@
...
+        phonon_format
+            Format of the phonon_name file if it isn't obvious from the
+            phonon_name extension, one of {'yaml', 'hdf5'}
+        summary_name
+            Name of Phonopy summary file to read the crystal information
+            from. Crystal information in the phonon_name file takes
+            priority, but if it isn't present, crystal information is
+            read from summary_name instead
+        """
+        data = phonopy._read_phonon_data(
```
ajjackson, line 237: Similar to castep module, maybe we should make this public.

euphonic/qpoint_phonon_modes.py
```diff
@@ -1,57 +1,63 @@
...
+            frequencies, ordered according to increasing q-point number.
+            Default units meV
+        eigenvectors
+            Shape (n_qpts, 3*crystal.n_atoms, crystal.n_atoms, 3)
+            complex ndarray. The dynamical matrix eigenvectors
+        weights
+            Shape (n_qpts,) float ndarray. The weight for each q-point.
+            If None, equal weights are assumed
         """
         _check_constructor_inputs(
```
ajjackson, line 62: Could most of this `__init__` be replaced with `super().__init__(**args_except_for_eigenvectors)`?

euphonic/readers/castep.py
```diff
@@ -1,29 +1,46 @@
...
 
 import numpy as np
 
 from euphonic import ureg
 from euphonic.util import is_gamma
 
 
-def _read_phonon_data(filename, cell_vectors_unit='angstrom',
-                      atom_mass_unit='amu', frequencies_unit='meV'):
+def _read_phonon_data(filename: str,
```
ajjackson, line 11: This seems public, maybe drop the underscore? The name is a bit generic but in the context of `euphonic.readers.castep` it's fairly clear that we're reading CASTEP files so I guess that's fine.

euphonic/readers/phonopy.py
```diff
@@ -105,24 +114,28 @@ def _extract_phonon_data_yaml(filename):
             [atom['symbol'] for atom in phonon_data['points']])
     except KeyError:
         pass
-
     return data_dict
 
 
-def _extract_phonon_data_hdf5(filename):
+def _extract_phonon_data_hdf5(filename: str,
+                              read_eigenvectors: Optional[bool] = True
```
ajjackson, line 121: `Optional[bool]` means that the value can be `True`, `False` or `None` but I don't think that's correct? For a `bool` with a default value I think we just want `read_eigenvectors: bool = True`.

Or we could do a similar thing to `use_c` and have `None` mean "try" and `True` raise an error if eigenvectors unavailable. I don't think it's really necessary; it's easy enough to check `if data.get('eigenvectors')`.

euphonic/readers/phonopy.py
```diff
@@ -181,36 +195,55 @@ def _extract_band_data_hdf5(hdf5_file):
 
 
 def _read_phonon_data(
-        path='.', phonon_name='band.yaml', phonon_format=None,
-        summary_name='phonopy.yaml', cell_vectors_unit='angstrom',
-        atom_mass_unit='amu', frequencies_unit='meV'):
+        path: Optional[str] = '.', phonon_name: Optional[str] = 'band.yaml',
```
ajjackson, line 198: Missing a newline between these args

ajjackson, line 198: Again, I'm not sure that all of these should have the `Optional` type, which is really a shorthand for `Union[None, type]`. Arguments with default values can just be set as e.g. `path: str = '.'`.

euphonic/structure_factor.py
```diff
@@ -31,43 +34,51 @@ class StructureFactor(object):
...
     temperature : float Quantity or None
         The temperature used to calculate any temperature-dependent
         parts of the structure factor (e.g. Debye-Waller, Bose
         population factor). None if no temperature-dependent effects
         have been applied
     """
-
-    def __init__(self, crystal, qpts, frequencies, structure_factors,
-                 temperature=None):
+    def __init__(self, crystal: Crystal, qpts: np.ndarray,
```
ajjackson, line 45: I think this `__init__()` should be using `__super__()__init__(*some_args)` to avoid redundancy.

euphonic/structure_factor.py
```diff
@@ -121,49 +133,59 @@ def calculate_1d_average(self,
...
             e_bins, calc_bose=calc_bose, temperature=temperature)
 
         spectrum = np.average(
             sqw_map.magnitude, axis=0, weights=weights)*sqw_map.units
         return Spectrum1D(e_bins, spectrum)
 
-    def calculate_sqw_map(self, e_bins, calc_bose=True, temperature=None):
+    def calculate_sqw_map(self,
+                          e_bins: Quantity,
+                          calc_bose: Optional[bool] = True,
```
ajjackson, line 168: Not `Optional` as `None` is not handled

ajjackson (changes requested):
You've done a lot of work here! Lots of cleanup and improved documentation. The file moves seem sensible and clearly I support adding QpointFrequencies as discussed. It's great to see type hinting rolled out over more of the codebase, it really does help with readability and should also help us catch some bugs.

Have made a few comments and niggles here and there, but the two main concerns are:

- We should be using `super()` to avoid redundancy in init methods of derived classes. I don't think this will conflict with the argument-checking framework?
- I noticed about 3/4 of the way through reviewing that some arguments were given the `Optional` type-hint inappropriately; I highlighted a few with explanation, but didn't go back to the other files. All new uses of `Optional` should be checked.

I haven't looked at the updated tests yet, maybe that can be another session!

I also have not yet tried running `mypy` on the new type hints. This might find some bugs, it usually does :sweat_smile:

euphonic/qpoint_phonon_modes.py
```diff
@@ -1,57 +1,63 @@
...
-class QpointPhononModes(object):
+T = TypeVar('T', bound='QpointPhononModes')
+
+
+class QpointPhononModes(QpointFrequencies):
     """
     A class to read and store vibrational data from model (e.g. CASTEP)
     output files
 
     Attributes
```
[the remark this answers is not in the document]
rebeccafair: I've updated Sphinx so maybe that's changed things, this looks ok to me? https://euphonic.readthedocs.io/en/122_qpt_frequencies/qpoint-phonon-modes.html#docstring

euphonic/qpoint_frequencies.py
rebeccafair: This is probably left over from when Euphonic could only read from CASTEP files, to hint that the frequency units would be in meV. Removed now as its not generally applicable anymore.

euphonic/qpoint_frequencies.py
rebeccafair: Its sort of hinting to users "don't use this, use Euphonic objects please", but since it is being called from elsewhere it should probably be public. Have made it public now.

euphonic/readers/phonopy.py
rebeccafair: Agreed, have removed `Optional` in all places where it isn't applicable

euphonic/qpoint_phonon_modes.py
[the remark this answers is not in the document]
ajjackson: This is how it looks for me

Screenshot from 2021-02-23 11-32-11

ajjackson (approved):

ajjackson:
Looks good! I'd just like to run mypy on this before giving the all-clear

ajjackson:
Right. Basically mypy is a disaster, but that's not really a fault in this PR; we have some problem-solving to do to get it playing nicely with pint and numpy. Already the new type-hinting is very nice for readability/documentation. I've no objection to merging this.

Issue #128, opened 2021-02-17
Improve parallelism settings


rebeccafair:
Currently by default `calculate_qpoint_phonon_modes` has `use_c=False` and `n_threads=1`. This means that by default it uses serial Python (even if the C extension is installed) which is undesirable, and means that to make the most of Euphonic's performance `use_c` and `n_threads` have to be passed all the way through the call stack. This is a particular problem for command line scripts, which are supposed to be quick and easy to use, having to pass these arguments all the time would be cumbersome.

To make this more user friendly, a few changes are proposed:
* `use_c` should be retained as it is good to give users full control if they want it, and it is also useful for testing, but with some changes.
  * `use_c=True` should force use of the C extension, and raise an exception if it isn't installed
  * `use_c` should have a new default `None` which acts as 'auto' and will use the C extension if it is installed, and quietly fall back to Python if it isn't installed
* `n_threads` should also be retained as this fine control could be useful for e.g. running multiple Euphonic instances at once, but with a new default `None`. If `n_threads=None`, the environment variable `EUPHONIC_NUM_THREADS` will be read, and if the environment variable isn't set, `n_threads` will be set to <s>1</s> the output of `multiprocessing.cpu_count()`. This allows users to either set the environment variable once, then not have to worry about the number of threads, or get a sensible default. This is particularly useful for command line tools.

ajjackson:
This all looks good to me.

Might be worth discussing whether the fallback default should indeed be 1; another option would be
```python
import multiprocessing
default_n_threads = multiprocessing.cpu_count()
```
On my i7 that returns 8, so it doesn't distinguish between true cores and hyper-threading. But I get better performance on `calculate_qpoint_phonon_modes` with hyperthreading so that's not a bad thing!

rebeccafair:
That's a good point, I do worry that occasionally `cpu_count()` will give a count that is too high and decrease performance, but I haven't actually seen this much in practice, some quick profiling on my laptop (`cpu_count() = 4`) and IDAaaS (`cpu_count() = 30`) shows there is still improvement. There are probably far more cases where it provides a benefit, so I'll use `cpu_count()` as the default.

ajjackson:
If we do introduce some serial chunking to limit memory usage during DOS/powder calculations, then I suggest we target 1GB/thread or less as workstations generally have a bit over 1GB/thread of memory.

rebeccafair:
[...]

PR #129, opened 2021-02-25
Improve Parallelism Defaults

rebeccafair:
Addresses #128


euphonic/force_constants.py
```diff
@@ -445,19 +449,28 @@ def _calculate_phonons_at_qpts(
...
-                        'have been installed. Attempting to fall back '
-                        'to pure Python calculation'), stacklevel=2)
+                    if use_c is None:
+                        warnings.warn((
+                            cext_err_msg
+                            + ' Falling back to pure Python calculation.'),
+                            stacklevel=3)
                     raise
             else:
                 raise ImportError
```
ajjackson, line 468: I do find this stack of if functions and try/Except a bit tricky to follow. I'm also wary also of using try/except/raise as a form of control logic over a large block of code. What happens if something other than `import euphonic._euphonic` raises an ImportError? Then it shouldn't be excepted. I think this was motivated by the problem of reaching the python implementation from different paths. How about instead we put that in a function and have something like:

```python
if use_c is not False:
    try:
        import euphonic._euphonic as euphonic_c
    except ImportError:
        if use_c:
            raise
        else:
            warnings.warn(...)
            _do_it_with_python(*args)
    else:
        _do_it_with_c(*args)
else:
    _do_it_with_python(*args)
```

ajjackson (changes requested):
I agree with the approach here. The new tests look great (and taught me something neat about testing warnings!)

The threadctl context manager is very cool, and a non-intrusive way of dealing with this problem.

I just raised one concern about the try/except logic which is a bit hard to follow and also could be safer.

euphonic/force_constants.py
rebeccafair: Ok done! I've implemented it a little differently, as I found handling the different output arguments for Python implementation (depending on whether `return_eigenvectors=True/False`) was kind of awkward. But I think it works!

euphonic/force_constants.py
```diff
@@ -453,19 +453,23 @@ def _calculate_phonons_at_qpts(
...
+                        cext_err_msg
+                        + ' Falling back to pure Python calculation.'),
+                        stacklevel=3)
+                else:
+                    raise ImportCError(cext_err_msg)
+
+        if use_c_status is True:
             if n_threads is None:
                 try:
                     n_threads = int(os.environ['EUPHONIC_NUM_THREADS'])
```
ajjackson, line 475: I think this will raise a ValueError if the user has set `EUPHONIC_NUM_THREADS=''`; setting to '' is a common way of pseudo-clearing an environment variable. (It will return the same thing to bash as if the variable was unset.) It will also ValueError on other invalid values (e.g. `'TWO'`) but in that case raising the error is probably best.

As well as accepting `''`, we can avoid the try/except for KeyError with `os.environ.get()`:
```python
n_threads_env = os.environ.get('EUPHONIC_NUM_THREADS', '')   # type: str
if n_threads_env:
    n_threads = int(n_threads_env)
else:
    n_threads = cpu_count()
```

euphonic/force_constants.py
```diff
@@ -453,19 +453,23 @@ def _calculate_phonons_at_qpts(
...
-                        warnings.warn((
-                            cext_err_msg
-                            + ' Falling back to pure Python calculation.'),
-                            stacklevel=3)
-                    raise
-            else:
-                raise ImportError
+
+        # Check if C extension can be used and handle appropriately
+        use_c_status = False
```
ajjackson, line 458: Good call. Using a flag like this isn't as "pretty" as the other approaches we considered, but in this case it's clearly the most readable.

ajjackson (approved):
Looks good! I flagged up one last tweak, sorry :sweat:

euphonic/force_constants.py
rebeccafair: Good call, I didn't think of this case

Issue #118, opened 2021-01-27
Enable adaptive broadening for DOS


rebeccafair:
Adaptive broadening allows a more realistic DOS to be produced with fewer q-points, and ensures features aren't either exaggerated by the choice of q-points, or masked by a large broadening width. The basic premise is that the mode gradients are used to apply a broadening width to each mode at each q-point, and these are broadened individually and added to the DOS. Flat modes will have low gradients, smaller broadening widths and contribute sharp peaks to the DOS, whereas rapidly changing modes will contribute flatter, broader peaks.

See http://dx.doi.org/10.1103/PhysRevB.75.195121 section D for details on the method.

Issue #116, opened 2021-01-26
DOS-neutron weighted


davidvoneshen:
We have a routine to compute the DOS but not the neutron-weighted DOS which, within the incoherent approximation is what we measure. 

The neutron weighted DOS is 

> 
![pdos](https://user-images.githubusercontent.com/48722944/105860841-eccdec80-5fe5-11eb-87f8-0b29b90513ba.png)

where A is some scaling factor, b_j is the neutron scattering length (it is probably best to subsitute b_j^2 for the total scattering cross section) of the element/isotope, m_j is the mass and PDOS_j_ is its contribution to the DOS.

This is most typically what you see in papers comparing powder data to calculation. When we have proper powder averaging with resolution convolution it will be more accurate to compute that and then integrate it but this is quick, easy and surprisingly good most of the time.

rebeccafair:
Would it make sense to have the option of doing this with both coherent and incoherent cross sections? If so this could be implemented as a keyword argument on `calculate_dos` e.g. `calculate_dos(dos_bins, weighting='coherent')`. Is the fact that it's a neutron scattering length sufficiently obvious in Euphonic's context or should it be more specific like `weighting='coherent-neutron'`? Are there any other types of DOS weighting that we could want to implement in the future?

ajjackson:
Natural thing would be to have options "coherent", "incoherent", and "total", with the default being `None`.

davidvoneshen:
People do measure this with x-ray techniques. The thing is that is more complex to model (you need a form factor like the magnetic one which depends on Q and varies from element to element) and I would think you would want to compute the spherical average for the Q you did the measurement at.

Fundamentally though I like it as a keyword for the DOS routine that is already there with the default as None. I don't have strong feelings about how general it should be. It is sometimes nice to be able to edit the scattering lengths to simulate doping and see if anything interesting is happening or if changes are just down to the cross sections. I guess if you manually specified the weighting for each element that would be the most general approach. This would be like it currently is in the horace interface where I need to give the individual weightings. I don't think this is an unreasonable burden on the user and is the most flexible.

ajjackson:
What we have done for "proper" coherent weighting is include an argument to select the data source. So if the user wants to pull the cross-sections from a customised data file, they can :smile: . I expect we can use the same system here.

rebeccafair:
[...]

PR #131, opened 2021-03-17
Add adaptive broadening

rebeccafair:
Addresses #118
 
**Initial changes:**
* A new argument `return_mode_widths` for `calculate_qpoint_phonon_modes/frequencies` which returns the mode widths rather than just the phonon object
* A new argument `mode_widths` to `calculate_dos` which causes adaptive broadening to be used rather than the histogram binning
* Another new argument `are_bin_edges` to `calculate_dos` which allows use of either bin centres or edges (previously only edges were used). This is because the histogram binning uses bin edges directly, whereas the adaptive broadening implementation `scipy.stats.norm.pdf` calculates at specific points, so in this case bin centres make more sense.
* In `np.histogram` in `calculate_dos` I've used a new argument `density=True`, which has changed all the DOS results by a scale factor (I've checked this and updated the test data). The reason for doing this is that the results from `np.histogram` (no broadening) are now on the same scale as those from `norm.pdf` (adaptive broadening) so they are more easily comparable. This can be seen in the example below. This may change the results for differently sized bins but I don't think we've clearly defined that behaviour yet (and at least we're not testing it) so this should be ok?
* Euphonic can now read CASTEP `.phonon_dos` files via `Spectrum1D.from_castep_phonon_dos`. I had to write a file reader for this anyway to check Euphonic's results so thought I might as well include it... It currently only reads the full DOS and not PDOS, but maybe that's for #116 
* Added relevant tests and documentation

**Main thoughts/concerns:**
* I'd like feedback on whether having the `return_mode_widths` argument and returning them separately is a good way to do it or we should have a new `mode_widths` attribute? But I'm not sure if its generally applicable enough for that yet. _Resolved: is an appropriate solution for now with a single use-case_
* `scipy.stats.norm.pdf` is very slow. If this did become an issue we could probably get around this by writing our own Gaussian generator (in C?). _Resolved: Will be addressed in another PR https://github.com/pace-neutrons/Euphonic/issues/135_
* Calculating the mode widths can be slow, up to 10 times slower than just calculating the eigenvalues/vectors in C. Luckily this has no effect on `return_mode_widths=False` behaviour, and `pdf` is definitely the bottleneck, but this may be something that wants optimising in the future. _Resolved: Will be addressed in the future as/when needed_
* Is `are_bin_edges` required or should we just make a note that you should use bin edges unless `mode_widths` is specified, in which case bin centres are recommended? _Resolved: `are_bin_edges` has been removed and `calculate_dos` will always require bin edges, and will quietly convert to bin centres if required_
*  Does `Spectrum1D.from_castep_phonon_dos` belong and is this name ok? I still need to add some documentation for this. _Resolved: Fine as is_
* I haven't added the adaptive broadening as an option for the dos cli tool, should I? _Resolved: Done_
* There may be some merit in applying this for broadening 2D intensity maps, but I think that's probably in the far future. _Resolved: Left for future work_

Examples:

Comparison between no broadening, 1meV fixed broadening and adaptive broadening:
image
The same but without the 'no broadening' line to make things clearer
image
Zoomed version of the above to show some of the more subtle differences between fixed/adaptive broadening:
image

Other issues that will be addressed in another PR, so that this can be closed out more quickly:
- Getting proper scaling of DOS. Using `density=True` is not completely correct, but this can be resolved during  #116 
- Optimisation of application of adaptive broadening #135
- Addition of adaptive broadening to powder map #139
- Finalising maths in description of mode widths docstring #143 

Still to do for this PR:

- [x] Add user documentation for `from_castep_phonon_dos`
- [x] Update changelog


rebeccafair:
@krefson Could you take a look at the method I've written for the mode gradients calculation and make sure its correct? Its in the 'Notes' section of the docstring here: https://euphonic.readthedocs.io/en/118_adaptive_broadening/force-constants.html#euphonic.force_constants.ForceConstants.calculate_qpoint_phonon_modes

rebeccafair:
I've made some updates after the addition of metadata, allowing reading of either the total DOS or a specfic PDOS from a CASTEP .phonon_dos with with `Spectrum1D.from_castep_phonon_dos`, and added a `Spectrum1DCollection.from_castep_phonon_dos` method to read all of them. This gives an example of how PDOS might look in #116.

The main thoughts/concerns raised in the original post still apply and now might be a good time for a review before I start finalising things @ajjackson

ajjackson:
Ok, some thoughts on stuff raised in the post, before I look at the code. Maybe you already did some of this!

> * In `np.histogram` in `calculate_dos` I've used a new argument `density=True`, which has changed all the DOS results by a scale factor (I've checked this and updated the test data). The reason for doing this is that the results from `np.histogram` (no broadening) are now on the same scale as those from `norm.pdf` (adaptive broadening) so they are more easily comparable. This can be seen in the example below. This may change the results for differently sized bins but I don't think we've clearly defined that behaviour yet (and at least we're not testing it) so this should be ok?

This scares me a little bit; in other projects where I have been sloppy about histogram/spectrum normalisation it always came back to bite me. Firstly, broadening should not change the area under the curve if it is implemented correctly. Then the question is how to weight the initial spectrum. The easy answer is to put the DOS weights (which sum to 3N or some other meaningful number - @krefson ?) directly into histogram bins, but this gives a bin-width-dependence (and the wrong area) unless the bin widths are accounted for. I believe we should be putting an _area_ equal to the DOS integral (i.e. 3N or whatever) into the histogram.

I think `density=True` normalises the area to one. That's a fair choice in its own right and can be scaled back up to 3N easily enough if that's your preferred convention. However, that will give incorrect results if the histogram range was already limited to not contain all the data - then we have normalised some subset of the data with a mystery scale factor. So if we do normalise to 1 I'm not sure that `density=True` is the responsible way to do it.


> * I'd like feedback on whether having the `return_mode_widths` argument and returning them separately is a good way to do it or we should have a new `mode_widths` attribute? But I'm not sure if its generally applicable enough for that yet.

We could have a `mode_widths` attribute on QPointFrequencies/PhononModes, or if we are being very OO-purist we could define child classes for this. I think the bar to clear here is "do we ever want to pass the object on with that data?" I don't see an immediate use-case for serialising with the mode widths or passing to an external plotter; we currently expect to _immediately_ take our QPointFrequencies and widths and use them to make a Spectrum1D(collection). I don't see the harm in keeping an array around for a few lines of code. Let's call this one a YAGNI, it's easy enough to change later.

I'm not _wild_ about function arguments that change the number of returned values (makes testing and docs clunky), but this does seem like a case where it's justified :sweat_smile: And it will be easy to quietly stop using and deprecate if we _do_ ever add a combined object.

> * `scipy.stats.norm.pdf` is very slow. If this did become an issue we could probably get around this by writing our own Gaussian generator (in C?)
> 
> * Calculating the mode widths can be slow, up to 10 times slower than just calculating the eigenvalues/vectors in C. Luckily this has no effect on `return_mode_widths=False` behaviour, and `pdf` is definitely the bottleneck, but this may be something that wants optimising in the future

It could be worth shooting-out some different Gaussian generators, but as discussed I have ideas about how to drastically cut down the number of evaluations! Happy for the initial implementation to be a bit slow, the main thing is having an accurate reference before we tinker further.

> * Is `are_bin_edges` required or should we just make a note that you should use bin edges unless `mode_widths` is specified, in which case bin centres are recommended?

Gonna go with the "Zen of Python" here: "explicit is better than implicit". FWIW I think it is better to pass bin edges in nearly all cases because then there is no guesswork about the data range (i.e. the outer bin edges.) We just need to avoid computing bin centres inside loops!

When working on Abins I did actually play with a scheme that replaces "evaluate the Gaussian at each bin centre" with "compute the area under the Gaussian in each bin" (using `erf`). Performance-wise it didn't make much difference, but on coarse grids it looks different which might upset people. I'd argue it's the "more correct" broadening implementation though as it gets the area right :grin: 

I would note that `are_bin_edges` reads oddly, it looks a bit like a question? Is that following a pattern from elsewhere or would `use_bin_edges` be better? Or we could replace `frequencies=` with mutually-exclusive `bin_edges=` and `bin_centres=` arguments?

> * Does `Spectrum1D.from_castep_phonon_dos` belong and is this name ok? I still need to add some documentation for this.

ForceConstants already provides a `from_castep` so there's a precedent. If we add a bunch more formats then maybe think about cleanup but at this point it's useful and not too cumbersome.
 
> * I haven't added the adaptive broadening as an option for the dos cli tool, should I?

Yes please!
 
> * There may be some merit in applying this for broadening 2D intensity maps, but I think that's probably in the far future...

There's a motive to return and make it run fast :smile:

rebeccafair:
> > * In `np.histogram` in `calculate_dos` I've used a new argument `density=True`, which has changed all the DOS results by a scale factor (I've checked this and updated the test data). The reason for doing this is that the results from `np.histogram` (no broadening) are now on the same scale as those from `norm.pdf` (adaptive broadening) so they are more easily comparable. This can be seen in the example below. This may change the results for differently sized bins but I don't think we've clearly defined that behaviour yet (and at least we're not testing it) so this should be ok?
> 
> This scares me a little bit; in other projects where I have been sloppy about histogram/spectrum normalisation it always came back to bite me. Firstly, broadening should not change the area under the curve if it is implemented correctly. Then the question is how to weight the initial spectrum. The easy answer is to put the DOS weights (which sum to 3N or some other meaningful number - @krefson ?) directly into histogram bins, but this gives a bin-width-dependence (and the wrong area) unless the bin widths are accounted for. I believe we should be putting an _area_ equal to the DOS integral (i.e. 3N or whatever) into the histogram.
> 
> I think `density=True` normalises the area to one. That's a fair choice in its own right and can be scaled back up to 3N easily enough if that's your preferred convention. However, that will give incorrect results if the histogram range was already limited to not contain all the data - then we have normalised some subset of the data with a mystery scale factor. So if we do normalise to 1 I'm not sure that `density=True` is the responsible way to do it.

That's a fair point about values outside the bin range, I guess we could try and correct for it by calculating a scale factor based on the weights of the modes outside the range? But wouldn't that be a problem of any method, not just `density=True`? There's also the issue of different units. The DOS values have been computed with the energy bins in Hartree, which is why they are so high. Again I guess I could scale them based on the units used, but once the `Spectrum1D` object is produced, people could change the energy bin units to whatever, meaning the DOS area will no longer be 1. I feel like trying to keep to a specific area under the DOS might be a fruitless battle, maybe the most important things are making sure we're consistent about handling different bin widths and making sure broadening doesn't change the area (which I think at least `density=True` does better at than before)... But I'm not really sure how to proceed with this. 
 
> > * Is `are_bin_edges` required or should we just make a note that you should use bin edges unless `mode_widths` is specified, in which case bin centres are recommended?
> 
> Gonna go with the "Zen of Python" here: "explicit is better than implicit". FWIW I think it is better to pass bin edges in nearly all cases because then there is no guesswork about the data range (i.e. the outer bin edges.) We just need to avoid computing bin centres inside loops!
> 
> When working on Abins I did actually play with a scheme that replaces "evaluate the Gaussian at each bin centre" with "compute the area under the Gaussian in each bin" (using `erf`). Performance-wise it didn't make much difference, but on coarse grids it looks different which might upset people. I'd argue it's the "more correct" broadening implementation though as it gets the area right 😁
> 
> I would note that `are_bin_edges` reads oddly, it looks a bit like a question? Is that following a pattern from elsewhere or would `use_bin_edges` be better? Or we could replace `frequencies=` with mutually-exclusive `bin_edges=` and `bin_centres=` arguments?

I agree bin edges are generally better, I thought about adding the bin centres option because the bin values in the CASTEP `.phonon_dos` file are centres, and `scipy.stats.norm.pdf` also requires centres. But that could change if we change the implementation. Maybe I should remove the option altogether, always require bin edges, then quietly convert to bin centres if required? Although something about that makes me uneasy...

> > * I haven't added the adaptive broadening as an option for the dos cli tool, should I?
>
> Yes please!

Will do!

ajjackson:
Somehow I edited your comment rather than reply! Sorry, will see if I can figure out how to roll that back...

> That's a fair point about values outside the bin range, I guess we could try and correct for it by calculating a scale factor based on the weights of the modes outside the range? But wouldn't that be a problem of any method, not just `density=True`?

It's a problem of any method that loses track of the "true" values. I think the answer is to normalise while we still have that information.

> There's also the issue of different units. The DOS values have been computed with the energy bins in Hartree, which is why they are so high. Again I guess I could scale them based on the units used, but once the `Spectrum1D` object is produced, people could change the energy bin units to whatever, meaning the DOS area will no longer be 1. 

There's a distinction between conserving the area the same and keeping it the same _number_. The unit of DOS are states/volume/energy. Say our DOS units are states/BZ/Ha; if we integrate the DOS over all energy we get 3N states per BZ. If the x-axis units are changed to eV, the integral doesn't change. But if we numerically integrate a value in states/BZ/Ha over energy values in eV then we'll get an answer in states*eV/BZ/Ha; the units still need to be cleaned up.

> I agree bin edges are generally better, I thought about adding the bin centres option because the bin values in the CASTEP `.phonon_dos` file are centres, and `scipy.stats.norm.pdf` also requires centres. But that could change if we change the implementation. Maybe I should remove the option altogether, always require bin edges, then quietly convert to bin centres if required? Although something about that makes me uneasy...

It makes me less uneasy than discarding bin edges does. But yes, the implementations of things will usually deal with centres. Maybe some kind of caching can be used?

rebeccafair:
> It's a problem of any method that loses track of the "true" values. I think the answer is to normalise while we still have that information.
> 
Ok, I'll try and implement something...

> > There's also the issue of different units. The DOS values have been computed with the energy bins in Hartree, which is why they are so high. Again I guess I could scale them based on the units used, but once the `Spectrum1D` object is produced, people could change the energy bin units to whatever, meaning the DOS area will no longer be 1.
> 
> There's a distinction between conserving the area and keeping it the same _number_. The unit of DOS are states/volume/energy. Say our DOS units are states/BZ/Ha; if we integrate the DOS over all energy we get 3N states per BZ. If the x-axis units are changed to eV, the integral doesn't change. But if we numerically integrate a value in states/BZ/Ha over energy values in eV then we'll get an answer in states*eV/BZ/Ha; the units still need to be cleaned up.
> 
> (Using BZ as a lazy shorthand for "reciprocal cell volume", which is not an accurate use of the term!)
> 

So what units should DOS be returned in? (From a `pint` point of view) 1/energy? Currently its just 'dimensionless' which is a bit of a cop out

> > I agree bin edges are generally better, I thought about adding the bin centres option because the bin values in the CASTEP `.phonon_dos` file are centres, and `scipy.stats.norm.pdf` also requires centres. But that could change if we change the implementation. Maybe I should remove the option altogether, always require bin edges, then quietly convert to bin centres if required? Although something about that makes me uneasy...
> 
> It makes me less uneasy than discarding bin edges does. But yes, the implementations of things will usually deal with centres. Maybe some kind of caching can be used?

That's fair, I'll remove the option to input bin centres. I imagine calculating the bin centres is far from being the bottleneck at the moment, but we can think about including some caching in the future if it does become an issue.

euphonic/spectra.py
```diff
@@ -345,6 +346,27 @@ def from_dict(cls: S1D, d: Dict[str, Any]) -> S1D:
...
+            The path and name of the .phonon_dos file to read
+        element
+            Which element's PDOS to read. If not supplied reads
+            the total DOS.
+        """
+        data = read_phonon_dos_data(filename)
+        if element is None:
+            element = 'Total'
+        return Spectrum1D(data['dos_bins']*ureg(data['dos_bins_unit']),
+                          data['dos'][element]*ureg('dimensionless'),
```
ajjackson, line 367: Are we aiming to sort out DOS dimensions in this PR or do it afterwards?

tests_and_analysis/test/script_tests/test_dos.py
```diff
@@ -79,16 +81,36 @@ def test_invalid_file_raises_value_error(self, dos_args):
...
+            euphonic.cli.dos.main([quartz_fc_file, '--adaptive',
+                                   '--shape', 'lorentz'])
+
 
 @patch('matplotlib.pyplot.show')
 @pytest.mark.skip(reason='Only run if you want to regenerate the test data')
 def test_regenerate_dos_data(_):
-    json_data = {}
+    # Read from existing file first to allow option of only replacing for
+    # certain test cases or keys
```
ajjackson, line 94: Presumably that has to be accompanied by commenting out a bunch of cases? Still, seems a good idea! Better than picking through the JSON diff to find the _intended_ changes.

ajjackson (commented):
Lots of code! All seems pretty sensible. The functionality looks cool but I still should give it a test-run before hitting "approve". In the meantime I made a few comments here and there.

We had some discussion elsewhere about whether the equations are clear enough; I might be able to give more useful feedback on that once I've read the referenced paper.

tests_and_analysis/test/script_tests/test_powder_map.py
```diff
@@ -112,16 +112,27 @@ def test_invalid_weights_raises_causes_exit(self, powder_map_args):
 @patch('matplotlib.pyplot.show')
 @pytest.mark.skip(reason='Only run if you want to regenerate the test data')
 def test_regenerate_powder_map_data(_):
```
ajjackson, line 114: I notice that the powder-map tests are being tweaked, but the adaptive broadening has not been added to this CLI tool. I think we should allow the option, even if it is tediously slow to run? (The test case can still be very small.)

euphonic/qpoint_frequencies.py
```diff
@@ -93,28 +94,45 @@ def calculate_dos(self, dos_bins: Quantity) -> Spectrum1D:
...
-        dos, _ = np.histogram(freqs, dos_bins, weights=weights)
+            dos_bins_calc = dos_bins.to('hartree').magnitude
+        if mode_widths is not None:
+            from scipy.stats import norm
+            dos_bins_calc = Spectrum1D._bin_edges_to_centres(dos_bins_calc)
+            dos = np.zeros(len(dos_bins_calc))
+            mode_widths = mode_widths.to('hartree').magnitude
+            # Set limit on very small widths of, say, 5e-7 hartree
+            # (~0.013meV) due to divergent behaviour
+            mode_widths = np.maximum(mode_widths, 5e-7)
```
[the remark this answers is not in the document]
rebeccafair: I'm not sure, there's a few places in Euphonic where there are magic numbers like this and I don't really know what to do with them... Do we want to be able to override them? I feel like don't really want to add an argument because 99% of the time you won't want to change it, but this specific case isn't used anywhere else so it wouldn't necessarily make sense to make it more global...

euphonic/readers/castep.py
```diff
@@ -91,7 +187,7 @@ def read_phonon_data(
                                      dtype=np.complex128)
                 for i in range(n_branches):
                     qeigenvec[i, :, :] = lines_i[i*n_atoms:(i+1)*n_atoms, :]
-            qpt_line = f.readline()
+            #qpt_line = f.readline()
```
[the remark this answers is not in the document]
rebeccafair: Good spot

euphonic/util.py
```diff
@@ -237,6 +238,28 @@ def custom_decode(dct):
             if isinstance(value, (float, complex))}
 
 
+def mode_gradients_to_widths(mode_gradients: Quantity, cell_vectors: Quantity
```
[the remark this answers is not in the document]
rebeccafair: This function logic is actually taken from the CASTEP implementation, as I wanted to get a close a comparison as possible to the CASTEP reference data. In CASTEP itself it only uses this version if the points have been supplied by `phonon_fine_kpoints_list` rather than `phonon_fine_mp_grid`, otherwise it uses a different estimation based on the grid.

Now that we know the adaptive broadening 'works', I don't mind using a different estimation. But in the end, I don't think the scaling of the mode widths matters too much, it depends really on how much broadening you want... I guess the important thing is this function gives a sensible estimate.

> I think it might be safer to determine the characteristic q-point distance before the DOS is calculated, and then pass that information along to where it is needed later

But it is? This is called in `force_constants.py` just before returning the mode widths. Unless it would be better to return the mode gradients from that calculation and let users do their own conversion (or use this function if they wish)? The main reason I moved this into its own function was because `.phonon_dos` includes the mode gradients, not mode widths, and I needed an easy way to convert for testing.

euphonic/spectra.py
rebeccafair: I should be putting up an initial PR for the neutron weighted DOS soon, it would be good to discuss that there, for reasons which will become obvious

tests_and_analysis/test/script_tests/test_powder_map.py
rebeccafair: Yep the main reason I didn't add it was because I thought it'd be slow, but I'll give it a go

euphonic/qpoint_frequencies.py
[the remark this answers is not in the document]
ajjackson: In this case, it could be an optional argument with a default value.

One approach would be what Abins does which is to define constants in one module and import them elsewhere. But that's more suited to constants that are used in multiple places. I'd prefer the kwarg approach in this particular case.

euphonic/util.py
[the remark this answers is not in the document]
ajjackson: I see. It wasn't entirely obvious where the function is used because it's in `util.py`. In that case maybe it is enough to state the assumption in the docstring: e.g. "The set of calculated mode_gradients should not be symmetry-reduced; the density of q-point sampling is estimated using the size of this array."

euphonic/spectra.py
ajjackson: Great!

tests_and_analysis/test/script_tests/test_powder_map.py
ajjackson: If that turns out to be messy it could become a separate PR?

rebeccafair: Yeah, since the DOS could change slightly in #116, maybe easier to leave it until after. Opened PR #139

euphonic/util.py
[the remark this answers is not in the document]
rebeccafair: Done!

euphonic/qpoint_frequencies.py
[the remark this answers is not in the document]
rebeccafair: Kwarg added

rebeccafair:
Ok, I think everything's been closed off/kicked down the road and this PR is ready for final review. @ajjackson

ajjackson (approved):
Ok, I think that's ready!

Slightly unfortunate that we'll likely end up changing the DOS scaling in two consecutive releases, but if nobody was relying on it to be "correct" yet then it shouldn't be a big deal.

Issue #123, opened 2021-02-10
Add metadata to Spectrum1DCollection


rebeccafair:
Packaging the outputs of a PDOS calculation (#116) could be sensibly done by a `Spectrum1DCollection`, but information about which `y_data` relates to which atom/species currently cannot be stored. This could be fixed with a generic `metadata` attribute on `Spectrum1DCollection`, which contains a list of dictionaries the same length as `y_data`.

Issue #130, opened 2021-03-09
Data file outputs from CLI tools


ajjackson:
While using the work-in-progress powder-averaging CLI tool, it can be tedious to adjust the colourmap range or colour choices as it takes a while to recalculate the data each time.

It would be useful if this tool could also write out a Spectrum2D .json file containing the data, that can be rapidly re-plotted. I guess `euphonic-intensity-map` would actually be the logical choice for replotting this output. It would also be useful if some plotting parameters such as axis labels and plot title could be included as metadata in this file.

Would this kind of functionality be useful for some of the other CLI tools? For example, for saving a configuration of PDOS labels/title/colour-scheme that might need re-plotting to a different size for publication.

ajjackson:
We already have a .json format for Spectrum2D objects so would fit the existing ecosystem nicely without introducing a new format.

`pickle` is not suitable for file sharing / exchange due to security concerns. `npy` is a good option for compact array storage, but I don't think it really does metadata? One can always use .json with file compression if file size is a concern; the benefit of .npy is faster data loading. I wonder how .json with numpy byte strings compares speed-wise...

rebeccafair:
> I wonder how .json with numpy byte strings compares speed-wise...
I'd imagine its quite a bit slower, but if you have so much data in the .json that reading it is becoming slow, then probably rendering the plot would be even slower, so hopefully we don't have to worry about that (at least not yet)...

Using metadata to store plot options is an interesting idea, I think originally we said metadata could contain absolutely anything, and we wouldn't use it in any computations, but I suppose we could force it to be a dictionary and have a few reserved keys for plot options. In addition to being more efficient for 2D plots it probably nicely fits in with the PDOS metadata stuff #123.

ajjackson:
We could reserve a single key `plot_options`, which is a dict of entries passed to `plot_Xd` as kwargs :smile:

PR #133, opened 2021-03-17
Add metadata to spectra

rebeccafair:
Addresses #123


rebeccafair:
@ajjackson I've only implemented some kind of metadata on `Spectrum1DCollection` at this point, but it would be good to get some feedback. Some of the main things:

- Is Dict[str, Any] the correct metadata type?
- How does the example for `methane_pdos` look?  Would we want to use a different key/format?
- Currently when we read in most data from .json files, if it is a list it gets converted to a Numpy array. I assume we don't want to make any assumptions, and don't want to do any conversion of anything in `metadata`? I haven't implemented it yet, but do we also want to do the same the other way, i.e. no conversion of Numpy arrays in metadata to lists before writing to .json? And we just make a note that anything in metadata must be JSON serialisable if the user wants to write to a .json?

ajjackson:
```python
{
    "__euphonic_class__": "Spectrum1DCollection",
    "__euphonic_version__": "0.4.0+2.g6696045.dirty",
    "metadata": {
        "PDOS": [
            "C1",
            "H2",
            "H3",
            "H4",
            "H5"
        ]
    }
```

It looks to me that this is collecting two pieces of information and needlessly linking them:

- Identifying the origin of the data as a PDOS. (Might be used for enabling some plotting settings?)
- Assigning labels to the lines

It's reasonable to label lines in data that is not a PDOS; e.g. quantum orders contributions to a simulated S(q,w), or an index relative to Fermi energy in an electronic band structure. It's also reasonable that this file might be read by some code that doesn't care about the specific nature of the data but _does_ want to use line labels.

I do think it's a good idea to use a dict as the outer container, even if the _main_ intended content is a list, because this gives us unlimited space for additional metadata (e.g. plotting parameters!)

How about something like:

```python
{
    "__euphonic_class__": "Spectrum1DCollection",
    "__euphonic_version__": "0.4.0+2.g6696045.dirty",
    "metadata": {
        "labels": [
            "C1",
            "H2",
            "H3",
            "H4",
            "H5"
        ],
        "plot_kwargs": {"y_label": "DOS", "title": "Methane (LDA)"},
        "broadening": {"shape": "Gauss", "sigma": 0.1}}
    }
```
where:
- *labels* is a reserved/standard category which Euphonic features will actively pick up and use (e.g. the `__repr__()` method of the spectrum class might show the included labels.)
- *plot_kwargs* is generally ignored but used by `plot_1d` and `plot_2d`
- *broadening* is unused but harmless metadata that gives information about the data and might be used in future (e.g. accounting for existing broadening when applying extra).

I'm not saying the *broadening* section should be added to PDOS output right now, just giving it as an example of the kind of extension we should leave open. We don't want to go crazy and clutter the namespace with useless stuff... yet :-)

rebeccafair:
This sounds sensible to me, do we still want a reserved `labels` key on `Spectrum1D` if there's only one line?

I like the `plot_kwargs` and `broadening` ideas, although I think I'll leave them for another time, possibly https://github.com/pace-neutrons/Euphonic/issues/130

ajjackson:
Reserved doesn't mean mandatory, I don't see the harm in _allowing_ labels on Spectrum1D but expect it would usually be omitted. And I guess it should be a singular `label: "TDOS"`?

rebeccafair:
> Reserved doesn't mean mandatory, I don't see the harm in _allowing_ labels on Spectrum1D but expect it would usually be omitted. And I guess it should be a singular `label: "TDOS"`?

Ok, so `label` in `Spectrum1D` gets converted to `labels` when creating a `Spectrum1DCollection` and vice-versa? Then we'd have 2 reserved keys, but I guess that does no harm...

ajjackson:
That's one reserved key for Spectrum1D and one reserved key for Spectrum1DCollection, there's no redundancy.

rebeccafair:
This actually ended up as more work than anticipated due to dealing with labels logic when splitting/combining spectra. How does it look?

One particular thing to mention is the handling of metadata when doing `Spectrum1DCollection.from_spectra`. Currently I've just copied the metadata from the first spectrum (much like the x_tick_labels), and ignored the metadata of the other spectra except for the `'label'` keys, which have been combined to create the `'labels'` key on the new `Spectrum1DCollection`.

euphonic/spectra.py
```diff
@@ -229,62 +242,69 @@ class Spectrum1D(Spectrum):
...
+            [(-1,), (), ()],
+            ['y_data', 'x_tick_labels', 'metadata'])
         ny = len(y_data)
         _check_constructor_inputs(
             [x_data], [Quantity],
             [[(ny,), (ny+1,)]], ['x_data'])
         self._set_data(x_data, 'x')
         self._set_data(y_data, 'y')
         self.x_tick_labels = x_tick_labels
+        self.metadata = metadata
```
ajjackson, line 293: As suggested in plotting:
```python
self.metadata = {} if metadata is None else metadata
```
- guarantee that metadata is always a dict and can be queried with `in` or `get`; saves on lots of type-checking elsewhere
- we still use `None` as the default empty argument to avoid Python pass-by-reference nasties

Maybe we should think think about using `dict.copy()` to avoid mutation as well. Which behaviour is more/less surprising?

euphonic/spectra.py
```diff
@@ -473,9 +527,24 @@ def _type_check(spectrum):
...
 
-        y_data = Quantity(y_data_magnitude, y_data_units)
+        # Process metadata
+        metadata = deepcopy(spectra[0]).metadata
+        try:
+            metadata.pop('label')
+        except (AttributeError, KeyError):
+            pass
+        for i, spectrum in enumerate(spectra):
+            if isinstance(spectrum.metadata, dict):
```
ajjackson, line 537: can skip this level if we make `spectrum.metadata` always a dict

euphonic/spectra.py
```diff
@@ -473,9 +527,24 @@ def _type_check(spectrum):
...
+                if 'label' in spectrum.metadata.keys():
+                    if not isinstance(metadata, dict):
```
ajjackson, line 539: ditto

euphonic/spectra.py
```diff
@@ -473,9 +527,24 @@ def _type_check(spectrum):
...
+                        metadata = {}
+                    if not 'labels' in metadata.keys():
+                        metadata['labels'] = ['']*len(spectra)
+                    metadata['labels'][i] = spectrum.metadata['label']
```
ajjackson, line 543: This took me a few tries to understand. Can we add a comment above L541 along the lines of "This is the first line with a label, initialise labels list"? At first glance it looks like L542 can clobber a list that was previously modified on L543 which would be bad.

euphonic/spectra.py
```diff
@@ -473,9 +527,24 @@ def _type_check(spectrum):
             assert spectrum.x_tick_labels == x_tick_labels
             y_data_magnitude[i + 1, :] = spectrum.y_data.magnitude
 
-        y_data = Quantity(y_data_magnitude, y_data_units)
+        # Process metadata
+        metadata = deepcopy(spectra[0]).metadata
```
ajjackson, line 531: What happens if spectra[1] has totally different metadata?

I think ideally, with the exception of `labels`, we would retain metadata that is the same for all component spectra and either discard or reorganise data that differs by line.

Maybe `labels` shouldn't be a special case, and there should be a general pattern for Spectrum1dCollection metadata such that

```python
spec1.metadata = {'element': 'K', 'label': 'Potassium 1', 'XC': 'PBE', 'calctime': 107.43}
spec2.metadata = {'element': 'K', 'label': 'Potassium 2', 'XC': 'PBE', 'calctime': 105.23}

Spectrum1DCollection.from_spectra([spec1, spec2]).metadata = {
    'element': 'K',
    'XC': 'PBE',
    'line_data': [{'label': 'Potassium 1', 'calctime': 107.43},
                  {'label': 'Potassium 2', 'calctime': 105.23}]}
```

Pros:
- completely reversible to recover consistent Spectrum1D.metada when iterating
- `'label'` and `'labels'` are not inconsistent between metadata dictionaries, and are only used in function arguments as appropriate.
- extends naturally without adding more special cases, would play very nicely with a `plot_args` option for storing desired linewidths, colours etc.

Cons:
- I appreciate this may not be a welcome suggestion after all the work you did to implement my previous suggestion :disappointed:

euphonic/spectra.py
```diff
@@ -565,53 +652,48 @@ def __init__(self, x_data, y_data, z_data, x_tick_labels=None):
...
-    @property
-    def x_data(self):
-        return self._x_data*ureg(self._internal_x_data_unit).to(
-            self.x_data_unit)
-
-    @property
-    def y_data(self):
-        return self._y_data*ureg(self._internal_y_data_unit).to(
-            self.y_data_unit)
+        self.metadata = metadata
```
ajjackson, line 655:
```python
self.metadata = {} if metadata is None else metadata
```

ajjackson (changes requested):
This was a lot of work! A mixture of
- more type-hinting
- supporting metadata dicts in general
- dealing with the `label`/`labels` special case.

See my code comments; the two main things are
- initialise metadata to an empty dict by default, there's no significance to `None` and it makes everything else easier
- maybe there's a better strategy for things like `label`/`labels`. Sorry I didn't realise it earlier, it's a lot easier when there's code to look at!

euphonic/spectra.py
rebeccafair: That would make the overall logic simpler, the only slight issue is that `"metadata": {}` would have to be added to every specrum JSON output in the tests. Unless I add something to avoid writing an empty metadata attribute? There is already logic to omit `None` values when converting to dicts/json. Which would be preferable, adding metadata to every .json or adding logic to omit it if it's empty?

The copy behaviour is a little inconsistent currently - if a `Quantity` is passed it gets copied during unit conversion, but if an array is passed it is just set as the attribute and no copy is made. We could set the behaviour to always copy but for things like eigenvectors this could potentially be quite expensive... I guess for the case of metadata it makes sense to copy for now as it should be quite small, but we should probably decide on something consistent or at least document it somewhere...

rebeccafair: I'll add some logic to omit empty metadata from outputs for now so that the tests keep passing, and then update the test data later if needed.

euphonic/spectra.py
rebeccafair: Yep, I agree this would probably be a better solution, these things aren't always obvious on the first pass. I'll have a go at changing the implementation and then address your other comments based on the state of the code once I'm finished..

euphonic/spectra.py
ajjackson: Omitting empty metadata from JSON seems fine to me. Think of it as part of **kwargs for `__init__`, where missing keys will be replaced by the default (None) during instantiation.

euphonic/plot.py
```diff
@@ -119,7 +149,11 @@ def plot_1d(spectra: Union[Spectrum1D,
     y_max
         Maximum value on the y-axis.
     labels
-        Legend labels for spectra, in the same order as spectra
+        The label to be applied to each line in spectra, must be the same
```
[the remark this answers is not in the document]
rebeccafair: Whoops, think I meant to write 'The label to be applied to each line in spectra, must be the same _length_ as the number of lines'

euphonic/plot.py
```diff
@@ -33,37 +33,67 @@ def plot_1d_to_axis(spectra: Union[Spectrum1D, Spectrum1DCollection],
...
+                            pass
+                else:
+                    label = labels[i]
+
             p = ax.plot(spectrum.get_bin_centres().magnitude[x0:x1],
                         spectrum.y_data.magnitude[x0:x1],
-                        color=color, **mplargs)
+                        color=color, label=label, **mplargs)
+
+    # Update legend if it exists, in case new labels have been added
```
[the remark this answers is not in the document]
rebeccafair: It depends, as this is a user-facing function, do we want them to have to manually redraw the legend after they've added a line? I would've thought if they're adding a line with a label it would be expected that the legend will be updated to reflect that automatically.

euphonic/plot.py
```diff
@@ -20,7 +20,7 @@
 
 
 def plot_1d_to_axis(spectra: Union[Spectrum1D, Spectrum1DCollection],
-                    ax: Axes,
+                    ax: Axes, labels: Optional[List[str]] = None,
```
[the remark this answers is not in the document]
rebeccafair: I guess having `Sequence` could be confusing, because then what does a `str` mean? Is that a label for one line or N 1-character labels? Almost certainly the former, but still...

Do we want to support just having `str` for one line label? `Optional[Union[List[str], str]]`?

euphonic/spectra.py
ajjackson: When I implemented something like this for ASE, I found `set(some_dict.items())` to be a useful tool: a set of 2-tuples allows you to easily find the common and different key-value pairs.

euphonic/plot.py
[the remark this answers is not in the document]
ajjackson: It's a straight trade-off in usability vs complexity, I'd be happy either way. Users are going to insert a string occasionally.

We either keep it strict with `Optional[List[str]]` and follow that up with an explicit `isinstance(labels, list)` check to make sure users don't get creative, or we go for the permissive `Optional[Sequence[str]]` with a check to convert `str` to `[str]`. (`str` *is* a `Sequence[str]`, so no `Union` is necessary.)

euphonic/plot.py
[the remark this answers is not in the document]
ajjackson: The usual design pattern for matplotlib scripts is not to draw the legend until you're done plotting. I suppose this is nice for interactive use? I can't think of a case where it would be a _problem_, except that it might make debugging confusing if people don't know the feature is there.

euphonic/plot.py
```diff
@@ -78,11 +78,10 @@ def plot_1d_to_axis(spectra: Union[Spectrum1D, Spectrum1DCollection],
...
                 if labels is None:
-                    if spectrum.metadata is not None:
-                        try:
-                            label = spectrum.metadata['label']
-                        except KeyError:
-                            pass
+                    try:
+                        label = spectrum.metadata['label']
+                    except KeyError:
+                        pass
```
ajjackson, line 84: Can drop L78 and replace this try/except block with `label = spectrum.metadata.get('label')`.

rebeccafair:
Ok, I think that's the updates made!

ajjackson (approved):
This is looking cleaner and more extensible :-)

I made one note but happy for merge as-is.

euphonic/spectra.py
```diff
@@ -478,22 +487,19 @@ def __getitem__(self, item: slice) -> SC:
...
-        if 'labels' in new_metadata.keys():
-            labels = new_metadata.pop('labels')
-        else:
-            labels = None
+        line_metadata = new_metadata.pop('line_data',
+                                         [{} for _ in self._y_data])
         if isinstance(item, int):
-            if labels is not None:
-                new_metadata['label'] = labels[item]
+            new_metadata = {**new_metadata, **line_metadata[item]}
```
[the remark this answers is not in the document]
rebeccafair: I don't think this will break? Using {**x, **y}, any duplicate keys in y will overwrite the keys in x. Although maybe  `update` is clearer, I've changed this and added a new test for duplicate keys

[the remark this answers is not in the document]
ajjackson: You're quite right. I didn't realise this behaviour was defined, but it is! https://docs.python.org/3/reference/expressions.html#dict

I was thinking of function arguments, which do raise an error if there is redundancy. Still, `update()` does make this clearer.

PR #137, opened 2021-04-09
Adjust compiler settings, use CC if set and don't assume homebrew

ajjackson:
The conda-forge CI isn't working with the current setup as it doesn't have =brew= available on Mac and the =gcc= binary on linux has a platform-dependent name.

These tweaks allow the environment CC to be used, and avoid the build script blowing up if homebrew is unavailable.

If this PR is accepted it would be helpful to make a Euphonic point release as soon as possible for use in the conda-forge package.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -33,16 +33,27 @@ def get_c_extension():
         compile_args = ['/openmp']
         link_args = None
     elif platform == 'darwin':
-        # OSX - assume brew install llvm
-        brew_prefix_cmd_return = subprocess.run(["brew", "--prefix"],
-                                                stdout=subprocess.PIPE)
-        brew_prefix = brew_prefix_cmd_return.stdout.decode("utf-8").strip()
-        os.environ['CC'] = '{}/opt/llvm/bin/clang'.format(brew_prefix)
-        compile_args = ['-fopenmp']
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -33,16 +33,27 @@ def get_c_extension():
...
-        link_args = ['-L{}/opt/llvm/lib'.format(brew_prefix), '-fopenmp']
+        # OSX - if CC not set, assume brew install llvm
+        try:
+            brew_prefix_cmd_return = subprocess.run(["brew", "--prefix"],
+                                                    stdout=subprocess.PIPE)
+            brew_prefix = brew_prefix_cmd_return.stdout.decode("utf-8").strip()
+        except FileNotFoundError:
+            brew_prefix = None
+
+        if brew_prefix and not os.environ.get['CC']:
```
rebeccafair, line 44: Should be `os.environ.get('CC')`

Sorry for all the failed Mac build emails, I thought because of the force-push the Mac wasn't pulling the correct version of the branch but it turns out I was looking at the wrong part of `setup.py` 😅

rebeccafair (changes requested):

setup.py
ajjackson: Whoops!

PR #138, opened 2021-04-13
116 neutron weighted dos

rebeccafair:
I thought I'd have a go at this off the adaptive broadening branch as it has the `.phonon_dos` reading so I can check the results. I've only implemented PDOS so far, but I think it's reasonably correct, and enough to have a discussion. One of the main things that has become apparent is that using `density=True` of `np.histogram` is not the way to go when computing PDOS, as it normalises the area to 1, which  is definitely not what we want. I've instead tried implementing a by mode weighting, where each mode is scaled so that the area under the contribution of **that mode at that q-point** is equal to its mode weighting (this should account for different bin widths, and should give the same result even if some modes are outside the bin range). How does this implementation look?

This is currently causing some of the script tests to fail, but because the previous implementation is incorrect. Any script tests involving DOS and `--e-min` or `--e-max` are failing because they don't include the full range of the data, so again the new behaviour has changed the scaling in these cases.

Some examples of the PDOS below, with CASTEP PDOS too for comparison:
image
image
image

Its obvious that although the CASTEP/Euphonic PDOS look about the same, the scale is completely different. This would probably benefit from having some kind of proper scaling and units. I'll have to look at the CASTEP source and figure out what the current scaling is to see if I can transform it to be on the same scale as Euphonic.

What's the best way to go about reviewing/merging this? The features that #116 and #118 require overlap quite a lot and I've got quite a bit to add to both!

**Update:**

To-do list:

- [x] Change `calculate_pdos` to return per-atom PDOS **Done in #167**
- [x] Add a `Spectrum1DCollection.group_by` method to organise into species, probably as a PR to this branch, see issue #161. **Done in #167**
- [x] Alias `--weights` to `--weighting` and deprecate `--weights` in CL tools. **Done in #127**
- [x] Replace `weighting=total-dos` with something more descriptive - `coherent+incoherent`? **Done in #127**
- [x] Update S(Q,w) map units to be area/energy, for consistency with weighted DOS. **Moved to a separate issue #168**
- [x] Add `calculate_dos_map` function, taken from PR #155 
- [x] Add PDOS options to CL tools **Done in #127**
- [x] Update user documentation **Done in #177**
- [x] Update changelog **Done in #177**


ajjackson:
> What's the best way to go about reviewing/merging this? The features that #116 and #118 require overlap quite a lot and I've got quite a bit to add to both!

I suggest reducing the scope of #131 to close it sooner if the remaining requirements for #118 can stand alone. Maybe add a checklist to #118 to make this easier to track?

rebeccafair:
Thinking about [this comment](https://github.com/pace-neutrons/Euphonic/pull/131#issuecomment-814931677) more on DOS units, maybe we should just have the DOS units in Energy^-1, and make a comment in the docstring that it is 'per unit volume enclosed by the provided q-points e.g. if they cover the 1st BZ, the DOS is per reciprocal cell volume' or something similar, as we can't guarantee (and don't enforce) that they're a sensible sum over the 1st BZ? Neither CASTEP or Phonopy multiply by the reciprocal cell volume so maybe we should stick with what's usually done... what do ASE/other codes do?

ajjackson:
I think they manage that by always working in fractions of the recip cell volume (can think of this as being the units of q-point weights?) My pedantry about the "DOS" label aside, it seems this normalisation to the reciprocal cell is a widespread convention and if it makes the maths easier then I guess it's ok to use internally as well.

Under that scheme I don't see why we couldn't guarantee that the sum is sensible; we just need to make sure that the q-point weights add up to 1 (and normalise if they don't).

rebeccafair:
> I think they manage that by always working in fractions of the recip cell volume (can think of this as being the units of q-point weights?) My pedantry about the "DOS" label aside, it seems this normalisation to the reciprocal cell is a widespread convention and if it makes the maths easier then I guess it's ok to use internally as well.
> 
> Under that scheme I don't see why we couldn't guarantee that the sum is sensible; we just need to make sure that the q-point weights add up to 1 (and normalise if they don't).

Ah yeah, I hadn't thought of it in terms of q-point weights. Yeah I guess normalising to whatever the sum of the weights is is the thing to do, I've currently subsumed it into the `mode_weights` argument but I can change that

rebeccafair:
There have been some changes from last time, and this is probably ready for another review @ajjackson:
- DOS is normalised to the reciprocal cell by dividing by the sum of the q-point weights
- The area under the DOS curve will now sum to the number of modes (if the full range of data is included), to me this makes sense if our DOS units are states/energy/reciprocal cell? I've fudged the tests for now to check its working as expected, and will regenerate the test data if we agree this is a correct approach.
- DOS units are 1/energy - by default in the same units as `dos_bins`.
- DOS read from a CASTEP `.phonon_dos` is correctly converted from `cm` to `1/meV`, which then matches in scale to Euphonic-produced DOS in `1/meV`, see plot: 
image

We now have neutron-weighted DOS! It _seems_ to produce sensible results (see plot below). Only issue is the cross_section/mass factor gives different units to the regular DOS (area/(mass*energy)), does this even make sense?

image
image

So things that need to be addressed:
- Do we agree on DOS units of 1/energy? 
- Do we agree normalising so that the area under the curve = number of modes?
- What should the units of neutron-weighted DOS be? area/(mass*energy)? Is there some kind of unit reduction I'm not seeing?
- What can I check the neutron-weighted DOS outputs against? Maybe @davidvoneshen would know?

ajjackson:
If we consider "neutron-weighted DOS" to be a serious approximation of S(q,ω):

S(q,w) is defined with respect to "solid angle" (Ω, which is a function of angles θ & ϕ) so that should be in there somewhere (and quite likely is already included in the cross sections). The intensity will also depend on the sample depth, beam width, beam flux and detector sensitivity/calibration. Those can be lumped together into a parameter that should be consistent for a given instrument (and can be set to some kind of agreeable unity value for simulations?) I think the reciprocal unit cell should be accounted for (i.e. re-introduce the sample density) in order to give appropriate scaling for different materials on a given instrument.

Most likely there is some good information available wherever the cross-sections were initially defined and calculated.

davidvoneshen:
So, if we put our measurements into absolute units they are traditionally

mbarns /sr/meV/f.u.

this sorts out all of the terms you mention before (things can still go wrong but they are things the experimenter needs to worry about not euphonic). This is done by also measuring a standard vanadium with the same settings used to collect the data. It lets us normalise out source and instrument terms and if we are going to produce a direct comparison with no scaling parameter needed that would be the way to go. That would be for the line/colour plots to compare with Horace or abins. The formula units part is user defined (it could just as easily be per unit cell or other arbitrary choice). We have the knowledge needed for this, it is just a little tedious to work through.

For just the DOS we could multiply the euphonic one through by the say unit cell mass to give something in units of mbarns/meV/cell. Unit cell mass seems to me to be the most sensible choice as it doesn’t require another user specified parameter (formula unit might) but I guess the other option would be average atomic mass. Then it is up to me to make sure that our routines to produce the experimental DOS also do this properly.

David

rebeccafair:
Thanks both for your helpful comments!

> For just the DOS we could multiply the euphonic one through by the say unit cell mass to give something in units of mbarns/meV/cell. Unit cell mass seems to me to be the most sensible choice as it doesn’t require another user specified parameter (formula unit might) but I guess the other option would be average atomic mass. Then it is up to me to make sure that our routines to produce the experimental DOS also do this properly.

This would make sense to me, currently structure factor in euphonic is in units of length^2 per unit cell (although the 'unit cell' part is just mentioned in the docstring and is not formally included in the `pint` units, do we want/need to create our own unit for this? I feel like it could be overly cumbersome) so they would then match up. One slight oversight is that the intensities calculated by `StructureFactor.calculate_sqw_map` are still in length^2, rather than length^2/energy, the same for `StructureFactor.calculate_1d_average`, so it would make sense to change the units on these too.

So are we to leave out any solid angle related units and assume it is included in the cross section in some way? (The original data source is here but I don't seem to have access at the moment https://doi.org/10.1080/10448639208218770). If people are allowed to use their own cross sections, I don't think we could formally include a solid angle unit correctly anyway...

ajjackson:
Unfortunately reduced formula units are not used in molecular chemistry (e.g. the unit of ethane is C2H6, not CH3) so I don't think we can apply FU reduction consistently without additional information?

rebeccafair:
Ok, so I think I've got to a reasonable place where it makes sense to do a first pass review what's been done so far (and make sure I've actually got the neutron-weighted PDOS right before going any further). @davidvoneshen could you have a look at these plots and my logic to make sure I haven't done something silly? So currently I've got a separate `calculate_pdos` function that returns a `Spectrum1DCollection`, which contains the per-element PDOS and some kind of 'aggregated DOS'. If `weighting=None` this is just the sum of all the PDOS, so is the **total DOS**. If `weighting=coherent/incoherent`, this is the **average atom DOS**. I've given these different labels in the metadata, but I'm hoping this isn't too confusing or inconsistent. 

This is an example of what the plots might look like.
For `weighting=None`:
image
For `weighting='coherent'`:
image
Does this seem reasonable and would people generally understand the difference?

Also, the way I've calculated the 'average atom DOS' is just by taking x% of PDOS 1 + y% of PDOS 2 depending on the fraction of each species in the unit cell, does that make sense? There could be one problem with this though, I've set up the calculation so that the area under the 'total DOS' will be 3N. However, this means that when calculating the 'average atom DOS' by taking fractions of PDOS, the total area will be less than 3N. But we're now in units of `barn/meV/atom` rather than `1/meV` so I don't know if that even is a problem... 

There are still quite a few things to do, they are broadly:
- [ ] Convert `sqw_map` units to `[length]^2/[energy]` for consistency with neutron-weighted dos (current units are `[length]^2`)
- [ ] Add coherent/incoherent and pdos options to command line tools
- [ ] Update tests
- [ ] Update docs

I might end up doing some parts as PRs to this branch, as the CL bits for example have ended up being more complex than I anticipated, and it might help us finalise the API before putting it in master.

ajjackson:
It's a bit confusing to have the "total" average-atom-S(q,w) lower in magnitude than it's constituent parts. I'd scale those element contributions by the same amount so that it resembles a PDOS. They then become the contribution of that element to S(q,w) per atom of sample.

rebeccafair:
Thanks, I think I hadn't appreciated that all spectra should be in units of 'average sample atom' rather than per 'O' atom etc. I've corrected it now (I think), it's a lot more consistent and if you divide by average atom mass you recover the values per amu from the earlier plots, which gives me confidence that its correct.

euphonic/qpoint_phonon_modes.py
```diff
@@ -390,6 +395,123 @@ def calculate_debye_waller(
...
+                species_mass = self.crystal._atom_mass[spec_idx[0]]
+                dos *= cs[i]*avg_atom_mass/species_mass
+            if i == 0:
+                all_dos_y_data = np.zeros((
+                    len(spec_idx_dict) + 1, len(dos_bins) - 1))*dos.units
+            all_dos_y_data[i + 1] = dos
+
+        all_dos_y_data[0] = np.sum(all_dos_y_data[1:], axis=0)
+        metadata = {'line_data':
+            [{'label': x} for x in ['Total'] + list(spec_idx_dict.keys())]}
```
ajjackson, line 510: This seems to rely on `dict.keys()` having a consistent order with `spec_idx_dict.items()`. Having the base dict() object be consistently ordered is

- an implementation detail of CPython 3.6
- part of the language specification from Python 3.7

So in principle it could break if somebody uses e.g. pypy 3.6 or IronPython 3.6. (IronPython only just caught up with version 3.4, so this is not totally farfetched.)

A cheap and robust fix would be to modify `Crystal.get_species_idx` to use `collections.OrderedDict`. This is a no-op for CPython and makes the intent to _use_ the ordering explicit.

ajjackson (changes requested):
A few comments so far! I still need to look at the tests and try running some examples vs Phonopy and Mantid. But generally this looks good and, despite the large number of lines changed, not too complicated.

euphonic/qpoint_phonon_modes.py
```diff
@@ -390,6 +395,123 @@ def calculate_debye_waller(
...
+
+        spec_idx_dict = self.crystal.get_species_idx()
+        if isinstance(cross_sections, str) and weighting is not None:
+            cross_sections_data = get_reference_data(
+                collection=cross_sections,
+                physical_property=f'{weighting}_cross_section')
+        elif isinstance(cross_sections, dict):
+            cross_sections_data = cross_sections
+        else:
+            cross_sections_data = None
```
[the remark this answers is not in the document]
rebeccafair: Good spot, fixed!

euphonic/qpoint_phonon_modes.py
rebeccafair: This did occur to me at one point but then I forgot... thanks, fixed!

Issue #136, opened 2021-04-08
LICENSE is not in tarball


ajjackson:
The LICENSE file in the project root is not packaged into tarball distributions. I think this just needs a tweak to setup.py

ajjackson:
This was discovered while making a package for conda-forge. At the moment this has to ship a copy the missing file!

PR #144, opened 2021-04-15
Changelog update for 0.5.0 release

ajjackson:
I would like to make a release as soon as possible so that the setup.py tweaks are available from PyPI for conda builds.

We have added some significant features comparable to other 0.x releases, so I suggest incrementing to 0.5.0.

I have tried building this 0.5.0 release locally with release.py, was able to install and test the resulting .tar.gz package without issue. (I set the v0.5.0 tag locally, have not pushed the tag in case further changes are needed.)


rebeccafair:
Did you want https://github.com/pace-neutrons/Euphonic/issues/136 to be fixed first? Save including the license in the Conda package

ajjackson:
I guess it would make sense to do on this branch actually, it shouldn't be a big job.

ajjackson:
Ok, the new LICENSE setup seems to be playing well with conda-forge

PR #146, opened 2021-04-28
142 symmetrise debye waller

### LINKED ISSUE #142: Symmetrise Debye-Waller
It has been found that calculating the Debye-Waller exponent on symmetry-reduced q-point grids gives drastically different values to calculating it on full, unfolded q-point grids (see below plots). 

image
image

This has been found to be due to the fact that the Debye-Waller factor has not been symmetrised (compared to other codes where it has been, see the Atomic Displacement Parameter calculation in CASTEP). Symmetrisation should also be implemented in Euphonic to avoid this issue.

rebeccafair:
Closes #142


rebeccafair:
With symmetrisation, the Debye-Waller factors for both full and symmetry-reduced (weighted) grids are the same
quartz_dw_symm
nb_dw_symm

Issue #150, opened 2021-04-30
Cache C extension library location


rebeccafair:
Currently the library file containing `zheevd_` has to be found and loaded on every call to `calculate_qpoint_phonon_modes`. We should cache this as it shouldn't change between calls.

PR #151, opened 2021-05-01
Change load_libs.c to use scipy cython refs

### LINKED ISSUE #149: Improve process of finding the DLL in C Extension
#147 has shown that, depending on the install method, the DLL containing the required `zheevd_` function could have many possible names and locations. Is it probably unsustainable to keep adding special cases, so a better solution should be found.

Possibilities:
- Use https://github.com/erocarrera/pefile (thanks @mducle)
```
import pefile
filepath = r'C:\tools\miniconda3\envs\py37e\Lib\site-packages\scipy\linalg\_flapack.cp37-win_amd64.pyd'
pe = pefile.PE(filepath)
for entry in pe.DIRECTORY_ENTRY_IMPORT:
    funcs = [f.name.decode().lower() for f in entry.imports]
    if 'zheevd_' in funcs:
        print(entry.dll.decode())
```

- Enable specifying the DLL with an environment variable

mducle:
This PR presents an alternative way to get a pointer to the LAPACK `zheevd` matrix eigensolver function. Instead of loading this from a shared library, it uses the [function pointers](https://docs.scipy.org/doc/scipy/reference/linalg.cython_lapack.html#module-scipy.linalg.cython_lapack) exposed by `scipy` for use by `cython`. These functions have the same Fortran calling convention as `zheevd_` which is loaded from the libraries but should point to whichever function `scipy`/`numpy` was compiled with on the user's system.

The PR:

* Removed searching / loading of shared libraries.
* Instead use the function pointers provided by scipy for use with Cython
* Also implement a simple caching mechanism using a global variable.

Fixes #149, #150

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -18,9 +18,10 @@
     frequencies. This improves on the previous behaviour which only excluded
     gamma-point acoustic modes, so would miss small/negative frequencies
     elsewhere
-  - When loading libraries for the C extension in Windows, there are now
-    additional search paths, and ``liblapack.dll`` or ``mkl_rt.1.dll`` will
-    be attempted to be loaded if ``libopenblas.dll`` can't be found
+  - Loading the LAPACK libraries for the C extension now uses the
+    `interface <https://docs.scipy.org/doc/scipy/reference/linalg.cython_lapack.html`_
+    provided by `scipy` for `cython` instead of loading directly from a DLL.
+    The new method means we don't have to guess the DLL filename anymore!
 
 - Changes:
 
--- a/c/_euphonic.c
+++ b/c/_euphonic.c
@@ -32,7 +32,6 @@ static PyObject *calculate_phonons(PyObject *self, PyObject *args) {
     int reciprocal_asr;
     int splitting;
     int n_threads = 1;
-    const char *scipy_dir;
 
     // Define vars to be obtained from ForceConstants attributes
     PyObject *py_crystal; // Crystal object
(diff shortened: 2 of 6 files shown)


rebeccafair (changes requested):
Thanks a lot for this Duc, it's a lot simpler and more robust now! I've managed to build it locally, and have got the CI working for [Mac ](https://anvil.softeng-support.ac.uk/jenkins/job/PACE-neutrons/job/Euphonic/job/Euphonic%20Mac%20Pipeline/job/origin%252Fpull%252F151%252Fmerge/3/console). It also works on SCARF with 24 threads and has only a very small effect on performance (felt like I had to test this due to the GIL stuff). Unfortunately the Linux build is failing, but I think that might be unrelated...

Due to these changes, the `scipy_dir` argument is no longer needed, can you remove all references to it in the C and Python? This should be straighforward, just note that one line: https://github.com/pace-neutrons/Euphonic/blob/master/c/_euphonic.c#L97
`if (!PyArg_ParseTuple(args, "OO!O!O!O!O!O!O!O!O!iiiO!O!O!O!is",`
will need to be changed to
`if (!PyArg_ParseTuple(args, "OO!O!O!O!O!O!O!O!O!iiiO!O!O!O!i",`
as the `euphonic._euphonic.calculate_phonons` signature has changed to remove the `scipy_dir`

And a few smaller changes:
- Rebase onto master with the new CI changes (maybe this will fix the Linux build)
- Update the changelog

rebeccafair (changes requested):
One more thing, we no longer need the `import scipy` or `from scipy.linalg.lapack import zheev` at the top of `force_constants.py` (second one is unrelated to this PR but might as well clean up while we're at it). Once that's done I think we're good!

PR #154, opened 2021-05-05
Correct SHA for PR in Jenkinsfile

mducle:
Changes the `Jenkinsfile` to get the correct SHA1 for the commit which triggers a PR build (rather than the merge commit which github passes to Jenkins).


rebeccafair (approved):
Looks good to me, and seems to work! Thanks for this

Issue #139, opened 2021-04-13
Add adaptive broadening to powder map


rebeccafair:
Adaptive broadening is available as an option in `euphonic-dos`, it should also be made available in `powder-map`

ajjackson:
I'm wondering how this would work. Would we use the radial component of the derivatives to broaden in the q direction, and the perpendicular components to broaden in energy?

PR #155, opened 2021-05-13
116 Neutron weighting CL tool options

rebeccafair:
So here's a first pass attempt at adding PDOS/weighting options to the CL tools, sorry in advance...

Apart from just adding the CL arguments there have been a few other additions:
* Found a bug in `powder-map` where bins weren't generated correctly if different units were used, as `euphonic.cli.utils._get_energy_bins_and_units` didn't check units so silently gave incorrect results if `modes.frequencies_unit = energy_unit` was not called first. Have refactored this slightly and regenerated the test data
* Created a new `calculate_dos_map` method on `QpointPhononFrequencies`, and used this in `euphonic-intensity-map` rather than the original ad-hoc method. This will make it more consistent with `calculate_dos`, and will help when adding more DOS options, this will feed into the API discussion further down...
* When playing with the DOS map, I found that `get_energy_bins_and_units` returns `--ebins` values, but they are then used in the CL tools as bin edges, so we end up with `ebins-1` energy bins, I've fixed this in `intensity-map` and have regenerated the data, but haven't fixed it in `powder-map` yet

Summary of CL additions:
* New options `coherent-dos`, `incoherent-dos` and `total-dos` for `--weights`. 
* `euphonic-dos` now also has a `--weights` argument but it only contains the DOS options (I didn't add the coherent INS option as I'm not sure it's useful/relevant if we're only summing in the 1st BZ, and its not a DOS). I imagine getting a 1D coherent INS spectrum would be useful, but maybe it belongs in another tool which sums at specific Q...
* New `--pdos` option for `euphonic-dos`, which accepts any number of string arguments, if not provided total DOS will be shown, if provided with no string arguments, all per-species PDOS and total DOS will be shown, if provided with species e.g. `--pdos La Zr` only those species will be shown.
* New `--pdos` option for `intensity-map`/`powder-map`, which behaves as above, except it accepts only **one** string argument. Ideally I would like it to behave the same as the `--pdos` for `euphonic-dos`, but in that case we would have to show multiple plots. We could do it, but would require some more work. Would we want multiple windows, or multiple subplots? How could we test it?

API issues raised:
* Having to translate between the CL `--pdos x y z` and the API, calling the correct version `calculate_pdos/calculate_dos` and then find the correct returned spectrum index is very cumbersome. Maybe we should just have a single `calculate_dos` method with a `pdos` argument which behaves similarly to the CL options instead? This was the other option for the API, and now I'm thinking it might be the way to go instead
* The above point also applies to `calculate_dos_map` and `calculate_pdos_map` methods
* It _also_ applies to `sample_sphere_dos`, see the horrible solution I've currently implemented. Being able to just pass through a `pdos` argument would be much simpler.
* We have `Spectrum1DCollection` to represent multiple PDOS, what do we want to do about representing multiple 2D PDOS maps? So far I've just used a list of `Spectrum2D`
* Inconsistency between `--weights` argument in the CL tools and `weighting` argument in `calculate_pdos`, maybe we should rename the `calculate_pdos` argument?
* `--weights total-dos` is quite confusing, it really does sound like total DOS rather than the sum of coherent/incoherent cross sections. We should have the `--weights` and `calculate_pdos` argument be consistent, but I'm not sure what it could be renamed to? `coh-incoh`? `sum`?
* `Spectrum1DCollection` should have a broaden method, calling `Spectrum1DCollection.from_spectra([spec.broaden(bwidth) for spec in spectra])` every time is not nice...

Thoughts on adaptive broadening:
* I was originally going to add adaptive broadening to `powder-map` as part of this PR (#139 ), but its ended up being a lot more complicated than I realised. The `mode_gradients_to_widths` function relies on the length of `qpts` to generate the widths, this means the broadening will be different depending on the number of q-points in that shell, which seems wrong. This probably needs some more thought...
* I don't know if adaptive broadening (at least not with mode widths generated by the dynamical matrix gradients as we currently have) is actually applicable along a q-path, I think it only applies to BZ sums. I think the more correct way for q-paths would be to actually generate them from the dispersion gradients, which we have for a q-path. Again this would require more work.

To do (based on discussion below):

- [x] Update `powder-map` to have correct number of ebins. **Done: de68ac5**
- [x] Add nice-to-have label to Euphonic Issues **Done: https://github.com/pace-neutrons/Euphonic/labels/nice%20to%20have**
- [x] Add issue for handling multiple Spectrum2D objects **Done: https://github.com/pace-neutrons/Euphonic/issues/158**
- [x] Ensure CL tool argument `--pdos` in `euphonic-dos` and `intensity-map` take the same arguments, returning a plot with multiple subplots for Spectrum2D **Done: Decided to de-scope and add this at a later date, issue created #164**
- [x] Change `calculate_pdos` to return per-atom PDOS, with a `Spectrum1DCollection.group_by` method to organise into species **Done: Moved to PR #138 and issue #161 created**
- [x] Alias `--weights` to `--weighting` and deprecate `--weights` in CL tools **Done: Moved to PR #138**
- [x] Replace `weighting=total-dos` with something more descriptive - `coherent+incoherent`? **Done: Moved to PR #138**
- [x] New issue for `Spectrum1DCollection.broaden` **Done: #157**
- [x] Determine method for adaptive broadening on 2D Q-E maps, create relevant issues **Done: #159 #160 created**


ajjackson:
Some replies to these notes before I have a proper comb through the code!

> * Created a new `calculate_dos_map` method on `QpointPhononFrequencies`, and used this in `euphonic-intensity-map` rather than the original ad-hoc method. This will make it more consistent with `calculate_dos`, and will help when adding more DOS options, this will feed into the API discussion further down...

Sounds good!

> * When playing with the DOS map, I found that `get_energy_bins_and_units` returns `--ebins` values, but they are then used in the CL tools as bin edges, so we end up with `ebins-1` energy bins, I've fixed this in `intensity-map` and have regenerated the data, but haven't fixed it in `powder-map` yet

Cool

> * `euphonic-dos` now also has a `--weights` argument but it only contains the DOS options (I didn't add the coherent INS option as I'm not sure it's useful/relevant if we're only summing in the 1st BZ, and its not a DOS). I imagine getting a 1D coherent INS spectrum would be useful, but maybe it belongs in another tool which sums at specific Q...

I think having weighted DOS vs other methods in different tools would be a good way to avoid confusion there.

> * New `--pdos` option for `euphonic-dos`, which accepts any number of string arguments, if not provided total DOS will be shown, if provided with no string arguments, all per-species PDOS and total DOS will be shown, if provided with species e.g. `--pdos La Zr` only those species will be shown.
> 
> * New `--pdos` option for `intensity-map`/`powder-map`, which behaves as above, except it accepts only **one** string argument. Ideally I would like it to behave the same as the `--pdos` for `euphonic-dos`, but in that case we would have to show multiple plots. We could do it, but would require some more work. Would we want multiple windows, or multiple subplots? How could we test it?

The limit I could see for _overlaying_ spectral plots would be two elements in contrasting colours over a neutral background. Even this would likely be an accessibility disaster, so realistically we are looking at vertical/horizontal stacks of axes. As we've sort-of-established that horizontal splits are used for x-axis discontinuities, a vertical stack seems best. If we were only calling `plot_2d_to_axis()` this would be pretty straightforward to implement. It's not really compatible with `plot_2d()` and x-axis breaks though. Maybe a `plot_2d_stack()` would form a useful intermediate step? I don't think it's a high priority to provide this now, are we tracking "nice-to-have"s somewhere?

> * Having to translate between the CL `--pdos x y z` and the API, calling the correct version `calculate_pdos/calculate_dos` and then find the correct returned spectrum index is very cumbersome. Maybe we should just have a single `calculate_dos` method with a `pdos` argument which behaves similarly to the CL options instead? This was the other option for the API, and now I'm thinking it might be the way to go instead

How about if there's a nice utility function/method for converting a Spectrum1DCollection-by-atom to a collection by *\<preferred grouping scheme\>*? I think it's generally a good idea to leave this kind of flattening/discarding step as late as possible, and we don't want to make calculate_pdos more complicated.

e.g.
```python
pdos = modes.calculate_pdos(*args, **kwargs)
pdos_for_plotting = pdos.group_by_element(elements=args.pdos)
plot_1d(pdos_for_plotting, **plot_kwargs)
```
This leaves the door open for other grouping schemes too :grin: 

> 
> * The above point also applies to `calculate_dos_map` and `calculate_pdos_map` methods
> 
> * It _also_ applies to `sample_sphere_dos`, see the horrible solution I've currently implemented. Being able to just pass through a `pdos` argument would be much simpler.

Of course

> * We have `Spectrum1DCollection` to represent multiple PDOS, what do we want to do about representing multiple 2D PDOS maps? So far I've just used a list of `Spectrum2D`

Without a standard way of plotting them, it's not clear how useful such an object would be. Hold out until we need it?

> * Inconsistency between `--weights` argument in the CL tools and `weighting` argument in `calculate_pdos`, maybe we should rename the `calculate_pdos` argument?

I think there are arguments in favour of both names. We have the `-w` abbreviation available at command-line anyway so verbosity is not a big concern. Let's use whichever you prefer - but to avoid API breakage we should introduce an alias and raise DeprecationWarning before removing in next major release. Overkill perhaps, but it's a nice case to practice doing this!

> * `--weights total-dos` is quite confusing, it really does sound like total DOS rather than the sum of coherent/incoherent cross sections. We should have the `--weights` and `calculate_pdos` argument be consistent, but I'm not sure what it could be renamed to? `coh-incoh`? `sum`?

I agree this is horribly confusing. How about `incoherent+coherent` or `coherent+incoherent`? It's a bit clunky, but at least fairly explicit about what is going on. It's also rather clunky to include `dos` in the arguments for already-explicitly-a-DOS tool(s)... but good to be consistent with `powder-map`/`intensity-map` which can do both methods.

> * `Spectrum1DCollection` should have a broaden method, calling `Spectrum1DCollection.from_spectra([spec.broaden(bwidth) for spec in spectra])` every time is not nice...

Agreed! Should that be in the PR or shall we make a new Issue for it?

> 
> Thoughts on adaptive broadening:
> 
> * I was originally going to add adaptive broadening to `powder-map` as part of this PR (#139 ), but its ended up being a lot more complicated than I realised. The `mode_gradients_to_widths` function relies on the length of `qpts` to generate the widths, this means the broadening will be different depending on the number of q-points in that shell, which seems wrong. This probably needs some more thought...

Sounds like it can wait then.

> * I don't know if adaptive broadening (at least not with mode widths generated by the dynamical matrix gradients as we currently have) is actually applicable along a q-path, I think it only applies to BZ sums. I think the more correct way for q-paths would be to actually generate them from the dispersion gradients, which we have for a q-path. Again this would require more work.

Intuitively I'd expect that adaptive broadening for calculation along a path could
- use the gradients in the x-direction to set x-axis broadening. (This would be quite painful as we have no path to applying per-mode x-axis broadening.)
- use the gradients normal to that direction to set the y-axis broadening. (This is more feasible if we can somehow introduce the directional information when calculating widths; something like |**g**| <sub>yz</sub> = |**g**| - **g.x**/|**x**|) ?)

rebeccafair:
> > * `euphonic-dos` now also has a `--weights` argument but it only contains the DOS options (I didn't add the coherent INS option as I'm not sure it's useful/relevant if we're only summing in the 1st BZ, and its not a DOS). I imagine getting a 1D coherent INS spectrum would be useful, but maybe it belongs in another tool which sums at specific Q...
> 
> I think having weighted DOS vs other methods in different tools would be a good way to avoid confusion there.

Ok good, gives me less to do in this PR

> > * New `--pdos` option for `euphonic-dos`, which accepts any number of string arguments, if not provided total DOS will be shown, if provided with no string arguments, all per-species PDOS and total DOS will be shown, if provided with species e.g. `--pdos La Zr` only those species will be shown.
> > * New `--pdos` option for `intensity-map`/`powder-map`, which behaves as above, except it accepts only **one** string argument. Ideally I would like it to behave the same as the `--pdos` for `euphonic-dos`, but in that case we would have to show multiple plots. We could do it, but would require some more work. Would we want multiple windows, or multiple subplots? How could we test it?
> 
> The limit I could see for _overlaying_ spectral plots would be two elements in contrasting colours over a neutral background. Even this would likely be an accessibility disaster, so realistically we are looking at vertical/horizontal stacks of axes. As we've sort-of-established that horizontal splits are used for x-axis discontinuities, a vertical stack seems best. If we were only calling `plot_2d_to_axis()` this would be pretty straightforward to implement. It's not really compatible with `plot_2d()` and x-axis breaks though. Maybe a `plot_2d_stack()` would form a useful intermediate step? I don't think it's a high priority to provide this now, are we tracking "nice-to-have"s somewhere?

We're not currently tracking nice-to-haves, that would probably be a good label for some of our issues as currently they all appear to have the same priority! I'll add an issue for this once we decide specifically what the issue actually is..

So for now, what are we aiming for? I can allow `--pdos` in `euphonic-dos` and `euphonic-intensity-map` to behave the same, and just plot them on the same plot by using `plot_2d_to_axis`? If we keep it just in the command tools that's probably easiest until we have more use-cases. Hopefully I can find an easy way to test it...

And is it ok for `calculate_pdos_map` to just return a list of Spectrum2D? Although this may cause some issues if we're to be consistent with `calculate_pdos`, see below.

> > * Having to translate between the CL `--pdos x y z` and the API, calling the correct version `calculate_pdos/calculate_dos` and then find the correct returned spectrum index is very cumbersome. Maybe we should just have a single `calculate_dos` method with a `pdos` argument which behaves similarly to the CL options instead? This was the other option for the API, and now I'm thinking it might be the way to go instead
> 
> How about if there's a nice utility function/method for converting a Spectrum1DCollection-by-atom to a collection by _<preferred grouping scheme>_? I think it's generally a good idea to leave this kind of flattening/discarding step as late as possible, and we don't want to make calculate_pdos more complicated.
> 
> e.g.
> 
> ```python
> pdos = modes.calculate_pdos(*args, **kwargs)
> pdos_for_plotting = pdos.group_by_element(elements=args.pdos)
> plot_1d(pdos_for_plotting, **plot_kwargs)
> ```
> 
> This leaves the door open for other grouping schemes too 😁

Ok, so instead of returning per-species PDOS, `calculate_pdos` should keep it as decomposed as possible and return per-atom PDOS, which can be easily grouped into per-species PDOS. `group_by_element` seems slightly specific for a generic `Spectrum1DCollection` object, maybe we can have a generic `group_by` which accepts certain special strings (e.g. 'species'), but I imagine in the future it could be extended (e.g. allowing other special strings, or a regex containing a capturing group). Something like `group_by(grouping='species', matches=args.pdos)`.

The only issue with this then is the inconsistency between `calculate_pdos` and `calculate_pdos_map`. If `calculate_pdos` moves to returning it by atom, then so should `calculate_pdos_map`, but we have no way of easily grouping lists of `Spectrum2D`. For now I might just provide a function in `euphonic.cli.util` (or `euphonic.util`?) to do this. This should be ok, I don't know what the demand would be for DOS maps outside of the CL tools.

> > * Inconsistency between `--weights` argument in the CL tools and `weighting` argument in `calculate_pdos`, maybe we should rename the `calculate_pdos` argument?
> 
> I think there are arguments in favour of both names. We have the `-w` abbreviation available at command-line anyway so verbosity is not a big concern. Let's use whichever you prefer - but to avoid API breakage we should introduce an alias and raise DeprecationWarning before removing in next major release. Overkill perhaps, but it's a nice case to practice doing this!
>

Sounds good. I can't say why but I do prefer `weighting`, maybe `weights` makes it sound like it should be a list of values.

> > * `--weights total-dos` is quite confusing, it really does sound like total DOS rather than the sum of coherent/incoherent cross sections. We should have the `--weights` and `calculate_pdos` argument be consistent, but I'm not sure what it could be renamed to? `coh-incoh`? `sum`?
> 
> I agree this is horribly confusing. How about `incoherent+coherent` or `coherent+incoherent`? It's a bit clunky, but at least fairly explicit about what is going on. It's also rather clunky to include `dos` in the arguments for already-explicitly-a-DOS tool(s)... but good to be consistent with `powder-map`/`intensity-map` which can do both methods.

`coherent+incoherent` would work in `calculate_pdos`, but then we would have `-w coherent+incoherent-dos` which looks a bit weird.. but better than `total` in any case.

> > * `Spectrum1DCollection` should have a broaden method, calling `Spectrum1DCollection.from_spectra([spec.broaden(bwidth) for spec in spectra])` every time is not nice...
> 
> Agreed! Should that be in the PR or shall we make a new Issue for it?

New issue incoming!

> > * I don't know if adaptive broadening (at least not with mode widths generated by the dynamical matrix gradients as we currently have) is actually applicable along a q-path, I think it only applies to BZ sums. I think the more correct way for q-paths would be to actually generate them from the dispersion gradients, which we have for a q-path. Again this would require more work.
> 
> Intuitively I'd expect that adaptive broadening for calculation along a path could
> 
> * use the gradients in the x-direction to set x-axis broadening. (This would be quite painful as we have no path to applying per-mode x-axis broadening.)
> * use the gradients normal to that direction to set the y-axis broadening. (This is more feasible if we can somehow introduce the directional information when calculating widths; something like |**g**| yz = |**g**| - **g.x**/|**x**|) ?)

Ah I'd never thought of doing the x-axis broadening too! We do have the mode gradients in each Cartesian direction at each q-point as an intermediate calculation step (https://github.com/pace-neutrons/Euphonic/blob/master/euphonic/force_constants.py#L692-L694), so converting from this to gradients parallel and perpendicular to the q-direction should be possible. I'm now wondering whether `calculate_qpoint_phonon_modes/frequencies` should return these xyz gradients rather than the final mode widths, it would give us more freedom in the future (and would help slightly with our powder sampling mode widths problem above). If we do have to change it, its probably easier to do it sooner rather than later. It might be a bit of work though, maybe we should talk to some others (e.g. Keith) to make sure this method makes sense before making any changes to the code.

And as the code is at the moment yes having gradients in both directions would be quite painful... but with some internal refactoring it should be possible.

I've added a to-do list to the top of this PR just to gather things together (they're not all appropriate for this PR) and will edit as necessary

ajjackson:
> > I think there are arguments in favour of both names. We have the `-w` abbreviation available at command-line anyway so verbosity is not a big concern. Let's use whichever you prefer - but to avoid API breakage we should introduce an alias and raise DeprecationWarning before removing in next major release. Overkill perhaps, but it's a nice case to practice doing this!
> 
> Sounds good. I can't say why but I do prefer `weighting`, maybe `weights` makes it sound like it should be a list of values.

Me too!

> > > * `--weights total-dos` is quite confusing, it really does sound like total DOS rather than the sum of coherent/incoherent cross sections. We should have the `--weights` and `calculate_pdos` argument be consistent, but I'm not sure what it could be renamed to? `coh-incoh`? `sum`?
> > 
> > 
> > I agree this is horribly confusing. How about `incoherent+coherent` or `coherent+incoherent`? It's a bit clunky, but at least fairly explicit about what is going on. It's also rather clunky to include `dos` in the arguments for already-explicitly-a-DOS tool(s)... but good to be consistent with `powder-map`/`intensity-map` which can do both methods.
> 
> `coherent+incoherent` would work in `calculate_pdos`, but then we would have `-w coherent+incoherent-dos` which looks a bit weird.. but better than `total` in any case.

It's not great, is it? Naming things is hard, this may need a session with a whiteboard to map out the various programs and options.

> ... If we do have to change it, its probably easier to do it sooner rather than later. It might be a bit of work though, maybe we should talk to some others (e.g. Keith) to make sure this method makes sense before making any changes to the code.

We should talk to Keith about the theory. From an API perspective
>  I'm now wondering whether `calculate_qpoint_phonon_modes/frequencies` should return these xyz gradients rather than the final mode widths,

seems an excellent idea.

ajjackson:
> `group_by_element` seems slightly specific for a generic `Spectrum1DCollection` object, maybe we can have a generic `group_by` which accepts certain special strings (e.g. 'species'), but I imagine in the future it could be extended (e.g. allowing other special strings, or a regex containing a capturing group). Something like `group_by(grouping='species', matches=args.pdos)`.

This is the sort of thing I had in mind when we defined the metadata for Spectrum1DCollection, and is close to what I did for the ASE DOS objects https://gitlab.com/ajjackson/ase/-/blob/master/ase/spectrum/doscollection.py

I'm sure that there is some proper formal scheme out there for this kind of select/merge-by/split-by logic. I think I had the paper open in a Firefox tab once, and lost it with my session history by accidentally closing the single default window after opening a new browser instance. I never found it again, perhaps it's worth another hunt to ensure that the set of operations we provide is sufficient to cover all uses.

I am becoming convinced that this is a compelling use-case for a Spectrum2DCollection; even though it would be unusual to plot more than one, it is useful to be able to select and sum components in this form. And then we can extend plot_2d to plot these things in a vertical stack (rather than superimposed like Spectrum1D) and we have consistent syntax where a list of Spectrum2D continues to mean discontinuous regions on the x-axis.

rebeccafair:
> > ... If we do have to change it, its probably easier to do it sooner rather than later. It might be a bit of work though, maybe we should talk to some others (e.g. Keith) to make sure this method makes sense before making any changes to the code.
> 
> We should talk to Keith about the theory.

Issue created #160

> From an API perspective
>
> > I'm now wondering whether `calculate_qpoint_phonon_modes/frequencies` should return these xyz gradients rather than the final mode widths,
> 
> seems an excellent idea. Again, can we do some deprecation magic for the API breakage? I'd do it by giving the new option a different name, and then an if statement at the top of `calculate_qpoint_phonon_modes`:
> 
> ```python
> def calculate_qpoint_phonon_modes(self, *args, return_mode_widths=False, return_mode_gradients=False, **kwargs):
>     if return_mode_widths:
>         raise DeprecationWarning(...)
>         modes, gradients = calculate_qpoint_phonon_modes(*args, return_mode_gradients=True, **kwargs)
>         return modes, calculate_isotropic_mode_widths(gradients)
>     ...
> ```
> 
> Then at the appropriate version number bump we just need to remove the input argument and that small block, leaving the rest of the method untouched.

Sounds good, issue created #159

rebeccafair:
> > `group_by_element` seems slightly specific for a generic `Spectrum1DCollection` object, maybe we can have a generic `group_by` which accepts certain special strings (e.g. 'species'), but I imagine in the future it could be extended (e.g. allowing other special strings, or a regex containing a capturing group). Something like `group_by(grouping='species', matches=args.pdos)`.
> 
> This is the sort of thing I had in mind when we defined the metadata for Spectrum1DCollection, and is close to what I did for the ASE DOS objects https://gitlab.com/ajjackson/ase/-/blob/master/ase/spectrum/doscollection.py
> 
> I'm sure that there is some proper formal scheme out there for this kind of select/merge-by/split-by logic. I think I had the paper open in a Firefox tab once while working on the ASE version, and lost it with my session history by closing my browser windows in the wrong order. I never found it again, perhaps it's worth another hunt to ensure that the set of operations we provide is sufficient to cover all uses.

Yes, it would be really good to get this right. But at the same time, this is really required for the PDOS feature and time may not be on our side... if we make something general enough hopefully it will be ok, have created a new issue for now #161

> I am becoming convinced that this is a compelling use-case for a Spectrum2DCollection; even though it would be unusual to plot more than one, it is useful to be able to select and sum components in this form. And then we can extend plot_2d to plot these things in a vertical stack (rather than superimposed like Spectrum1D) and we have consistent syntax where a list of Spectrum2D continues to mean discontinuous regions on the x-axis.

Ok, this has probably bumped #158 from a nice-to-have to a necessity for the PDOS feature.

rebeccafair:
As this PR has become slightly complicated, I will probably be closing this PR in favour of multiple smaller PRs once all the features of this PR have been covered. I'll link to those PRs here as they appear.

rebeccafair:
I think doing this work has created some useful discussion and made it clearer on what the PDOS APIs should be, but I'm closing this in favour of multiple other smaller PRs/commits. The following is a summary of how this work has been split up:

* Fix of `euphonic_powder_map` units made onto master instead 00134e8
* Fix of number of `ebins` in CL tools has been made onto master instead f538606
* Adding of `calculate_dos_map` has been moved to a task on the to-do list on #138 https://github.com/pace-neutrons/Euphonic/pull/138#issue-614418854
* Determining how to group/select PDOS has been moved to a task on the to-do list on #138 https://github.com/pace-neutrons/Euphonic/pull/138#issue-614418854 
* Adding of CL tools has been moved to as task on the to-do list on #138 https://github.com/pace-neutrons/Euphonic/pull/138#issue-614418854 
* Overall PDOS implementation has been de-scoped and PDOS maps will be added at a later date, issue #164 created

There have also been some PRs created from issues discussed in this thread #162, #163

Issue #159, opened 2021-05-19
Return Cartesian vector mode gradients instead of scalar widths


rebeccafair:
Based on discussion in #155 it's been found that returning the final mode widths from `calculate_qpoint_phonon_modes`/`calculate_qpoint_frequencies` with `return_mode_widths` is problematic for two main reasons:

* Mode widths are currently calculated based on the number of q-points used in the calculation. This is ok as an estimate for DOS over the 1st BZ, however for powder averaging there will be a different number of q-points in each shell (but potentially a similar density of q-points), so broadening based on the number of q-points may not make sense
* The current method of reducing the vector mode gradients to an average gradient and then scalar mode width may only be applicable to DOS, not Q-E maps where we know the Q-direction

The above cases still need to be investigated, but to give us more flexibility in the future for applying mode-specific broadening for these cases, a solution is instead to return the mode gradients in each Cartesian direction. This can be used to calculate the scalar mode widths, and probably any other quantities we require. The current kwarg `return_mode_widths` should be deprecated and removed in the future.

Issue #143, opened 2021-04-15
Improve mode widths calculation docstring


rebeccafair:
In the docstring for the `ForceConstants.calculate_qpoint_phonon_modes/frequencies` the maths in the description of the mode widths calculation is a bit sloppy, this should be cleaned up.

rebeccafair:
Maths is not perfect but understandable at least, am closing for now

PR #162, opened 2021-05-24
159 Return vector mode gradients

rebeccafair:
Addresses #159 

Some things to note:
-  While doing this work, I found a very small bug in passing the reciprocal ASR correction to the C extension: In the C extension, as the diagonalisation library used is in Fortran, the dynamical matrix is constructed with Fortran ordering, the reciprocal ASR is then added. The reciprocal ASR should also use this same ordering, so needs to be conjugate transposed before being passed to C. This makes no practical difference (you'll notice the other tests all still pass as usual), but only shows up in nearly flat mode gradients, where the C and Python were giving slightly different answers. Rather than just cranking up the tolerance, this has been fixed. This shouldn't really matter as the matrices should be Hermitian, but numerically they aren't exactly.
- The mode gradients are the result of multiplying complex arrays (eigenvectors and dynamical matrix gradients), but the resultant mode gradients should be real. I couldn't find an identity proving this, so for now I've added a test when casting the mode gradients to real values to warn if data is likely to be lost. I haven't been able to trigger this warning with any test cases I've tried, so hopefully we should be good.
- `euphonic.util.mode_gradients_to_widths` accepts either dx, dy, dz mode gradients, or the overall mode gradients magnitude (obtained by `np.linagl.norm` for example). The reason for this is euphonic now calculates the dx, dy, dz gradients, whereas CASTEP `.phonon_dos` files produce the mode gradient magnitudes, and this allows us to easily deal with both cases.
- I don't know what the next release will be, so in the `.. deprecated::` directive I've guessed `0.6.0`. I may add a note to the release documentation to look for those directives and check the versions
- The gradients calculation documentation is still a bit sloppy, I haven't made many changes. But its at least no worse than before and there's an issue to improve it #143


euphonic/force_constants.py
```diff
@@ -621,21 +647,49 @@ def _calculate_phonons_at_qpts(
                         q, q_independent_args)
                 if return_eigenvectors:
                     reigenvecs[q] = evecs
-                if return_mode_widths:
+                if return_mode_gradients:
                     rmode_gradients[q] = grads
```
ajjackson, line 651: Having a little trouble keeping track of the shape of the data through this. Am I right in thinking that
- the einsum in _calculate_phonons_at_q returns a 2D array of modes gradients where the x, y, z components have been serialised along one axis
- these fall back into an x, y, z axis due to the pre-allocated shape of rmode_gradients?

rebeccafair: Going into the einsum we have 3 arrays:
* `np.conj(evecs_sq_view)`, axes labelled `ij`, shape `(n_modes, n_modes)`
* `evecs_sq_view`, axes labelled `ik`, shape `(n_modes, n_modes)`
* `dmat_grad`, axes labelled `jkl`, shape `(n_modes, n_modes, 3)`

The output from the einsum is:
* `mode_grads_xyz`, axes labelled `il`, shape `(n_modes, 3)`

These output gradients get indexed into `rmode_gradients` (shape `(n_rqpts, n_modes, 3)`) for each q-point. Does that make sense?

ajjackson: I think I see now. The `n_modes` on j, k is a combination of atom indices and directions, right? I guess I was confused by the way the cartesian directions in dmat_grad aren't directly connected to the cartesian directions of the atoms. But if that's how the maths works I guess it's ok!

tests_and_analysis/test/euphonic_test/test_force_constants_calculate_qpoint_phonon_modes.py
```diff
@@ -123,6 +124,65 @@ def test_calculate_qpoint_phonon_modes(
...
+            gamma_atol = 0.5
+        check_qpt_ph_modes(qpoint_phonon_modes,
+                           expected_qpoint_phonon_modes,
+                           frequencies_atol=1e-4,
+                           frequencies_rtol=2e-5,
+                           acoustic_gamma_atol=gamma_atol)
+        assert modg.units == expected_modg.units
+        # Mode gradients are derived from eigenvectors - in the case of
+        # degenerate modes they may not be in the same order
+        summed_modg = sum_at_degenerate_modes(
```
ajjackson, line 177: That's a clever way to test as much information as possible!

tests_and_analysis/test/euphonic_test/test_util.py
```diff
@@ -63,28 +66,44 @@ def test_get_qpt_labels(self, qpts, kwargs, expected_labels):
...
+                                      expected_mode_widths_file):
         mode_widths = mode_gradients_to_widths(mode_grads, cell_vecs)
+        with open(os.path.join(get_fc_dir(), expected_mode_widths_file), 'r') as fp:
+            modw_dict = json.load(fp)
+        expected_mode_widths = modw_dict['mode_widths']*ureg(
+            modw_dict['mode_widths_unit'])
         assert mode_widths.units == expected_mode_widths.units
         npt.assert_allclose(mode_widths.magnitude,
-                            expected_mode_widths.magnitude, atol=1e-5)
+                            expected_mode_widths.magnitude, atol=3e-4)
```
ajjackson, line 109: Seems like a magic number? Is this to deal with variation across machines or Numpy versions?

tests_and_analysis/test/utils.py
```diff
@@ -202,3 +202,40 @@ def check_json_metadata(json_file: str, class_name: str):
         data = json.loads(f.read())
     assert data['__euphonic_class__'] == class_name
     assert data['__euphonic_version__'] == __version__
+
+def sum_at_degenerate_modes(values_to_sum, frequencies, TOL=0.05):
```
ajjackson, line 206: What's with the all-caps?

If we're using TOL to indicate some kind of global constant then it would make more sense to define at a higher scope. If it's just a function argument, I don't see a clash with `tolerance`?

euphonic/force_constants.py
```diff
@@ -589,6 +609,12 @@ def _calculate_phonons_at_qpts(
             cell_vectors = self.crystal._cell_vectors
             recip_vectors = self.crystal.reciprocal_cell().to(
                 '1/bohr').magnitude
+            # Get conj transpose - the dynamical matrix calculated in C
+            # is passed to Fortran libs so uses Fortran ordering, make
+            # sure reciprocal correction matches this. This only makes
+            # a very small difference and can only be seen in near-flat
+            # mode gradients, but should be corrected anyway
+            recip_asr_correction =  recip_asr_correction.conj().T
```
ajjackson, line 617: Why is a C-vs-Fortran-ordering problem solved by taking the _conjugate_ transpose?

euphonic/force_constants.py
rebeccafair: Yeah, some codes index it like `(n_atoms, 3, n_atoms, 3)` but I prefer to think of it as a 2D matrix and it makes the einsum stuff easier without having to make different views of the array, but it's just convention I guess

tests_and_analysis/test/utils.py
rebeccafair: Good point, this was left over from the previous version of the function which I wrote quite a while ago, I'll change it.

euphonic/force_constants.py
ajjackson: If `recip_asr_correction` is Hermitian, than taking the complex conjugate _is_ taking the transpose. (And taking both is an identity operation.)

If it is symmetric (as you say) and _not_ Hermitian then this is reversing the phase; is there somewhere else in the maths that this phase reversal should be reflected? Indexing in the "wrong" order for a phase reversal seems like a Hermitian thing.

rebeccafair: The reciprocal ASR correction is derived from the dynamical matrix at gamma, the dynamical matrix is Hermitian, but at gamma it is real (in practice numerically _almost_). To get the reciprocal ASR correction the gamma-point eigenvalues multiplied by the gamma-point eigenvectors and their conjugates of the 3 acoustic modes are subtracted from the gamma-point dynamical matrix to artificially shift the eigenvalues to 0. I'm not sure how this affects the symmetry/Hermicity of the resultant reciprocal ASR correction though... examining the correction this appears to make it exactly symmetric, but I'm not competent enough in the Maths to say why

ajjackson: I see! So in an ideal world none of these operations would do anything but in practice we are shuffling numerical errors around and they need to be lined up with their corresponding elements.

That makes it a bit harder for me to naively reason about how things should behave under transformation :sweat_smile: 
If you're happy that the implementation is now following the described method then all is good.

ajjackson (approved):

Issue #157, opened 2021-05-18
Add Spectrum1DCollection.broaden method


rebeccafair:
Currently broadening each spectrum in a Spectrum1DCollection has to be done manually with `Spectrum1DCollection.from_spectra([spec.broaden(bwidth) for spec in spectra])`

PR #163, opened 2021-05-24
157 Add Spectrum1DCollection.broaden

rebeccafair:
Addresses #157

This has also resulted in some refactoring of all Spectra.broaden methods. I've added some minimal tests for now checking that `Spectrum1DCollection.broaden` produces the same results as calling each `Spectrum1D.broaden` individually. I may add some other tests once the PDOS stuff is in and there are more options for test data.

### DIFF
--- a/euphonic/spectra.py
+++ b/euphonic/spectra.py
@@ -154,16 +154,35 @@ class Spectrum(ABC):
             return self._split_by_indices(indices)
 
     @staticmethod
-    def _gfwhm_to_sigma(fwhm: Quantity, ax_bin_centres: Quantity) -> float:
+    def _broaden_y(x_data: np.ndarray, y_data: np.ndarray,
+                   x_width: float, shape: str = 'gauss') -> np.ndarray:
+        """
+        Broaden y_data by x_width, returning the broadened data. x_data
+        and x_width are assumed to be in the same units, and x_data
+        should be bin centres
+        """
+        if shape == 'gauss':
+            xsigma = Spectrum._gfwhm_to_sigma(x_width, x_data)
+            y_broadened = gaussian_filter1d(y_data, xsigma, mode='constant')
+        elif shape == 'lorentz':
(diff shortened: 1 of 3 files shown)


euphonic/spectra.py
```diff
@@ -644,6 +661,40 @@ def from_castep_phonon_dos(cls: SC, filename: str) -> SC:
...
+        -------
+        broadened_spectrum
+            A new Spectrum1DCollection object with broadened y_data
+
+        Raises
+        ------
+        ValueError
+            If shape is not one of the allowed strings
+        """
+        y_broadened = np.zeros(self._y_data.shape)
```
ajjackson, line 686: A nice numpy shortcut for this (which also gets the dtype right) is `np.zeros_like(self._y_data)`. (Doesn't make much difference here, happy to leave as-is if you prefer.)

euphonic/spectra.py
```diff
@@ -759,39 +810,14 @@ def broaden(self: S2D, x_width: Optional[Quantity] = None,
...
-                    self.get_bin_centres('y').magnitude,
-                    y_width.to(self.y_data_unit).magnitude,
-                    shape=shape)
-                z_broadened = correlate1d(
-                    z_broadened, y_broadening, mode='constant', axis=1)
-        else:
-            raise ValueError(
-                f"Distribution shape '{shape}' not recognised")
+        bin_centres = [self.get_bin_centres(ax).magnitude for ax in ['x', 'y']]
+        bwidths = [None]*2
```
ajjackson, line 814: we do have space to write `bin_widths` here and it would be nicer to read

ajjackson (approved):
This looks good to me. Always nice to have some self-consistency tests in the mix!

euphonic/spectra.py
rebeccafair: Done

euphonic/spectra.py
rebeccafair: Done!

Issue #114, opened 2021-01-04
Rename Eta


rebeccafair:
The argument in `calculate_qpoint_phonon_modes` for describing the balance between real and reciprocal space sums `eta` has been adopted from the CASTEP implementation, and it is not very descriptive, and actually clashes with the notation in the [referenced publication](https://journals.aps.org/prb/pdf/10.1103/PhysRevB.55.10355) which describes it as Λ. This should probably be renamed before a 1.0 release, alongside the `optimise-eta` tool.

rebeccafair:
Closed by #170

PR #170, opened 2021-06-15
Rename Eta

rebeccafair:
Addresses #114 

When Euphonic calculates the Ewald sum, an initial 'guess' is made on what would be a good balance between real/reciprocal space sums (Λ in [Gonze & Lee, 1997](https://journals.aps.org/prb/pdf/10.1103/PhysRevB.55.10355)). As this guess depends on the material and is never going to be perfect, the user can input their own `eta_scale` which is a mulplicative factor on Λ.

For the renaming of the `eta_scale` keyword argument we decided to go for `dipole_parameter`, as it links is obviously with the `dipole` kwarg, and went for the suitably vague word 'parameter', as any other specific word 'ratio', 'fraction' would not be correct and probably misleading.

This kwarg is used in Horace-Euphonic-Interface, so `eta_scale` should also be updated to `dipole_parameter` in `euphonic_sqw_models` and`horace-euphonic-interface`. Doing this led to finding out that the `eta_scale` deprecation warning wasn't being passed through, so also led to an update to the underlying `light_python_wrapper`. The relevant PRs are:

- Light Python wrapper update to allow Python warnings via Matlab: https://github.com/pace-neutrons/light_python_wrapper/pull/2
- Euphonic_sqw_models update to add dipole parameter and allow warning propagation https://github.com/pace-neutrons/euphonic_sqw_models/pull/3
- Horace-Euphonic-Interface update to update the submodules and enable warning propagation from Python https://github.com/pace-neutrons/horace-euphonic-interface/pull/11

This PR and `light_python_wrapper` PR should be resolved first, followed by `euphonic_sqw_models`, and then `horace-euphonic-interface`. Integration tests for the stack are passing at https://anvil.softeng-support.ac.uk/jenkins/job/PACE-neutrons/job/pace-integration/job/Branch-Scientific-Linux-7-2019b/12


euphonic/cli/optimise_dipole_parameter.py
```diff
@@ -0,0 +1,145 @@
+# -*- coding: UTF-8 -*-
+"""
```
cmarooney-stfc, line 2: Noting this is the optimize_eta.py with eta->dipole_parameter. I think it is possible to coerce git into thinking that the rename is part of the git update rather than two separate ops, delete file and new file. Then github will show the compariison.

euphonic/cli/optimise_dipole_parameter.py
```diff
@@ -0,0 +1,145 @@
...
+        print_to_terminal: bool = False):
+    """
+    Calculate the optimum dipole_parameter and other dipole_parameters
+    from the filename castep_bin file
+
+    Parameters
+    ----------
+    filename : str
+        The path and name of the .castep_bin/.check to read from
+    dipole_parameter_min
```
cmarooney-stfc, line 39: Noting that you have removed the default values from the comment her, but also that you have much improved the layout of the def statement making the default implementations clearer. Leaving it up to you as to whether the good pythonista would also like the defaults in the comments.

euphonic/force_constants.py
```diff
@@ -852,13 +867,13 @@ def _dipole_correction_init(self, eta_scale=1.0):
...
         abc_mag = np.linalg.norm(cell_vectors, axis=1)
         mean_abc_mag = np.prod(abc_mag)**(1.0/3)
-        eta = (sqrt_pi/mean_abc_mag)*n_atoms**(1.0/6)
-        # Use eta = lambda * |permittivity|**(1/6)
-        eta = eta*np.power(np.linalg.det(dielectric), 1.0/6)*eta_scale
-        eta_2 = eta**2
+        upper_lambda = (sqrt_pi/mean_abc_mag*n_atoms**(1.0/6)
+                        *np.linalg.det(dielectric)**(1.0/6))
+        upper_lambda *= dipole_parameter
+        lambda_2 = upper_lambda**2
```
cmarooney-stfc, line 876: Now you have used upper_lambda as a name with no indication of how it is upper for eta. A comment may be useful

tests_and_analysis/test/euphonic_test/test_force_constants_calculate_qpoint_frequencies.py
```diff
@@ -37,6 +37,10 @@ def get_quartz_fc():
         (get_quartz_fc(), 'quartz',
          [get_test_qpts(), {'asr': 'reciprocal', 'splitting': False}],
          'quartz_reciprocal_qpoint_frequencies.json'),
+        (get_quartz_fc(), 'quartz',
+         [get_test_qpts(), {'asr': 'reciprocal', 'splitting': False,
+                            'dipole_parameter': 0.75}],
+         'quartz_reciprocal_qpoint_frequencies.json'),
```
cmarooney-stfc, line 43: This looks like some duplication above and below. Also the lines below still have eta_scale in them

tests_and_analysis/test/euphonic_test/test_force_constants_calculate_qpoint_phonon_modes.py
```diff
@@ -59,6 +59,10 @@ def get_quartz_fc():
         (get_quartz_fc(), 'quartz',
          [get_test_qpts(), {'asr': 'reciprocal', 'splitting': False}],
          'quartz_reciprocal_qpoint_phonon_modes.json'),
+        (get_quartz_fc(), 'quartz',
+         [get_test_qpts(), {'asr': 'reciprocal', 'splitting': False,
+                            'dipole_parameter': 0.75}],
+         'quartz_reciprocal_qpoint_phonon_modes.json'),
```
cmarooney-stfc, line 65: ditto duplication and eta scale

cmarooney-stfc (changes requested):
If you don't need changes in fact, a response comment is appreciated.

c/_euphonic.c
```diff
@@ -41,7 +41,7 @@ static PyObject *calculate_phonons(PyObject *self, PyObject *args) {
     // Extra vars only required if dipole = True
     PyArrayObject *py_born;
     PyArrayObject *py_dielectric;
-    double eta;
+    double lambda;
```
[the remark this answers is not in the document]
rebeccafair: Lambda is exactly the same as lambda in the paper, so I think for anyone looking at the paper and trying to understand what's happening lambda will be clearer. It was probably a mistake to ever use eta...

This is different from the user input `dipole_parameter` which is a multiplier on lambda. I can amend the relevant commit to explain this so it will be clear to anyone using git blame in the future.

euphonic/force_constants.py
rebeccafair: I only used upper_lambda as[ lambda](https://docs.python.org/3/tutorial/controlflow.html#lambda-expressions) is a reserved keyword in Python. I can add a comment to explain this.

tests_and_analysis/test/euphonic_test/test_force_constants_calculate_qpoint_frequencies.py
rebeccafair: I feel like it's good practice to keep testing `eta_scale` even though its deprecated until it's been removed, right? Unless deprecation also means, 'not testing anymore, use at your own risk'?

There is a lot of duplication in testing `QpointPhononModes/Frequencies` as they essentially behave the same and use the same underlying code, except one doesn't contain any eigenvectors. It's probably still a good idea to explicitly test both of them though in case we break passing through the arguments through, for example.

euphonic/cli/optimise_dipole_parameter.py
rebeccafair: I've tried doing the `git mv` and editing as two separate commits to try and get git to think it's just been edited, but it turns out this is limited by git's 'similarity index', which will automatically detect a file as new if its been changed by more than 50% - see [git diff docs](https://git-scm.com/docs/git-diff#Documentation/git-diff.txt--Mltngt).

If we do:
`git diff master 114_rename_eta --stat -M -- euphonic\cli\optimise_eta.py euphonic\cli\optimise_dipole_parameter.py`
if shows as a delete & new file, whereas if we manually set the threshold to 40% similarity with `-M40%`
`git diff master 114_rename_eta --stat -M40% -- euphonic\cli\optimise_eta.py euphonic\cli\optimise_dipole_parameter.py`

It shows as a single modified file. Interesting, but nothing can be done about it I don't think!

rebeccafair:
Ok I think that's all changes made @cmarooney-stfc

Issue #166, opened 2021-05-26
Citations and contributors


rebeccafair:
Euphonic should have a recommended citation, and a way of crediting all of its authors/contributors, some options for this are:

- Citation file format (CFF) https://citation-file-format.github.io/
- ~~Investigate what the RAL Library recommends~~ - this had led me to SCD's Software Engineering Group, who also recommend CFF and can give us version-specific DOIs
- CONTRIBUTORS.md https://github.com/cezaraugusto/github-template-guidelines/blob/master/.github/CONTRIBUTORS.md could be a good way of crediting other contributions to the project

rebeccafair:
Closed by f074c62

PR #171, opened 2021-06-15
Add CITATION.cff

rebeccafair:
Addresses #166 

To make it easy to cite Euphonic and give credit to all contributors, given the many recommendations it is probably sensible to go with the Citation File Format (https://github.com/citation-file-format/citation-file-format, http://doi.org/10.5281/zenodo.1117789). Things to discuss (probably elsewhere, but this PR serves as an example):

 - `authors` I've added myself as an initial example, but we should decide what the full list will be and what metadata we will try to add for each author (`orcid` might be nice but I don't have one myself yet)
 - `doi` Needs to be requested from the Software Engineering Group. We have the option of an overall DOI and version-specific DOIs or any combination of both/either. We should decide what we want, and what our landing page could be (Github repo? Github pages? Euphonic docs?)
 - `url` This may be different to `repository`, it could be the DOI landing page?
 - `abstract` What's our one-liner to describe Euphonic?
 - `keywords` What keywords do we want? 
 - `references` It would be sensible to add the Euphonic paper here once it exists. Do we want/need to reference anything else?
 - Are there any other keys in the specification that should be included?
 - Should `CITATION.cff` be distributed/installed with Euphonic? I can't find any best practice for this, and cff-related Python packages don't do it themselves (e.g. [cff-converter](https://github.com/citation-file-format/cff-converter-python)) but I feel like it might be a good idea
 - Would it be useful to provide a function to return the citation (e.g. `euphonic.get_citation`?). It could just be a thin wrapper around [cff-converter](https://github.com/citation-file-format/cff-converter-python))


rebeccafair:
I think this is mostly done, so its probably time for a light review from an appropriate subset of people. I'm still waiting on a DOI, but I don't think that alone should delay everything, it can always be added later.

Issue #168, opened 2021-06-04
Change S(Q,w) map units


rebeccafair:
As discussed in #138 to be consistent with neutron-weighted PDOS, S(Q,w) as produced by `calculate_sqw_map` should be in area/energy units

PR #174, opened 2021-06-21
Update S(Q,w) units

rebeccafair:
Addresses #168

- As discussed in #138 the units of DOS will be changed to `1/energy`, and units of neutron-weighted PDOS will be `mbarn/energy` per sample atom as part of the PDOS feature
- For consistency then the units of `StructureFactor.structure_factors` should also be changed to `mbarn` from `angstrom**2`, this also makes their magnitudes a lot easier to handle, they're on the order of `0.1 mbarn`, compared to `1e-12 angstrom**2`
- As we're now using absolute rather than relative units (so any prefactors shouldn't be neglected) and we're moving from per unit cell to per sample atom, `StructureFactor.structure_factors` will be multiplied by a factor of `1/(2*N_atoms)` relative to before
- The units of S(Q,w) map as produced by `StructureFactor.calculate_sqw_map` will now have default units of `mbarn/energy` rather than `angstrom**2`

This will also change the scaling of the data in the `euphonic_sqw_models`, `Horace-Euphonic-interface` and the `pace-integration` tests so will need updates to those repositories. PRs are:

https://github.com/pace-neutrons/euphonic_sqw_models/pull/4
https://github.com/pace-neutrons/horace-euphonic-interface/pull/12
https://github.com/pace-neutrons/pace-integration/pull/3

Full integration tests using this branch and `update_sf_units` branches of pace-integration, euphonic_sqw_models and horace-euphonic-interface are passing:
https://anvil.softeng-support.ac.uk/jenkins/job/PACE-neutrons/job/pace-integration/job/Branch-Scientific-Linux-7-2019b/13


euphonic/qpoint_phonon_modes.py
```diff
@@ -265,10 +272,11 @@ def calculate_structure_factor(
         # Take mod squared and divide by frequency to get intensity
         sf = np.real(
             np.absolute(term*np.conj(term))/np.absolute(self._frequencies))
+        sf = sf/(2*self.crystal.n_atoms)
```
cmarooney-stfc, line 275: I think that python has divide and assign /= operator, so `sf /= (2*....` could be neater.

euphonic/qpoint_phonon_modes.py
```diff
@@ -309,18 +317,20 @@ def calculate_debye_waller(
...
+        Note that internally Euphonic uses atomic units so :math:`\\hbar`
+        has been omitted from the formulation:
 
         .. math::
 
           W^{\\kappa}_{\\alpha\\beta} =
-          \\frac{1}{4M_{\\kappa}\\sum_{q}{weight_q}}
-          \\sum_{q\\nu}weight_q\\frac{\\epsilon_{q\\nu\\kappa\\alpha}\\epsilon^{*}_{q\\nu\\kappa\\beta}}
+          \\frac{1}{4M_{\\kappa}\\sum\\limits_{q}{weight_q}}
+          \\sum_\\limits{q\\nu}weight_q\\frac{\\epsilon_{q\\nu\\kappa\\alpha}\\epsilon^{*}_{q\\nu\\kappa\\beta}}
```
cmarooney-stfc, line 333: Have you got the subscript _ in the right place? Previous lines have \sum\limits_.

tests_and_analysis/test/euphonic_test/test_calculate_structure_factor.py
```diff
@@ -79,9 +79,9 @@ def test_calculate_structure_factor(self, material, qpt_ph_modes,
             sf = qpt_ph_modes.calculate_structure_factor()
         sf_file = os.path.join(get_sf_dir(material), expected_sf_file)
         expected_sf = sf.from_json_file(sf_file)
-        check_structure_factor(sf, expected_sf, sf_atol=8e-14, sf_rtol=3e-4,
+        check_structure_factor(sf, expected_sf, sf_atol=4e-4, sf_rtol=3e-4,
                                freq_atol=1e-4, freq_rtol=2e-5,
-                               freq_gamma_atol=0.5, sf_gamma_atol=1e-7)
+                               freq_gamma_atol=0.5, sf_gamma_atol=6e2)
```
cmarooney-stfc, line 84: Appreciating that this is test rather than production code, but this makes me a bit nervous of raw numbers in what I guess is a unit conversion process. As we change the unit system, is this a permanent once-for-all change or should we  be providing unit-free code into which unit-ed numbers can be put?

cmarooney-stfc (changes requested):
NB I have not reviewed the json.

euphonic/qpoint_phonon_modes.py
rebeccafair: Good spot, corrected.

tests_and_analysis/test/euphonic_test/test_calculate_structure_factor.py
rebeccafair: Yes, this could be a problem... I don't have time to fix it now but I've made an issue for it https://github.com/pace-neutrons/Euphonic/issues/175

PR #176, opened 2021-06-25
Add widget for changing intensity to powder-map plots

rebeccafair:
After playing with the powder-map script for the workshop, I was finding it painful to have to regenerate plots just to change the intensities, so I've added some widgets to allow putting in min/max intensity without having to recalculate all the q-points. It seems to work well enough (although I can't get it to work via a Jupyter notebook).

I've tried for a while and couldn't find a good way to test it yet, but maybe that could be for another PR, its only on `euphonic-powder-map` and can be disabled so any damage should be minimal.

### DIFF
--- a/euphonic/cli/powder_map.py
+++ b/euphonic/cli/powder_map.py
@@ -46,8 +46,8 @@ def get_parser() -> 'argparse.ArgumentParser':
                                help="Minimum |q| in 1/LENGTH_UNIT")
     sections['q'].add_argument('--q-max', type=float, default=3., dest='q_max',
                                help="Maximum |q| in 1/LENGTH_UNIT")
-    sections['plotting'].add_argument('--no-widgets', action='store_false',
-                                      dest='use_widgets', default=True,
+    sections['plotting'].add_argument('--no-widgets', action='store_true',
+                                      dest='disable_widgets', default=False,
                                       help=("Don't use Matplotlib widgets to "
                                             "enable interactive setting of "
                                             "colourmap intensity limits"))
@@ -170,14 +170,14 @@ def main(params: List[str] = None):
                                 y_label=y_label,
                                 title=args.title)
 
-    if args.use_widgets == True:
+    if args.disable_widgets is False:
         # TextBox only available from mpl 2.1.0
         try:
             from matplotlib.widgets import TextBox
         except ImportError:
-            args.use_widgets = False
+            args.disable_widgets = True
 
-    if args.use_widgets:
+    if args.disable_widgets is False:
         min_label = f'Min Intensity ({spectrum.z_data.units:~P})'
         max_label = f'Max Intensity ({spectrum.z_data.units:~P})'
         boxw = 0.15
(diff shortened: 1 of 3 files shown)


euphonic/cli/powder_map.py
```diff
@@ -44,6 +44,11 @@ def get_parser() -> 'argparse.ArgumentParser':
                                help="Minimum |q| in 1/LENGTH_UNIT")
     sections['q'].add_argument('--q-max', type=float, default=3., dest='q_max',
                                help="Maximum |q| in 1/LENGTH_UNIT")
+    sections['plotting'].add_argument('--no-widgets', action='store_false',
+                                      dest='use_widgets', default=True,
```
ajjackson, line 48: Confusingly, the `-h` help message for this argument says "default: True". It might be clearer to name the argument "no_widgets" and store_true?

ajjackson:
This is a brilliant improvement in usability.

I have run into one frustrating and probably solvable problem; after editing a number, I instinctively hit the "tab" key to change focus to the next field and update the display. Instead what happens is a TypeError appears in the log and the figure stops responding to keyboard input. There should be some way we can capture this and at least stop it from hanging.

euphonic/cli/powder_map.py
rebeccafair: Good point, fixed.

euphonic/cli/powder_map.py
```diff
@@ -144,12 +149,48 @@ def main(params: List[str] = None):
...
+    if args.use_widgets == True:
+        # TextBox only available from mpl 2.1.0
+        try:
+            from matplotlib.widgets import TextBox
+        except ImportError:
+            args.use_widgets = False
+
+    if args.use_widgets:
+        min_label = f'Min Intensity ({spectrum.z_data.units:~P})'
+        max_label = f'Max Intensity ({spectrum.z_data.units:~P})'
```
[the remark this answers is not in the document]
rebeccafair: Yeah, hopefully this will be fixed once DOS units are sorted. I'll wait until the PDOS stuff is merged before merging this just to make sure.

rebeccafair:
> I have run into one frustrating and probably solvable problem; after editing a number, I instinctively hit the "tab" key to change focus to the next field and update the display. Instead what happens is a TypeError appears in the log and the figure stops responding to keyboard input. There should be some way we can capture this and at least stop it from hanging.

That's a pain! After looking into it apparently it is a Tk backend specific issue so I didn't run into this locally (it sounds like its related to this https://github.com/matplotlib/matplotlib/issues/19744). It appears to be 'fixed' in Matplotlib 3.4. Pressing tab no longer causes an error, but it appears to make the box unclickable/typeable until you tab onto it again. At least it doesn't break the plot I guess.. Not sure there's much we can do about this without knowing more about Tk.

ajjackson:
Ok, I guess we can live with it then. Is the plan to wait for PDOS MR before merging this one?

rebeccafair:
Can confirm: janky units fixed with merge of #138 
image

Issue #165, opened 2021-05-25
Python type-hinting


ajjackson:
Raised by @DBJochym in a meeting: we are currently using Python's type-hinting in the Euphonic code-base, but not entirely consistently. It would be helpful to document the project's attitude towards this and outline a policy for contributed code if appropriate.

My current thoughts on this are that while type-hinting is an expressive way of writing and documenting Python functions, the static analysis tool Mypy does not work very usefully with Numpy arrays and Pint quantities. As these constitute a significant fraction of the data passed around Euphonic, I would be reluctant to impose an "all code should be annotated" or "mypy must pass" policy. But maybe if we think about this a little more there is another useful policy we can consider. At the very least we should mention our preferences in CONTRIBUTING.

@rebeccafair

rebeccafair:
Just updating some of the things we've touched on in other discussions here.

The generic use of `Quantity` and `np.ndarray` for typing is not ideal as it omits shape, type (float/int) and units information, but there are signs from the pint developers this may improve in the future in Pint issues 778, 1166 and currently there is a Pint PR in progress 1259. This would probably require bumping the versions of Pint/Numpy we depend on, but maybe Mantid's version restrictions will also have moved on by then so it might not be an issue. For now we should continue using `Quantity` and `np.ndarray` and putting shape, type and unit info in the docstring.

One other issue is that we currently have a bit of inconsistency in our use of `TypeVar`. `QpointFrequencies` and `QpointPhononmodes` use a generic `T` symbol to indicate 'whatever this class is', which works when subclassing because you get a generic `T` in the docs, rather than something contradictory. Whereas `Spectrum`, `Spectrum1D` etc. use `S`, `S1D` which means that in the `Spectrum1D` subclass you get a mix of `S` and `S1D` depending on which methods are inherited, which is a bit confusing. I would propose to move everything to use a generic `T` to avoid this.

In any case, Euphonic should be updated so all classes/methods/functions handle typing consistently, which would be an improvement, as currently some functions have no typing at all.

Whether we impose any hard restrictions on linting/type checking and what tools we'll use should maybe be another discussion. There are some brief Euphonic style guidelines here which could be added to https://github.com/pace-neutrons/pace-developers/blob/master/euphonic/development/03_style_guidelines.md.

PR #183, opened 2021-08-06
Improve type hinting

rebeccafair:
Addresses #165. I used mypy to guide this, but eliminating all (or even most) mypy errors is nearly impossible with our current dependencies, so I largely ignored errors to do with Quantity objects, for example.

There are a couple of things we can do to improve things in the future once we no longer have to support Python 3.6:
* Quantity type hinting - Pint PR 1259 has been merged and is set to be included in Pint 0.18 - it looks like it lets you hint something like `Quantity[np.ndarray]`, it doesn't solve everything but is a step in the right direction. Only problem is this Pint version also drops support for Python 3.6.
* In 3.7 we can use https://docs.python.org/dev/whatsnew/3.7.html#pep-563-postponed-evaluation-of-annotations to reference classes and do away with the `TypeVar` stuff
* For functions with multiple signatures (e.g. `calculate_qpoint_phonon_modes` that can return `QpointPhononModes` or `Tuple[QpointPhononModes, Quantity]`), we can use [Literal](https://mypy.readthedocs.io/en/stable/literal_types.html). This is available with 3.6 via `typing_extensions` but I wasn't sure whether it was worth it to add another Euphonic dependency just for this, and it would require duplicating the lengthy function signatures, so I'm not 100% on this yet.

In this PR I also fixed a few other small things:
* Reduced the number of arguments to `ForceConstants._calculate_phonons_at_q` - the q-direction is now calculated beforehand so simplifies the function signature.
* Added tests for reading a Phonopy BORN file when there is no factor on the first line. This came from looking more closely at the mypy errors for this function and I ended up doing a bit of refactoring and looking at other examples of BORN files.


ajjackson (commented):
Comments batch so far. (I have got up to the end of phonopy.py)

Basically this is all really good stuff and nothing looks very break-y

euphonic/cli/dos.py
```diff
@@ -11,7 +12,7 @@
                     _arrange_pdos_groups)
 
 
-def main(params: List[str] = None):
+def main(params: List[str] = None) -> None:
```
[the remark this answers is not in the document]
rebeccafair: Good point, have updated this on all the CL tools

euphonic/force_constants.py
```diff
@@ -417,8 +432,8 @@ def calculate_qpoint_frequencies(
     def _calculate_phonons_at_qpts(
             self,
             qpts: np.ndarray,
-            weights: np.ndarray,
-            asr: str,
+            weights: Optional[np.ndarray],
+            asr: Optional[str],
```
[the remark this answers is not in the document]
rebeccafair: At least it's an underscored function so we're probably ok for now

euphonic/force_constants.py
```diff
@@ -719,43 +755,102 @@ def _calculate_phonons_at_qpts(
...
+        unique_sc_origins
+            A list containing 3 lists of the unique supercell image
+            offsets in each direction. The supercell offset is
+            calculated by multiplying the supercell matrix by the
+            supercell image indices (obtained by _get_all_origins()). A
+            list of lists rather than a Numpy array is used as the 3
+            lists are independent and their size is not known
+            beforehand
+        unique_sc_i
+            Shape ((2*lim + 1)**3, 3) int ndarray. The indices needed
```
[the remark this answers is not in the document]
rebeccafair: I've renamed it to `n_sc_shells` and added some description, hopefully that helps

euphonic/force_constants.py
```diff
@@ -792,38 +894,47 @@ def _calculate_dyn_mat(self, q, fc_img_weighted, unique_sc_origins,
...
+            each direction. A list of lists rather than a Numpy array
+            is used as the 3 lists are independent and their size is
+            not known beforehand
+        unique_cell_i
+            Shape (n_cells_in_sc, 3) int ndarray. The indices needed to
+            reconstruct cell_origins from the unique values in
+            unique_cell_origins
+        all_origins_cart
+            Shape (n_cells_in_sc, n_atoms, n_atoms, n_images, 3) or
+            (0, 3) float ndarray. The Cartesian coordinate of each atom
```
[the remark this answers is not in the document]
rebeccafair: I can't spot the `Optional` in the function signature, am I being blind?

euphonic/validate.py
```diff
@@ -138,33 +144,37 @@ def _ensure_contiguous_args(*args):
...
-    args = list(args)
-    for i in range(len(args)):
-        args[i] = args[i].astype(_get_dtype(args[i]), order='C', copy=False)
+    for i, arg in enumerate(args):
+        contiguous_arg = arg.astype(
+            _get_dtype(arg), order='C', copy=False)
+        if i == 0:
+            contiguous_args = [contiguous_arg]
+        else:
+            contiguous_args.append(contiguous_arg)
```
ajjackson, line 161: Would

```python
return tuple(arg.astype(_get_dtype(arg), order='C', copy=False)
             for arg in args)
```

be equivalent here?

ajjackson (commented):
Comments pt 2/2

euphonic/spectra.py
```diff
@@ -93,12 +89,12 @@ def from_json_file(cls: S, filename: str) -> S:
...
     @abstractmethod
-    def _split_by_indices(self: S, indices: Union[Sequence[int], np.ndarray]
-                          ) -> List[S]:
+    def _split_by_indices(self, indices: Union[Sequence[int], np.ndarray]
+                          ) -> List[T]:
         """Split data along x axis at given indices"""
         ...
 
-    def _split_by_tol(self: S, btol: float = 10.0) -> List[S]:
+    def _split_by_tol(self, btol: float = 10.0) -> List[T]:
```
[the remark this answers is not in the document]
rebeccafair: I read the recommendation that self doesn't need to be type-hinted everywhere, but didn't think about cases like this. Fixed.

euphonic/util.py
```diff
@@ -284,7 +306,7 @@ def _cell_vectors_to_volume(cell_vectors: np.ndarray) -> float:
 
 def _get_unique_elems_and_idx(
         all_elems: Sequence[Tuple[Union[int, str]]]
-        ) -> 'OrderedDict[Union[str, int, tuple], np.ndarray]':
+        ) -> 'OrderedDict[Tuple[Union[int, str]], np.ndarray]':
```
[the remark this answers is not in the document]
rebeccafair: Whoops, that's supposed to be any number of elements. Added `...`

euphonic/validate.py
rebeccafair: Indeed it would!

euphonic/spectra.py
```diff
@@ -120,18 +116,19 @@ def _ranges_from_indices(indices: Union[Sequence[int], np.ndarray]
...
         if x_tick_labels is None:
             return None
         else:
             return [(int(x - x0), label) for (x, label) in x_tick_labels
                     if (x >= x0) and ((x1 is None) or (x < x1))]
 
-    def split(self: S, indices: Union[Sequence[int], np.ndarray] = None,
-              btol: float = None) -> List[S]:
+    def split(self, indices: Union[Sequence[int], np.ndarray] = None,
+              btol: float = None) -> List[T]:
```
ajjackson, line 131: I think this another case where it should be `self: T`?

euphonic/spectra.py
```diff
@@ -336,7 +337,7 @@ def __init__(self, x_data: Quantity, y_data: Quantity,
         self.x_tick_labels = x_tick_labels
         self.metadata = {} if metadata is None else metadata
 
-    def __add__(self, other: S1D):
+    def __add__(self, other: T) -> T:
```
ajjackson, line 340: Looking at the implementation, I think this one is actually guaranteed to return Spectrum1D even if inherited? It doesn't look like it would be based on the type of `other` directly.

euphonic/spectra.py
```diff
@@ -349,9 +350,9 @@ def __add__(self, other: S1D):
         spec_col = Spectrum1DCollection.from_spectra([self, other])
         return spec_col.sum()
 
-    def _split_by_indices(self: S1D,
+    def _split_by_indices(self,
```
ajjackson, line 353: `self: T`?

euphonic/spectra.py
```diff
@@ -423,11 +424,11 @@ def from_castep_phonon_dos(cls: S1D, filename: str,
...
         metadata['label'] = element
-        return Spectrum1D(data['dos_bins']*ureg(data['dos_bins_unit']),
-                          data['dos'][element]*ureg(data['dos_unit']),
-                          metadata=metadata)
+        return cls(data['dos_bins']*ureg(data['dos_bins_unit']),
+                   data['dos'][element]*ureg(data['dos_unit']),
+                   metadata=metadata)
 
-    def broaden(self: S1D, x_width: Quantity, shape: str = 'gauss') -> S1D:
+    def broaden(self, x_width: Quantity, shape: str = 'gauss') -> T:
```
ajjackson, line 431: `self: T`

euphonic/spectra.py
```diff
@@ -555,11 +560,11 @@ def __add__(self, other: SC):
         spectra are retained in the top level dictionary, any
         others are put in the individual 'line_data' entries
         """
-        return Spectrum1DCollection.from_spectra([*self, *other])
+        return type(self).from_spectra([*self, *other])
```
ajjackson, line 563: Function signature says it uses type of `other`, implementation says it uses type of `self`

ajjackson (commented):
I found a few more! (I think there are others below, posting comments now to avoid redundant parallel effort.)

euphonic/spectra.py
rebeccafair: Yes, maybe that is misleading. Would `(self: T, other: T) -> T` be better? We can't use `Spectrum1D` directly because Py 3.6 doesn't support forward references

ajjackson: `from_spectra` explicitly requires that the input must all be Spectrum1D. Can we not forward-reference a type hint using quotes? So it becomes
```python
def __add__(self, other: 'Spectrum1D') -> 'Spectrum1D'
```

euphonic/spectra.py
rebeccafair: Is this alleviated if we do `__add__(self: T, other: T) -> T`, now they're all explicitly the same type?

ajjackson: Perhaps it is still best to annotate it as `__add__(self: T, other: T) -> T` if we want to hide the fact that any iterable will work. It is an implementation detail that might get optimised. The inheritance scenario would still work, come to think of it.

rebeccafair: Mypy doesn't seem to complain either way, we can go with `Iterable`. Typing is hard.

ajjackson (approved):
The tests pass and the only functional changes are tested. So it should be fairly low-risk. Still, I did my best to find things that might impact behaviour and didn't find anything scary.

This is a massive cleanup and should make things easier to navigate and understand. :+1:

PR #184, opened 2021-08-10
Add conda-forge installation to index.rst

### LINKED ISSUE #156: Add Conda installation to documentation

ajjackson:
- Installation instructions currently live in index.rst

Hopefully this will help avoid frustration where e.g. pyyaml is not installed.

These dependencies are not included in the package so that a "lightweight" Euphonic dependency is possible for packages that bring their own force constants and make their own plots.

Closes #156

### DIFF
--- a/doc/source/index.rst
+++ b/doc/source/index.rst
@@ -52,7 +52,7 @@ To create a "complete" installation in a new environment:
 
 .. code-block:: bash
 
-  conda create -n euphonic-forge -c conda-forge python=3.8 euphonic matplotlib-base pyyaml tqdm
+  conda create -n euphonic-forge -c conda-forge python=3.8 euphonic matplotlib-base pyyaml tqdm h5py
 
 This creates an environment named "euphonic-forge", which can be
 entered with ``activate euphonic-forge`` and exited with
--- a/doc/source/index.rst
+++ b/doc/source/index.rst
@@ -40,6 +40,28 @@ If you don't require plotting or reading of Phonopy files, just use:
 
   pip install euphonic
 
+Conda
+-----
+
+For Conda users there is a ``euphonic`` package on conda-forge.
+This may be preferable to Pip installation as it should grab existing
+binary-compatible builds of Euphonic and its dependencies.
+The package is quite minimal, it is usually recommended to also
+install plotting, import and progress-bar dependencies.
+To create a "complete" installation in a new environment:
+
+.. code-block:: bash
+
+  conda create -n euphonic-forge -c conda-forge python=3.8 euphonic matplotlib-base pyyaml tqdm
+
+This creates an environment named "euphonic-forge", which can be
+entered with ``activate euphonic-forge`` and exited with
+``deactivate``. Once could also install the packages into an existing environment:
+
+.. code-block:: bash
+
+  conda install -c conda-forge euphonic matplotlib-base pyyaml tqdm
+
 Github
 ------
 To get the latest unreleased version, clone the Git repository at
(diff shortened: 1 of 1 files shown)


doc/source/index.rst
```diff
@@ -40,6 +40,28 @@ If you don't require plotting or reading of Phonopy files, just use:
...
+For Conda users there is a ``euphonic`` package on conda-forge.
+This may be preferable to Pip installation as it should grab existing
+binary-compatible builds of Euphonic and its dependencies.
+The package is quite minimal, it is usually recommended to also
+install plotting, import and progress-bar dependencies.
+To create a "complete" installation in a new environment:
+
+.. code-block:: bash
+
+  conda create -n euphonic-forge -c conda-forge python=3.8 euphonic matplotlib-base pyyaml tqdm
```
rebeccafair, line 55: This doesn't mention h5py, what's the plan for that? I've tried installing like this on my Windows laptop with h5py tacked on the end and it seems to work and pass tests....

But h5py is only required for a fairly small subset of use cases, so maybe we should just leave it as is and mention installing h5py at the end if you want to read Phonopy .hdf5 files?

rebeccafair (changes requested):

doc/source/index.rst
ajjackson: Good point, if we claim to support h5py for Phonopy then that should be included here as well.

ajjackson: Ok, here is a dockerfile that installs these deps including h5py and runs the tests when the container is run. I'm getting a few failures mostly small numerical differences :disappointed: 

```
FROM mambaorg/micromamba
USER root
RUN apt-get update && \
    apt-get install -y wget zip

USER micromamba
RUN micromamba install -y -n base -c conda-forge python=3.8 euphonic matplotlib-base pyyaml tqdm h5py
WORKDIR /home/micromamba
RUN wget https://github.com/pace-neutrons/Euphonic/archive/refs/heads/master.zip \
    && unzip master.zip
RUN ls
RUN python -m pip install -r Euphonic-master/tests_and_analysis/tox_requirements.txt

CMD pytest --import-mode=append Euphonic-master/tests_and_analysis/test/euphonic_test/
```
The remaining error is an odd one, when calling `ForceConstants.from_phonopy()`:
```
_ ERROR at setup of TestForceConstantsCreation.test_correct_object_creation[create_from_phonopy-create_from_phonopy1] _

...

        summary_dict = {}
>       pu = summary_object['physical_unit']
E       KeyError: 'physical_unit'

/opt/conda/lib/python3.8/site-packages/euphonic/readers/phonopy.py:605: KeyError
```
Is that related to the Born charge units thing in #183?

rebeccafair: Ah that last error will be due to 5e96067bdeb23cd82a3eebda53d4527ed3577bc5, so if you run the latest master tests against release 0.6.1 you'll get that failure as the thing it's testing for isn't in 0.6.1

I'm actually investigating moving our tests to github actions at the moment, this includes moving the release testing https://github.com/pace-neutrons/Euphonic/actions/workflows/test_release.yml (Ok so the Windows builds are failing right now but I'm working on it...). It checks out the same version of the tests as the release version so hopefully these sort of failures shouldn't happen. Once I've figured it out hopefully it should be easy to add conda tests too.

ajjackson: Ok! We should keep an eye on those numerical wobbles, but I don't suppose they matter for this PR.

rebeccafair, line 55: `h5py` is also needed here?

ajjackson: d'oh

rebeccafair (approved):

Issue #178, opened 2021-07-01
Minor docs issues


ajjackson:
Some unrelated issues that were noticed while reviewing #177 

- There are code blocks in the documentation which should have Bash syntax-highlighting but have Python syntax-hightlighting. This is caused by the defaul behaviour of the `::` codeblock RST shortcut, which could be overriden.
- `|q|` shows up as a link to nowhere in the CLI docs for euphonic-powder-map. Probably the wrong number of backticks.
- The left-hand menu on RTD points to subsections of tutorials.html. This is really more of an index than a "Tutorials" page. "Changelog" at least should skip this page and go straight to the actual Changelog.

PR #186, opened 2021-08-16
178 Fix minor docs issues

rebeccafair:
Addresses #178 

New docs can be viewed at https://euphonic.readthedocs.io/en/178_docs_issues/

In trying to fix the left menu options, I ended up just refactoring the whole front page, and splitting off the installation and citation into their own pages, which I think is much better. Also out of interest the reason |q| was showing up as a link was due to Sphinx [substitutions](https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#substitutions). I've managed to use that to literally display |q| without having to put any nasty looking escapes in the docstrings themselves. I also fixed any minor formatting issues that I noticed along the way.


ajjackson (approved):
This looks like a big improvement.

Nice trick using a pre-processor to clean up |q|! I guess that could lead to a discrepancy between autodocs in the sphinx docs vs other renderings of the docstrings, but that seems like a non-issue for us.

Marking the main pages as Python API makes more sense than as tutorials. I think the left menu is still a _little_ cluttered/inconsistent and will become harder to use as the code expands. We could pare it back to just

- Installation
- Units in Euphonic
- Python API
- Command-line tools
- Changelog
- Citing Euphonic

But the current version is bearable (and handy!) so I don't really mind if that is deferred for a future cleanup as things expand further. Maybe we could at least put the command-line tools as one entry now? They look a bit out-of-place with their literal (and lower-case) names.

rebeccafair:
Ok, I think I've fixed the issue of having too many things in the sidebar. It was a bit of a pain and I had to split CL tools and API into separate files, but I at least I think I'm starting to understand how Sphinx sidebars work...

ajjackson:
Looks good!

Issue #180, opened 2021-07-09
Naming of plot kwargs


rebeccafair:
Currently for `plot_1d/2d` we have kwargs like `y_label`, `x_label`, whereas the equivalent in Matplotlib are `xlabel`/`ylabel`. We should consider whether its worth the effort to move to using names more consistent with Matplotlib.

This also affects command-line arguments such as `--x-label`, `--y-label` which potentially could be more convenient if they were `--xlabel`, `--ylabel`. But this could then cause clashes with other arguments such as `--e-max`.

PR #188, opened 2021-08-20
Styling: add --style and --no-base-style to CLI plotters

### LINKED ISSUE #185: Plotting customisation
I think I mentioned this previously but can't find an issue for it; we should provide a way for command-line users to customise the fonts, figure size, subplot spacing, tick styles etc. for publication-quality plots. Matplotlib provides a format for this: [matplotlibrc file](https://matplotlib.org/stable/tutorials/introductory/customizing.html#customizing-with-matplotlibrc-files).

In [sumo](https://github.com/smtg-ucl/sumo) we make good use of matplotlib's cascading style system. Euphonic's default plots are bare-bones enough that we probably don't need to worry about that too much, but if we provide an argument that accepts inbuilt styles and matplotlibrc files then we don't have to add 100 different options for fine details in the order people request them.

As I'd like to make some publication-ready plots now I'll try hacking a Euphonic branch to support this and see how complex it gets. Are there any other problems I should consider? Or options that really should be settable from CLI without writing a config file?

ajjackson:
Matplotlib allows several stylesheets to be overlaid. To get the best
use of this, we should remove any explicit styling from plot functions
and replace them with style contexts. Then our settings can take a
higher priority than MPL defaults, but a lower priority than any user
style preferences. Styles can be selected by inbuilt names and from
user files.

This allows the size and formatting of plots to be customised in a
precise and consistent way for publication. It also simplifies
plotting functions by removing options that can be set by the context.

TODO
- [x] CLI options for font and text size
- [x] CLI argument for figure size
- [x] CLI argument for linewidth
- [x] Make the tests pass
- [x] Mention something in the user docs

Closes #185


ajjackson:
Screenshot from 2021-08-20 17-21-16

Codacy being a bit unreasonable as usual.

1. I'm not quite sure what the case convention is for type variables, a lot of them are in title case. Codacy doesn't understand that "MPLStyle" _is_ title-case for something containing an acronym, and instead thinks it's a malformed caps-case.

2. Flake8 shouts at me if I try to follow the recommendation. Which is good, because the recommendation is ugly. I can find an acceptable compromise (8-spaces hanging indent the arguments), but it's still a bit ugly.

3. I don't agree that a punctuation mark is required for one-line docstrings. At the very least an interrobang should be allowed.

ajjackson:
Ah I see, it's worse than that. Codacy thinks that any variable defined at module level is a CONSTANT and so SHOULD_BE_IN_CAPS_LIKE_A_C_MACRO_OR_SOMETHING

PEP8 does _mention_ this but I don't interpret it to mean that this is something that strictly _should_ be done.
https://www.python.org/dev/peps/pep-0008/#constants

Is there a way to turn that warning off?

rebeccafair:
I wouldn't worry too much about the Codacy stuff at the moment, I've been messing with it intermittently while trying to get the github actions stuff working, and I didn't realise it would start working on all PRs straight away... Feel free to use the info if it's useful or ignore it if not, I'll work on tuning up the settings soon!

ajjackson:
If you ever wished you could specify your figure size in Bohr radii, today is your lucky day.

ajjackson:
Test failures seem to be something related to tick labels, will take a look next week.

ajjackson:
I don't understand these test failures. Tests which use `xticklabels` are failing with messages like

```
            else:
>               assert value == expected_image_data[key]
E               AssertionError: assert ['$\\Gamma$', 'A', 'L', 'H', 'A', 'L', 'M', 'H', 'K', '', '', '', '', '', ''] == ['$\\Gamma$',\n 'A',\n 'L',\n 'H',\n 'A',\n 'L',\n 'M',\n 'H',\n 'K',\n '0.0',\n '0.2',\n '0.4',\n '0.6',\n '0.8',\n '1.0']
E                 At index 9 diff: '' != '0.0'
E                 Full diff:
E                   [
E                    '$\\Gamma$',
E                    'A',
E                    'L',
E                    'H',
E                    'A',
E                    'L',
E                    'M',
E                    'H',
E                    'K',
E                 -  '0.0',
E                 -  '0.2',
E                 -  '0.4',
E                 -  '0.6',
E                 -  '0.8',
E                 -  '1.0',
E                 +  '',
E                 +  '',
E                 +  '',
E                 +  '',
E                 +  '',
E                 +  '',
E                   ]

tests_and_analysis/test/script_tests/test_intensity_map.py:84: AssertionError
```

i.e. _in the reference data_ there are some xticks that follow the visible ones and range from 0. to 1. Whereas in this branch those ticks are empty. The actual plots look the same to me, and those ticks don't show up either way. Where do they come from?

ajjackson:
Right, I think the explanation is that ticklabels are collected from _all_ x axes and concatenated: from `script_tests/utils.py` we have:

```python
    # Get axis labels from whichever axis has them; collect all tick labels
    for ax in fig.axes:
        ax_label_data = get_ax_label_data(ax)

        for key in 'x_label', 'y_label':
            if ax_label_data[key]:
                label_data[key] = ax_label_data[key]

        label_data['x_ticklabels'] += ax_label_data['x_ticklabels']
```

Presumably these extra x tick labels are from the invisible axes that host the xlabel. (This helps with centring etc. when data is split in several ranges.)

Annoyingly I haven't succeeded in identifying exactly which change leads to this difference. But it seems harmless, if not an improvement. Is there any benefit from these unused x values hanging out on an invisible axis?

ajjackson:
I have found out how to get the numbers back / tests passing, but I don't like it. The key difference is the `fig.tight_layout()` which I removed from `plot_2d`. This should _not_ be changing the label data!

I think the new behaviour (no tick labels on the invisible axis) is better as it reduces the risk that the user will somehow enable them while playing with styling.

ajjackson:
Ok, I now have a better understanding and a new strategy. The key point is that "layout tightening" has been moved from _before_ the data is intercepted by the script to test to a final rendering stage that comes _after_ the test would intercept the in-progress Figure object.

Because `tight_layout()` needs to check that things fit, it fills in numbers to axes ticks that have not yet been _explicitly_ set - but are implicitly defined by some ticker. So when we `get_current_plot_image_data()` before a `tight_layout()`, there are gaps in the tick labels that would be filled by one (and, critically, would also be filled by the normal plot finalisation process.)

To address this I have added a call to `tight_layout()` inside the script test. I think this is reasonable if we consider it to mean "set the numbers now so we can check them".

ajjackson:
Have created a docs page, for preview see https://euphonic.readthedocs.io/en/styling/

I notice this is the first set of preview plots included in the docs. That might be a bit misleading, because they are a bit goofy :thinking:  Is there a reason we avoided them up until now?

rebeccafair (changes requested):
This looks very nice and is a definite improvement in plotting! I like the approach and it has resulted in a big benefit for relatively little coding.  The increased use of Matplotlib arguments has made me think we probably should rename `x_label` and `y_label` as suggested in #180. 

There are just a few more things:

- Consistency with `plot_1d`, `plot_2d`. We've got a bit of a random set of style arguments as **kwargs for these functions, I think either they should support the same arguments as the command-line tools, or we should remove styling arguments from plot_1d/2d altogether (probably the latter). If we recommend using stylesheets for plotting rather than kwargs we should also provide a `with matplotlib.style.context(style)` example in `plotting.rst`
- Testing. We should really test these new arguments somehow, but I'm not sure of the best way to go about it. Maybe mock `matplotlib.style.context` and check the expected arguments are passed?
- This could be for another PR, but maybe it would be nice if we had the option to output the Spectrum1D/2D to a .json file (`-s *.json`?) and have another tool (`euphonic-plot`?) for just plotting.

euphonic/cli/utils.py
```diff
@@ -748,3 +771,41 @@ def __call__(self, parser, args, values, option_string=None):
...
+    """Combine user-specified style options with default stylesheets"""
+    if user_args.no_base_style or base is None:
+        style = []
+    else:
+        style = base
+
+    if user_args.style:
+        style += user_args.style
+
+    # Explicit args take priority over any other
```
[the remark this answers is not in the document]
ajjackson: There's a reason I mentioned it in the user docs! But yeah could also mention here.

euphonic/plot.py
```diff
@@ -186,6 +187,7 @@ def plot_1d(spectra: Union[Spectrum1D,
 
     # Add an invisible large axis for common labels
     ax = fig.add_subplot(111, frameon=False)
+    ax.grid(False)
```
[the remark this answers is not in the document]
ajjackson: Something else. If you have grid enabled in your general style, then the invisible axis creates its own grid lines which are visible and unrelated to your data. This line makes sure the invisible axes will never create a grid, without interfering with any other styling.

doc/source/styling.rst
```diff
@@ -0,0 +1,97 @@
+.. _styling:
```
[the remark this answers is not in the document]
ajjackson: Yes

ajjackson:
>     * This could be for another PR, but maybe it would be nice if we had the option to output the Spectrum1D/2D to a .json file (`-s *.json`?) and have another tool (`euphonic-plot`?) for just plotting.

Thanks to the way we've written this, that tool shouldn't need many lines of code :smile: 

And it would be useful for tweaking the appearance of expensive powder maps. (Or fancy yet-to-be-implemented neutron scattering simulations...)

Separate Issue I think, but it seems handy for not much extra code. Would we detect a `.json` in the `--save-as` argument, or provide a separate `--to-json`? Latter is nicer from implemantation standpoint and allows you to do both at the same time.

ajjackson:
The

```rst
:ref:`styling`
```

links are failing and I don't know why :disappointed:

ajjackson:
I think I got it, the `.. highlight:: bash` block was separating the tag from the actual title.

ajjackson:
> * Consistency with `plot_1d`, `plot_2d`. We've got a bit of a random set of style arguments as **kwargs for these functions, I think either they should support the same arguments as the command-line tools, or we should remove styling arguments from plot_1d/2d altogether (probably the latter). If we recommend using stylesheets for plotting rather than kwargs we should also provide a `with matplotlib.style.context(style)` example in `plotting.rst`

I think `plot_1d` has this pretty good: apart from the data, the options to `plot_1d` are titles, labels and axis ranges, and then a generic set of `**line_kwargs`. None of this is stylesheet content except for the `**line_kwargs` which are optional and passed directly to matplotlib.

`plot_2d` takes `cmap` and does _not_ take a generic set of arguments to forward to matplotlib. NonUniformImage takes a bunch of kwargs including `cmap` so we could add an `**image_kwargs` block that goes here and is backwards-compatible with the existing cmap option (which would no longer be needed). (And do the same for `plot_2d_to_axis`.)

I don't think it's a bad thing to use styling arguments and sheets together: the best practice is to use stylesheets for a "normal" theme and arguments for one-off exceptions. For example, if making something for a presentation that uses a consistent sequence of dashed lines:

```python
with context('dark_background'):
    fig, ax = plt.subplots()
    for spectrum, linestyle, label in my_results:
        plot_1d_to_axis(spectrum, ax, linestyle=linestyle)
```

One big benefit of using styles is that it _gets our settings out of the way_ of explicit arguments; if users pass an explicit styling argument we know that it will always take priority. This is better than users having to learn which arguments are ignored, renamed or otherwise treated specially.

Certainly we could nudge people towards using these tools well by providing a good example in plotting.rst. But I think there are many good use cases for kwargs and wouldn't remove or hide them.

doc/source/styling.rst
```diff
@@ -0,0 +1,108 @@
...
+and may be accessed by name, e.g.::
+
+  euphonic-dos --style seaborn quartz.castep_bin
+
+will yield a plot on a grey background with white gridlines.
+
+.. image:: figures/plot-styling-seaborn.png
+  :width: 400
+  :alt: PDOS plot with thin dark blue, red and green lines against a
+        pale blue background divided by white gridlines. There are no
```
mducle, line 44: should `blue` be `grey`?

mducle (approved):
Looks good to me!

doc/source/styling.rst
ajjackson: Yup! Initially I used a different example.

ajjackson:
Ok, there is now a fairly thorough test on the styling arguments logic. For an integration test I'm checking a few non-standard values make it from the user arguments. I haven't yet figured out a way to make sure that some of the base stylesheet options are making it though; inspecting text is tricky.

ajjackson:
Ok I managed to test the tick label size, which in the test case depends on both the style sheet and user input.

I think this is ready to go?

euphonic/cli/utils.py
```diff
@@ -748,3 +771,53 @@ def __call__(self, parser, args, values, option_string=None):
...
+                                   'font': 'font.sans-serif',
+                                   'linewidth': 'lines.linewidth',
+                                   'figsize': 'figure.figsize'}.items():
+        if getattr(user_args, user_arg, None):
+            explicit_args.update({mpl_property: getattr(user_args, user_arg)})
+
+    if 'font.sans-serif' in explicit_args:
+        explicit_args.update({'font.family': 'sans-serif'})
+
+    if 'figure.figsize' in explicit_args:
```
rebeccafair, line 816: Trailing whitespace

tests_and_analysis/test/script_tests/test_styling.py
```diff
@@ -0,0 +1,78 @@
+from argparse import Namespace
+
+import pytest
+import matplotlib.pyplot
+from numpy.testing import assert_allclose
+
+import euphonic.cli.dos
+from euphonic.cli.utils import _compose_style
+
+from tests_and_analysis.test.utils import get_data_path, get_castep_path
```
rebeccafair, line 10: Unused import

tests_and_analysis/test/script_tests/test_styling.py
```diff
@@ -0,0 +1,78 @@
...
+                  '--style=dark_background',
+                  '--linewidth=4.',
+                  '--fontsize=11.',
+                  '--figsize', '4', '4',
+                  '--figsize-unit', 'inch']
+
+        euphonic.cli.dos.main(params=params)
+
+        fig = matplotlib.pyplot.gcf()
+        #plt.tight_layout()  # Text to be drawn so we can check it
```
rebeccafair, line 72: Should this line still be here?

tests_and_analysis/test/script_tests/test_styling.py
```diff
@@ -0,0 +1,78 @@
...
+        matplotlib.pyplot.close('all')
+
+    def test_dos_styling(self, inject_mocks):
+        nah_phonon_file = get_castep_path('NaH', 'NaH.phonon')
+
+        params = [nah_phonon_file,
+                  '--style=dark_background',
+                  '--linewidth=4.',
+                  '--fontsize=11.',
+                  '--figsize', '4', '4',
```
rebeccafair, line 66: Maybe it would be better to use different values rather than 4. for linewidth and both figsize arguments? Its unlikely, but it could obscure other issues.

tests_and_analysis/test/script_tests/test_styling.py
```diff
@@ -0,0 +1,78 @@
...
+        euphonic.cli.dos.main(params=params)
+
+        fig = matplotlib.pyplot.gcf()
+        #plt.tight_layout()  # Text to be drawn so we can check it
+
+        assert_allclose([0., 0., 0., 1.], fig.get_facecolor())
+        assert fig.axes[0].lines[0].get_linewidth() == pytest.approx(4.)
+        assert_allclose([4., 4.], fig.get_size_inches())
+        # Font size: base size 11 * "small" factor 0.833
+        assert fig.axes[0].get_xticklabels()[0].get_fontsize() == pytest.approx(0.833 * 11)
```
rebeccafair, line 78: Codacy says this line is too long. I'm not so strict on the test files and it may be harder to read if split onto multiple lines, up to you whether you want to split it or leave it.

rebeccafair (approved):
Looks good to me! Just a few more small updates:

- It would be nice to have an example of `with mp.style.context` with `plot1d/2d` somewhere? It's not Euphonic-specific, but would probably be useful for people (I didn't know about it myself). Maybe in `plotting.rst` and mention styles can also be used in CL tools and link to `styling.rst`? 
- Small code changes (see comments)
- Update the changelog

Codacy currently has all pylint issues enabled, I've started to disable the irrelevant ones but they don't seem to be filtering back to the PR once it's already been analysed. I guess just ignore them, once we've had a few PRs and decided what issues we want to ignore maybe we can put them in a config file.

And since we've now actually got some images in the docs, it would be nice to have some images for the standard command-line tool plots. I'll make a separate issue/PR for that.

tests_and_analysis/test/script_tests/test_styling.py
ajjackson: Not too bad to split it so will do that

ajjackson:
Ok, have cleaned up the PEP8 stuff. I think the remaining Codacy complaint is just the way Pytest mock fixtures work?

rebeccafair:
> Ok, have cleaned up the PEP8 stuff. I think the remaining Codacy complaint is just the way Pytest mock fixtures work?

Yep! I'm not sure what to do about that message, as unused arguments are pretty important to know about, it just doesn't quite work with Pytest fixtures... It's not pleasing to have that x on the PR but I don't know what we could do other than increase the number of allowed issues.

Just a changelog/small doc update now and I think we're good.

ajjackson:
> Just a changelog/small doc update now and I think we're good.

Done!

doc/source/plotting.rst
```diff
@@ -116,6 +116,34 @@ An example of plotting 2 DOS on different axes on the same figure:
   plot_1d_to_axis(dos2, axes[1])
   fig.show()
 
+Styling
```
rebeccafair, line 119: Sorry to nitpick, but could this heading be on a higher level like
```
Styling
=======
```
Currently its showing under the 1D Plots heading when really it applies to any kind of plot.

ajjackson: Where in the heading order would you like it?

rebeccafair: After the final docstrings (2D) heading?

ajjackson: Done! A bit annoying that it doesn't indent the Contents section appropriately.

Issue #181, opened 2021-07-28
Investigate moving testing to Github actions


rebeccafair:
Running our tests using Anvil requires a reasonable amount of effort - setting up branch/PR testing, maintaining our own MacOS and Windows nodes (unreliable), and test results are inaccessible to anyone who hasn't explicitly been given access. We should investigate using Github actions and determine if there are any significant downsides.

PR #189, opened 2021-08-26
Move testing to Github Actions

rebeccafair:
Addresses #181 

Ok, so I think the Github actions testing now has the same functionality as the Jenkins testing (with hopefully more reliable test machines). The main changes are:

- I changed many of the frequency/eigenvector tests to be run without the 'reciprocal' acoustic sum rule, otherwise I was having to increase the tolerance too much to get them to pass on the MacOS runners. I think this makes our testing better anyway - see 7d8de43 commit message
- I found that some of the script tests would mysteriously segfault on the MacOS runners when scipy > 1.1.0 is used - strangely I cannot reproduce this on PACE's Mac. I decided to just skip those few tests for now, at least we have a record and can investigate more if it affects more users. Strangely this segfault can be avoided sometimes when using _more_ q-points - I was finding it was segfaulting calculating the eigenvectors for a 20^3 grid for CaHgO2, but not for a 21^3 grid for example. See a94e159 commit message.
- I'm now trying out Codacy for static analysis (maybe it will even analyse this PR...). This needs to be set up on Codacy itself rather than a Github action, and may take a bit of tweaking yet to get the settings right. I decided to use Codacy rather than a Pylint action as I couldn't find a nice way of displaying the results otherwise. I don't know how the permissions work either, can anyone else view this page? https://app.codacy.com/gh/pace-neutrons/Euphonic/dashboard
- If we're using Codacy, we can also upload code coverage results and view them in the same place. Alternatively we could use the codecov Github action, but lets see how Codacy goes...
- Added Conda testing to `release_tox.ini` and a release testing workflow. I've set it to be triggered manually rather than by a Github release for now in case the pypi/conda uploads haven't been done yet. I've tested it as much as I can but the manual triggers don't work until they're in master so we'll have to see... (and I've changed the name of some files like `jenkins_requirements` -> `ci_requirements` so it won't work until we do a new release anyway)


.github/workflows/run_tests.yml
```diff
@@ -0,0 +1,52 @@
...
+    strategy:
+      matrix:
+        os: [ubuntu-latest, windows-latest, macos-latest]
+      fail-fast: false
+    runs-on: ${{ matrix.os }}
+    steps:
+      - uses: actions/checkout@v2
+      - uses: conda-incubator/setup-miniconda@v2
+        with:
+          python-version: 3.6
```
mducle, line 18: Do you mean to just run this on python-3.6? or also 3.7, 3.8 like on Jenkins?

mducle (approved):
Looks good to me!

.github/workflows/run_tests.yml
```diff
@@ -0,0 +1,52 @@
+name: run-tests
+on:
+  push:
+  pull_request:
```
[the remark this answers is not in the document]
rebeccafair: Good idea, I've gone with the second option so it'll run when the PR is initially created, and on all branches. One issue I thought of is that we'll probably want to rerun the PR tests (to test the merge also passes) after any review changes have been made just before merging. The easiest way I could think of for doing this is with labels - adding a 'run CI' label will now retrigger the PR merge tests without having to close and reopen the PR.

.github/workflows/run_tests.yml
rebeccafair: These are done via Tox environments, 3.6 is just to run the `python -m tox` command which will create the 3.6, 3.7 and 3.8 environments.

ajjackson:
The Mac segfault thing is a bit scary, it looks like tensorflow got around it by writing some C code to specify stack sizes https://github.com/tensorflow/tensorflow/commit/000b711096ea205f56104e907191d1c9716237dc

Should I be worried about how this will affect Mac Mantid users? There will be lots of other dependencies determining exactly which Scipy version will be shipped so I don't think it could be pinned to an old one.

rebeccafair:
> The Mac segfault thing is a bit scary, it looks like tensorflow got around it by writing some C code to specify stack sizes [tensorflow/tensorflow@000b711](https://github.com/tensorflow/tensorflow/commit/000b711096ea205f56104e907191d1c9716237dc)
> 
> Should I be worried about how this will affect Mac Mantid users? There will be lots of other dependencies determining exactly which Scipy version will be shipped so I don't think it could be pinned to an old one.

Yes, it is a bit scary, especially since I don't know exactly when it will happen... and I'm finding a bit odd that this seems to have started all the way back in 1.2 and is _still_ a problem. I've tried playing with `OMP_STACKSIZE` and `ulimit` with no effect, I'm not sure what code changes we'd need as we're not doing any thread handling manually. Maybe we should enlist Dominik's help on this and see if he can reproduce it and has any ideas?

In any case this hasn't been caused by this PR, just discovered by it, I've created a new issue to look into this #191

ajjackson (approved):
I've had a poke through the new test logs and they seem ok!

No big surprises in the test changes, seems wise to identify which tests are numerically wobbly and isolate them.

There are a few deprecation warnings in the test logs, I guess we should clean that up some time.

Issue #135, opened 2021-04-06
Optimise adaptive broadening


rebeccafair:
As implemented in #131, use of `scipy.stats.norm.pdf` to calculate a Gaussian for each q-point and phonon mode in `calculate_dos` is very slow, we should investigate strategies to speed this up.

ajjackson:
Update: an STFC Graduate rotation project has been assigned to this problem. In September we will be welcoming a mathematician to the team, who will look at the proposed strategies for this and do some implementation work as well as formal analysis.

PR #197, opened 2021-11-26
Optimise adaptive broadening

jess-farmer:
As mentioned in #135, the current method for adaptive broadening (calculating a Gaussian for each q-point and phonon mode) is very slow. This PR implements a fast, approximate method based on the 'interpolate' adaptive broadening method implemented in Abins (https://docs.mantidproject.org/nightly/concepts/AbinsInterpolatedBroadening.html). 

Taking the range of mode_widths values for the spectrum, a Gaussian function need only be evaluated for regularly spaced mode_widths values across this range. The kernels at intermediate mode_widths values can then be approximated using a weighted linear combination of the two closest kernels that have been calculated exactly - the error associated with this is dependent on the choice of spacing value. Using a larger spacing value means fewer kernel evaluations but increased error, so users can set an acceptable percentage error when calling `calculate_dos` which then determines how far the mode_width samples will be spaced.

To Do:
- [x] Add tests
- [x] Optimise implementation of the method
- [x] Add fast, approximate option to cli tools
- [x] Add documentation


ajjackson:
I don't understand why Codacy is complaining about division without `__future__`, this isn't Python 2! I think that can be ignored.

euphonic/qpoint_frequencies.py
```diff
@@ -107,6 +110,12 @@ def calculate_dos(self, dos_bins: Quantity,
             Scalar float Quantity in energy units. Sets a lower limit on
             the mode widths, as mode widths of zero will result in
             infinitely sharp peaks
+        adaptive_method
+            String. Specifies whether to use slow, reference adaptive method or
+            faster, approximate method. Allowed options are 'default', or 'fast'.
+        error
+            Scalar float. Acceptable percentage error for gaussian approximations when using
+            the fast adaptive method.
```
ajjackson: Is this actually "percentage" (i.e. 0.001% = 0.00001)? I expect we can give a stricter definition (e.g. RMS error/total intensity).

euphonic/qpoint_frequencies.py
```diff
@@ -107,6 +110,12 @@ def calculate_dos(self, dos_bins: Quantity,
             Scalar float Quantity in energy units. Sets a lower limit on
             the mode widths, as mode widths of zero will result in
             infinitely sharp peaks
+        adaptive_method
```
ajjackson, line 113: I wonder if we could name the existing method 'reference'; it would be nice to make the new method the default, once it is robust enough!

euphonic/qpoint_frequencies.py
```diff
@@ -89,7 +90,9 @@ def __setattr__(self, name: str, value: Any) -> None:
 
     def calculate_dos(self, dos_bins: Quantity,
                       mode_widths: Optional[Quantity] = None,
-                      mode_widths_min: Quantity = Quantity(0.01, 'meV')
+                      mode_widths_min: Quantity = Quantity(0.01, 'meV'),
+                      adaptive_method: Optional[str] = 'default',
+                      error: Optional[float] = 0.001,
```
ajjackson: Perhaps this should be 'adaptive_error' to link it more obviously to adaptive broadening.

euphonic/qpoint_frequencies.py
```diff
@@ -140,6 +152,13 @@ def _calculate_dos(self, dos_bins: Quantity,
         case, mode_widths and mode_weights must be shape (1, n_modes)
         arrays
         """
+        if adaptive_method!='fast' and error is not None:
+            warnings.warn('\n The error argument can only be applied when using the fast adaptive '
+                          'method, so will be ignored.')
+        if mode_widths is None and adaptive_method!='default' or mode_widths is None and error is not None:
```
ajjackson: This seems a bit cumbersome especially as it will _not_ warn somebody against a call to `_calculate_dos(mode_widths=None, adaptive_method='default')` (i.e. they have still non-sensically set an option for adaptive_method, but it happens to equal the default.).

I suggest either
- we don't warn in this case, or
- we let the default value = `None`, warn `if mode_widths is None and adaptive_method is not None` and then set it to a default if None. (The default should be mentioned in docstring.)

euphonic/qpoint_frequencies.py
```diff
@@ -168,6 +188,17 @@ def _calculate_dos(self, dos_bins: Quantity,
...
                     dos += pdf*weights[q]*mode_weights_calc[q, m]
+        elif mode_widths is not None and adaptive_method=='fast':
+            # fast, approximate method for adaptive broadening
+            freqs = np.ravel(freqs)
+            mode_widths = np.ravel(mode_widths).to('hartree').magnitude
+            n_weights = np.repeat(weights, n_modes)
+
+            # set default error value if none specified
+            if error is None:
+                error = 0.01
```
ajjackson, line 202: I don't see the harm in this having a default value in the function signature.

Admittedly we are accumulating arguments that "don't always do anything". `mode_widths_min` was already one of those. But short of breaking some argument groups into dicts (which impacts usability) I don't see a nice way of tidying it up 🤔

euphonic/qpoint_frequencies.py
```diff
@@ -168,6 +188,17 @@ def _calculate_dos(self, dos_bins: Quantity,
                     pdf = norm.pdf(dos_bins_calc, loc=freqs[q,m],
                                    scale=mode_widths[q,m])
                     dos += pdf*weights[q]*mode_weights_calc[q, m]
+        elif mode_widths is not None and adaptive_method=='fast':
+            # fast, approximate method for adaptive broadening
+            freqs = np.ravel(freqs)
+            mode_widths = np.ravel(mode_widths).to('hartree').magnitude
```
ajjackson, line 197: It looks a bit strange (and scary) that this one method requires the input data to be cast to particular units. Would it hurt performance to put that inside `fast_broaden()`?

ajjackson (commented):
Obviously this is still in progress, but I noted some quick thoughts on API things.

euphonic/qpoint_frequencies.py
ajjackson: I was initially going to suggest 'exact', but that might mislead people away from the fact that adaptive broadening is itself an approximation to speed up q-point convergence.

euphonic/qpoint_frequencies.py
jess-farmer: Yeah, I can see that it is a bit unclear what percentage refers to - I will add in a stricter definition

euphonic/qpoint_frequencies.py
jess-farmer: Good point!

euphonic/qpoint_frequencies.py
jess-farmer: Yeah that's true, I'll change the default value to None and add this to the docstring.

euphonic/qpoint_frequencies.py
jess-farmer: The reason I'd set the default value to None in the function signature was to avoid the case (as noted with the adaptive_method warning) where the user isn't warned if they happen to non-sensically set the error to the default option - I'll have a think about whether this approach can be modified/improved

euphonic/qpoint_frequencies.py
jess-farmer: Yeah fair point, I'll move these lines inside of the function

euphonic/qpoint_frequencies.py
ajjackson: If it is also named `adaptive_error` the problem is reduced a bit. (Or at least we can lay more blame on the user.)

If we are going to attach warnings to over half the arguments on this function, it should be done in a systematic way. But it has "bad smell" either way - is there a better way to avoid this problem?

rebeccafair:
> I don't understand why Codacy is complaining about division without `__future__`, this isn't Python 2! I think that can be ignored.

Me neither! I have disabled that issue in Codacy now so hopefully it shouldn't come up again.

euphonic/fast_adaptive_broadening.py
```diff
@@ -0,0 +1,166 @@
...
+
+    n_kernels = int(
+        np.ceil(np.log(max(mode_widths)/min(mode_widths))/np.log(spacing)))
+    mode_width_samples = spacing**np.arange(n_kernels+1)*min(mode_widths)
+    freq_range = 3*max(mode_widths)
+    kernel_npts_oneside = np.ceil(freq_range/bin_width)
+    kernels = gaussian(xvals=np.arange(-kernel_npts_oneside,
+                       kernel_npts_oneside+1, 1)*bin_width,
+                       sigma=mode_width_samples[:, np.newaxis],
+                       centre=0)*bin_width
```
ajjackson, line 69: Looking at all these "kernel_npts_oneside" reminded me that we do have the option to evaluate the kernels on one side and mirror them to make the full symmetric function. A quick play around in IPython suggests that
```
scipy.stats.norm.pdf(x, scale=sigma)
```
is _slightly_ faster than
```
half_pdf  scipy.stats.norm.pdf(half_x, scale=sigma)
pdf = np.concatenate([np.flip(half_pdf[:, 1:], axis=1), half_pdf], axis=1)
```
where x is 1001 points, half_x is 501 points and sigma is a column vector of 20 widths. But if we bump the numbers a bit to 4001 points and 50 sigma values, the mirror implementation is nearly twice as fast. 

I think what is happening is that `norm.pdf` has better-than-linear scaling for small numbers of points, as the meaningful work grows larger relative to overheads. Once we hit the actual linear-scaling regime it becomes possible for the factor-2 saving to outweight the overheads.

This transition point is uncomfortably close to the kind of numbers we are looking at using in practice. I guess for now we can stick to a simple implementation and then if kernel generation is a real-world bottleneck _then_ consider introducing this mirror technique.

euphonic/fast_adaptive_broadening.py
```diff
@@ -0,0 +1,166 @@
...
+    kernels_idx = np.searchsorted(mode_width_samples, mode_widths)
+
+    lower_coeffs = find_coeffs(spacing)
+    scaled_data_matrix = np.zeros((len(dos_bins)-1, len(kernels)))
+    for i in range(1, len(mode_width_samples)):
```
ajjackson, line 74: Might be worth a comment to explain that we intentionally skip points with insert-position 0 because they lie outside the bin range.

ajjackson (commented):
Some more comments! This is looking very promising already; masking seems to be a powerful way to keep this sort of thing clean.

euphonic/fast_adaptive_broadening.py
```diff
@@ -0,0 +1,162 @@
...
+        return (weight * gaussian(xvals, sigma=1)
+                + (1-weight) * gaussian(xvals, sigma=spacing))
+
+    lower_mix = np.zeros(len(sigma_values))
+
+    for i, s_val in enumerate(sigma_values):
+        actual_gaussian = gaussian(x_range, s_val)
+        mixl, _ = curve_fit(gaussian_mix, x_range,
+                            ydata=actual_gaussian, p0=[0.5], bounds=(0, 1))
+        lower_mix[i] = mixl[0]
```
ajjackson, line 159: It looks like if we need it there is a speedup to had by switching from nonlinear to linear-least-squares fitting. (scipy.linalg.lstsq)

```

In [128]: %timeit lstsq(np.array([norm.pdf(x_range, scale=1), norm.pdf(x_range, scale=1.5)]).T, actual_
     ...: gaussian)
242 µs ± 1.88 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)

In [129]: %timeit mixl, _ = curve_fit(gaussian_mix, x_range, ydata=actual_gaussian, p0=[0.5], bounds=(0
     ...: , 1))
2.96 ms ± 22.1 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

The catch is that we don't get to enforce bounds any more, so might have to watch out for strange negative solutions.

ajjackson (commented):
A couple more comments; I'll come back and look at the tests after the next round of revisions?

jess-farmer:
> A couple more comments; I'll come back and look at the tests after the next round of revisions?

Many thanks for your comments - I'll have a more thorough look through and address these individually over the next few days once I am happy with the unit tests I'm putting in place. I'll then also start profiling and looking for further ways to speed up the code.

euphonic/fast_adaptive_broadening.py
```diff
@@ -0,0 +1,166 @@
...
+    Returns
+    -------
+    dos
+        Float ndarray of shape (dos_bins - 1,) containing density of states
+        ydata
+    """
+    n_modes = freqs.shape[-1]
+    freqs = np.ravel(freqs)
+    mode_weights = np.ravel(mode_weights)
+    mode_widths = np.ravel(mode_widths).to('hartree').magnitude
```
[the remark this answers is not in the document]
jess-farmer: I have moved the hartree conversion back outside of the `fast_broaden` function and renamed the function variables to make it clear that the mode widths and bin edges are expected in hartree units

euphonic/fast_adaptive_broadening.py
```diff
@@ -0,0 +1,166 @@
...
+
+    lower_coeffs = find_coeffs(spacing)
+    scaled_data_matrix = np.zeros((len(dos_bins)-1, len(kernels)))
+    for i in range(1, len(mode_width_samples)):
+        masked_block = (kernels_idx == i)
+        sigma_factors = mode_widths[masked_block]/mode_width_samples[i-1]
+        lower_mix = np.polyval(lower_coeffs, sigma_factors)
+        upper_mix = 1-lower_mix
+
+        n_spec = np.zeros(len(mode_widths))
```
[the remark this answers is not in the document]
jess-farmer: Yeah I'm not sure what I was thinking regarding variable name - think it was something along the lines of the spectrum associated with the nth kernel but it doesn't make much sense really! Have now tidied up this bit of the code

euphonic/fast_adaptive_broadening.py
```diff
@@ -0,0 +1,162 @@
...
+        the gaussian kernel is exactly calculated.
+
+    Returns
+    -------
+    coeffs
+        Array containing the polynomial coefficients, with the highest
+        power first
+    """
+    sigma_values = np.linspace(1, spacing, 10)
+    x_range = np.linspace(-10, 50, 101)
```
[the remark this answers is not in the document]
jess-farmer: It shouldn't be - have now fixed this

euphonic/qpoint_frequencies.py
rebeccafair: I agree `adaptive_error` should be set in the function signature. I don't think we can/should give warnings for all sets of nonsensical arguments, as it could make the code more bloated and confusing if anything. For example, there's no warning elsewhere in Euphonic if someone sets `use_c=False` but `n_threads=4`. If we make the connections between arguments obvious enough in the docstring/variable names it should be ok, we should give the user _some_ credit. 

To give one other example, I've just had a look at [`numpy.histogram`](https://numpy.org/doc/stable/reference/generated/numpy.histogram.html) and it has a similar issue with having to support many potential options. If you give it an array of bin edges as the `bins` argument, the `range` argument is then ignored. Although you might suspect this it isn't stated in the docstring though.

Anyway, back to looking at this PR...

tests_and_analysis/test/euphonic_test/test_qpoint_frequencies.py
```diff
@@ -435,19 +441,53 @@ def test_calculate_dos(
...
+             np.arange(0, 100, 0.1)*ureg('meV'))])
+    def test_calculate_dos_with_mode_widths_fast_method(
+            self, material, qpt_freqs_json, mode_widths_json,
+            expected_dos_json, ebins):
+        qpt_freqs = get_qpt_freqs(material, qpt_freqs_json)
+        mode_widths = get_mode_widths(mode_widths_json)
+        dos = qpt_freqs.calculate_dos(
+            ebins, mode_widths=mode_widths, adaptive_method='fast')
+        expected_dos = get_expected_spectrum1d(expected_dos_json)
+        check_spectrum1d(dos, expected_dos, 1e-13)
```
rebeccafair, line 473: To avoid duplication, these parameters can probably be used in the existing `test_calculate_dos_with_mode_widths`. We just need another parameter to set the other kwargs to `calculate_dos`. Does that make sense?

tests_and_analysis/test/euphonic_test/test_qpoint_phonon_modes.py
```diff
@@ -96,6 +96,12 @@ def to_constructor_args(self, crystal=None, qpts=None, frequencies=None,
 def get_qpt_ph_modes_dir(material):
     return os.path.join(get_data_path(), 'qpoint_phonon_modes', material)
 
+def get_mode_widths(mode_widths_json):
```
rebeccafair, line 99: Put `get_mode_widths` in `utils.py` as above

tests_and_analysis/test/euphonic_test/test_qpoint_phonon_modes.py
```diff
@@ -544,19 +550,49 @@ def test_calculate_dos_from_phonopy(
...
+             np.arange(0, 155, 0.1)*ureg('meV'))])
+    def test_calculate_dos_with_mode_widths_fast_method(
+            self, material, qpt_ph_modes_json, mode_widths_json,
+            expected_dos_json, ebins):
+        qpt_ph_modes = get_qpt_ph_modes_from_json(material, qpt_ph_modes_json)
+        mode_widths = get_mode_widths(mode_widths_json)
+        dos = qpt_ph_modes.calculate_dos(
+            ebins, mode_widths=mode_widths, adaptive_method='fast')
+        expected_dos = get_expected_spectrum1d(expected_dos_json)
+        check_spectrum1d(dos, expected_dos, 1e-13)
```
rebeccafair, line 578: Unless I'm missing something this can also be tested in the same test as `test_calculate_dos_with_mode_widths_ref_method` (formerly `test_calculate_dos_with_mode_widths`) above if another parameter is added.

rebeccafair (commented):
I'd like to first say that anecdotally this really does make the adaptive broadening with large numbers of atoms much faster, yay!

Is there going to be a `pace-developers` doc describing the method (maybe [here](https://github.com/pace-neutrons/pace-developers/tree/master/euphonic/design))? @ajjackson? Some things in particular I'd like to see:
* An overview of the method. I think I basically understand it now after looking through the code for a while but it would be good to have a summary.
* Show the data/investigation that resulted in the 'magic' numbers in `spacing = np.polyval([ 612.7, -122.7, 15.40, 1.0831], adaptive_error)`
* How were the mode width samples decided? I see its some sort of power relation, and I can understand why you'd need different spacings to sensibly cover all the orders of magnitude, but not the Maths specifically

euphonic/fast_adaptive_broadening.py
```diff
@@ -0,0 +1,158 @@
...
+    freqs = np.ravel(freqs)
+    mode_widths = np.ravel(mode_widths_hartree)
+    combined_weights = np.ravel(mode_weights * weights[:, np.newaxis])
+
+    # determine spacing value for mode_width samples given desired error level
+    # coefficients determined from a polynomial fit to plot of
+    # error vs spacing value
+    spacing = np.polyval([ 612.7, -122.7, 15.40, 1.0831], adaptive_error)
+
+    bin_width = dos_bins_hartree[1]-dos_bins_hartree[0]
```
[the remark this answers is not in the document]
ajjackson: Discrete convolution-based methods will absolutely not work with uneven bin widths - I guess there should be a sanity check on that. It's quite possible there is also other stuff in Euphonic that will break on irregular grids. We should maybe think more generally about defining precisely what is/can be on an irregular grid.

euphonic/fast_adaptive_broadening.py
```diff
@@ -0,0 +1,158 @@
...
+
+        freqs_prev = freqs[masked_block]
+        upper_weights_prev = combined_weights[masked_block] - lower_weights
+
+        if i == len(mode_width_samples)-1:
+            hist, _ = np.histogram(freqs[masked_block], bins=dos_bins_hartree,
+                                weights=upper_weights_prev/bin_width)
+            scaled_data_matrix[:, i] = hist
+
+    dos = np.sum([convolve(scaled_data_matrix[:, i], kernels[i],
```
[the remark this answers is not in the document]
jess-farmer: Yeah this is a good point, I have now changed this so that the convolution happens inside the main loop over kernels

euphonic/fast_adaptive_broadening.py
```diff
@@ -0,0 +1,158 @@
...
+    bin_width = dos_bins_hartree[1]-dos_bins_hartree[0]
+
+    n_kernels = int(
+        np.ceil(np.log(max(mode_widths)/min(mode_widths))/np.log(spacing)))
+    mode_width_samples = spacing**np.arange(n_kernels+1)*min(mode_widths)
+    # Determine frequency range for gaussian kernel, the choice of
+    # 3*max(sigma) is arbitrary but tuned for acceptable error/peformance
+    freq_range = 3*max(mode_widths)
+    kernel_npts_oneside = np.ceil(freq_range/bin_width)
+    kernels = gaussian(np.arange(-kernel_npts_oneside,
```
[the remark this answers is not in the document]
jess-farmer: I had been using scipy for calculating gaussians but found that the custom function was slightly faster (see this [notebook](https://github.com/pace-neutrons/Broadening-Methods/blob/main/gaussian_evaluation.ipynb)) - to be fair both options are fast so can always switch to using scipy again if that would be preferred.

euphonic/fast_adaptive_broadening.py
```diff
@@ -0,0 +1,158 @@
...
+Functions for fast adaptive broadening of density of states spectra
+"""
+from typing import Optional
+from scipy.optimize import nnls
+from scipy.signal import convolve
+import numpy as np
+
+def fast_broaden(dos_bins_hartree: np.ndarray,
+                 freqs: np.ndarray,
+                 mode_widths_hartree: np.ndarray,
```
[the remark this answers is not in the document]
ajjackson: Makes sense. But we _usually_ have energy-like things in Euphonic as Quantities. I can't think of a convenient shorthand for "consistent dimensions, doesn't matter which".

I guess `dos_bins` and `mode_widths` would fine ok here in practice, the type-hinting in the signature makes it clear what is going on and this function doesn't deal with Quantities.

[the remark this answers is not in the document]
jess-farmer: Adam and I had a discussion about this earlier - could we make `fast_broaden` a private method but with a public wrapper which checks that units are consistent?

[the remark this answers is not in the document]
ajjackson: Something like
```python
def fast_broaden(bins: Quantity,
                 x: Quantity,
                 widths: Quantity,
                 weights: Quantity, *args) -> Quantity:
    return _fast_broaden(bins.to('Ha').magnitude,
                         x.to('Ha').magnitude,
                         widths.to('Ha').magnitude,
                         weights.to('1/Ha').magnitude, *args
                         ) * ureg('1/Ha')
                        
def _fast_broaden(bins: np.ndarray, x: np.ndarray, weights: np.ndarray, *args) -> np.ndarray:
    ...
```

How strict are we being in the surrounding Euphonic machinery about having appropriate units for `weights`?

euphonic/fast_adaptive_broadening.py
```diff
@@ -0,0 +1,166 @@
...
+    # set default error value if none specified
+    if adaptive_error is None:
+        adaptive_error = 0.01
+
+    # determine spacing value for mode_width samples given desired error level
+    spacing = 656.1*adaptive_error**3 - 131.8*adaptive_error**2 +\
+                15.98*adaptive_error + 1.0803
+
+    bin_width = dos_bins[1]-dos_bins[0]
+    mode_widths = np.maximum(mode_widths, bin_width/2)
```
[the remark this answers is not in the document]
jess-farmer: This line has been taken out, and the lower bound for `mode_widths` is now set by the `mode_widths_min` argument in `calculate_dos`

euphonic/fast_adaptive_broadening.py
jess-farmer: See this [notebook](https://github.com/pace-neutrons/Broadening-Methods/blob/main/Compare%20least%20squares%20solvers.ipynb) for a comparison of different least-squares solvers - the fastest choice appears to be `scipy.optimize.nnls`  which is a non-negative least squares solver and so this was used to replace `curve_fit`

euphonic/fast_adaptive_broadening.py
```diff
@@ -0,0 +1,162 @@
...
+            hist_x = np.concatenate((x_prev, x[masked_block]))
+            hist, _ = np.histogram(hist_x, bins=bins,
+                                   weights=mixing_weights/bin_width)
+
+        x_prev = x[masked_block]
+        upper_weights_prev = weights[masked_block] - lower_weights
+
+        spectrum += convolve(hist, kernels[i-1], mode="same", method="fft")
+
+        if i == len(width_samples)-1:
```
ajjackson, line 96: I think we can eliminate this logic by instead iterating i up to len(width_samples) + 1). The leftover upper_weights should get concatenated onto zero new data (as the masked_block will be empty).

euphonic/fast_adaptive_broadening.py
```diff
@@ -0,0 +1,162 @@
...
+
+    # start loop over kernels from 1 as points with insert-position 0
+    # lie outside of bin range but first include points
+    # where mode_widths = min(mode_width_samples)
+    masked_block = (widths == min(width_samples))
+    width_factors = widths[masked_block]/width_samples[0]
+    lower_weights = (np.polyval(lower_coeffs, width_factors))*weights[masked_block]
+    hist, _ = np.histogram(x[masked_block], bins=bins,
+                           weights=lower_weights/bin_width)
+    spectrum += convolve(hist, kernels[0], mode="same", method="fft")
```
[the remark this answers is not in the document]
ajjackson: I _think_ we can eliminate the whole block by changing a line above to `kernels_idx = np.searchsorted(width_samples, widths, side='right')` which will cause it to "round up" exact matches and place min(sigma) in block 1. It doesn't matter that other boundary values will also get moved; they get counted either way, and now we don't have to treat the lower bound as a special case.

But we need an appropriate test to be sure that the results are equivalent.

ajjackson (commented):
I have had a look at the looping logic and think we can eliminate the "special cases" at each end, cutting redundancy.

ajjackson:
As discussed this morning, we should think a bit more carefully now about the name/structure of this module. Currently we are adding a module named "fast_adaptive_broadening" but some of the code can be used for purposes other than "adaptive" broadening (i.e. of a DOS) - e.g. instrumental broadening.

So it may make sense in future to have a `broadening` module containing `adaptive`, `instrumental`, `4d`, `even_faster` etc. Or maybe YAGNI. My suggestion is that, similar to the approach with the Brille module (#104 ), we name this new file *broadening.py* and then if in future we want to spread out then `import broadening` could refer to *euphonic/broadening/__init__.py` and the public API doesn't have to break.

@rebeccafair  is likely to have opinions on this!

rebeccafair:
> So it may make sense in future to have a `broadening` module containing `adaptive`, `instrumental`, `4d`, `even_faster` etc. Or maybe YAGNI. My suggestion is that, similar to the approach with the Brille module (#104 ), we name this new file _broadening.py_ and then if in future we want to spread out then `import broadening` could refer to *euphonic/broadening/**init**.py` and the public API doesn't have to break.

I agree, `broadening.py` seems like the most future-proof name. I do think we should separate the name 'adaptive' (which usually refers to broadening of the different DOS contributions by amounts proportional to the mode gradients), and the actual implementation of broadening samples by different widths, which is generic and could be applied to other things such as instrumental resolution like you mentioned. I can't think of a word that accurately describes 'broadening each sample by a different width' though. We could just continue using the word adaptive and ensure the variables and method docstrings are generic.

ajjackson:
broadening.variable_width perhaps? This is a notation issue we will need to settle for reporting purposes, anyway!

rebeccafair:
> broadening.variable_width perhaps?

Ah of course, variable is a word! :sweat_smile:

rebeccafair (changes requested):
Looks pretty good to me! I just thought you can also add yourself as an author to the `CITATION.cff` file in this PR, the names are in alphabetical order.

One other thing to look at would be the test timings, the latest set of [tests](https://github.com/jess-farmer/Euphonic/runs/5106697505) on this branch took ~38 mins, whereas the [tests](https://github.com/pace-neutrons/Euphonic/runs/5070696763?check_suite_focus=true) on master took ~32 mins. It might be that's just variation, but worth checking none of the added tests are overly expensive to run. 

I still need to figure out how to get test results to show up in this PR...

tests_and_analysis/test/euphonic_test/test_broadening.py
```diff
@@ -0,0 +1,76 @@
...
+    assert adaptively_broadened_dos_area == pytest.approx(dos_area, rel=0.01)
+
+@pytest.mark.parametrize(
+        ('material, qpt_freqs_json, mode_widths_json,'
+         'expected_dos_json, ebins'), [
+            ('quartz', 'toy_quartz_qpoint_frequencies.json',
+             'toy_quartz_mode_widths.json',
+             'toy_quartz_adaptive_dos.json',
+             np.arange(0, 40, 0.1)*ureg('meV'))])
+def test_lower_bound_widths_broadened(material, qpt_freqs_json,
```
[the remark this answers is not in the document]
rebeccafair: Ah I see, good spot! I think because this set of inputs needs to be tested specifically to prevent the bug reoccurring it's fine to have it as a separate test, maybe just add a comment in the test describing the bug?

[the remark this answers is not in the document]
jess-farmer: Okay will do!

euphonic/broadening.py
```diff
@@ -0,0 +1,167 @@
+"""
+Functions for broadening spectra
+"""
+from scipy.optimize import nnls
+from scipy.signal import convolve
+import numpy as np
+
+from euphonic import ureg, Quantity
+
+def variable_width(bins: Quantity,
```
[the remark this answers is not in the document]
jess-farmer: At the moment I've changed the name of the fast method to `width_interpolated_broadening` but also thought it could be called one of the following: `log_spaced_kernel_broadening`, `width_scaled_spectra_broadening`, or `approximate_weight_scaled_broadening` - not sure how to convey the idea of the method in a more concise way or if these names would even make sense to people not already familiar with the method...

jess-farmer:
As Codecov hasn't been running on this PR, I have run a coverage report locally and as a result added a couple of extra tests.

Also, following a discussion yesterday, Gaussians are now evaluated on a grid the same length as the number of bins - this should mean that results are more accurate as the widest Gaussians are no longer being truncated and it also means that there's no zero padding to be added in the convolution step. Having timed this and the previous version of the method, the performance is also slightly better without the truncation step.

euphonic/cli/utils.py
```diff
@@ -721,6 +721,15 @@ def __call__(self, parser, args, values, option_string=None):
...
                           'using fixed width broadening'))
+                section.add_argument(
+                    '--adaptive-method', type=str, default='reference',
+                    dest='adaptive_method', choices=('reference', 'fast'),
+                    help='The adaptive broadening method')
+                section.add_argument(
+                    '--adaptive-error', type=float, default=0.01,
+                    dest='adaptive_error',
+                    help=('Acceptable absolute error '
+                          'for gaussian approximations'))
```
rebeccafair, line 732: Specify this applies to the 'fast' method

rebeccafair (changes requested):

Issue #198, opened 2021-11-30
Test Optional Dependencies


rebeccafair:
Currently all (including optional) dependencies are installed before tests are run. Because of this we don't _really_ know if these dependencies are optional as we don't know if the tests will pass without them. Tests which depend on specific optional dependencies should be marked and run with them installed, but all other tests should be run without optional dependencies.

PR #199, opened 2021-12-03
Test optional dependencies

rebeccafair:
Addresses #198. I started thinking about this because adding Brille to Euphonic will add another optional dependency and it's probably easier to get on top of this sooner rather than later. You can now run pytest with something like `pytest -v tests_and_analysis/test -m "not matplotlib"` if you don't have Matplotlib installed and want to run every test except the matplotlib-dependent ones, or `pytest -v tests_and_analysis/test -m matplotlib` to _only_ run the matplotlib-dependent ones. This sort of thing is now used in the CI, so if we accidentally use an optional dependency somewhere else we should know about it.

Summary:
* Removed `unit` and `integration` markers as we weren't using them for any practical purposes
* Added `phonopy_reader` and `matplotlib` markers to tests that require each of these extras
* Refactored tests so that optional dependencies aren't imported unless a test is run:
   * Avoid importing dependencies in fixtures (as these are evaluated whether or not the test that uses them is run). For example `ForceConstants/QpointPhononModes/QpointFrequencies.from_x` tests are now tests in their own right rather than using fixtures. I actually think this looks cleaner anyway.
   * Allow `import matplotlib` to fail at the top of test files that have `mark.matplotlib`. Even if the whole file is marked with `pytestmark = pytest.mark.matplotlib` at the top, the file still needs to be imported by pytest even if the tests are not run, which would cause pytest to fail. The other alternative would be importing matplotlib as needed inside every test, and somehow refactor the fixtures, but that seemed like more faff.
   * Split some tests (e.g. going from `QpointPhononModes.from_castep/phonopy` to `StructureFactor`) into separate tests so they can be marked appropriately. There is still some duplication with this but it's not too bad.
   * There are 921 tests in the full suite in master and on this branch so I haven't accidentally dropped any!
* Refactored `tox.ini` to run tests with different extras:
  * py37. py38 tests will run all tests will all extras as normal
  * py36 will run tests with no extras, with each extra matplotlib and phonopy_reader, then run any other tests that require both extras
  * I've done it this way as there's some overhead with creating multiple tox environments. The CI on master takes 21 mins, whereas on this branch takes 23 mins even though we're running the same number of tests. I think running the full suite on py37 and py38, and running for specific extras only on py36 probably gives us a good mix of being thorough and not bloating the CI.
* Have not refactored `release_tox.ini` to test with specific extras, instead only running the whole suite as we're already testing for py36, py37 and py38, conda and pypi wheels and pypi source so adding specific extras tests might get messy. But feel free to disagree.

@jess-farmer Jess, tagging you as an FYI. Hopefully there wont be an issue with merging this with any tests you've already written but if there are feel free to ask me! I imagine you're not using `matplotlib` or `from_phonopy` in any of your tests but if you are you will just need to add the relevant markers, again feel free to ask.

### DIFF
--- a/tests_and_analysis/test/euphonic_test/test_calculate_structure_factor.py
+++ b/tests_and_analysis/test/euphonic_test/test_calculate_structure_factor.py
@@ -20,68 +20,50 @@ from tests_and_analysis.test.euphonic_test.test_structure_factor import (
 @pytest.mark.integration
 class TestCalculateStructureFactorFromForceConstants:
 
-    def get_multithreaded_kwargs(self, n_threads):
-        kwargs = {}
+    def get_quartz_fc():
+        return ForceConstants.from_castep(
+            get_castep_path('quartz', 'quartz.castep_bin'))
+
+    def get_si2_fc():
+        return ForceConstants.from_castep(
+            get_castep_path('Si2-sc-skew', 'Si2-sc-skew.castep_bin'))
+
+
+    tol_kwargs = {'sf_atol': 3e-3,
+                  'sf_rtol': 3e-4,
+                  'freq_atol': 1e-4,
+                  'freq_rtol': 3e-4,
+                  'freq_gamma_atol': 0.55,
+                  'sf_gamma_atol': 3e2}
+
+    @pytest.mark.parametrize(
+        'material, fc, qpt_ph_modes_args, dw_file, expected_sf_file', [
+            ('quartz', get_quartz_fc(), (get_test_qpts('split'), {}),
+             'quartz_666_0K_debye_waller.json',
+             'quartz_0K_fc_structure_factor.json'),
+            ('quartz', get_quartz_fc(), (get_test_qpts('split'), {}),
+             'quartz_666_300K_debye_waller.json',
+             'quartz_300K_fc_structure_factor.json'),
+            ('quartz', get_quartz_fc(), (get_test_qpts('split'), {}),
+             None,
+             'quartz_fc_structure_factor.json'),
+            ('Si2-sc-skew', get_si2_fc(), (get_test_qpts(), {}),
+             'Si2-sc-skew_666_300K_debye_waller.json',
+             'Si2-sc-skew_300K_fc_structure_factor.json'),
+            ('quartz', get_quartz_fc(),
+             (get_test_qpts('split'), {'asr': 'reciprocal'}),
+             'quartz_666_300K_debye_waller.json',
+             'quartz_recip_asr_300K_fc_structure_factor.json')])
+    @pytest.mark.parametrize('n_threads', [0, 1, 2])
+    def test_calculate_structure_factor(self, material, fc, qpt_ph_modes_args,
+                                        dw_file, expected_sf_file, n_threads):
+        args, kwargs = qpt_ph_modes_args
         if n_threads == 0:
             kwargs['use_c'] = False
         else:
             kwargs['use_c'] = True
             kwargs['n_threads'] = n_threads
-        return kwargs
-
-    @pytest.fixture(params=[0, 1, 2])
-    def get_quartz_qpt_ph_modes_recip(self, request):
-        fc = ForceConstants.from_castep(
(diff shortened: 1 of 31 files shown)


tests_and_analysis/test/euphonic_test/test_calculate_structure_factor.py
```diff
@@ -17,109 +17,110 @@
...
-        return fc.calculate_qpoint_phonon_modes(get_test_qpts(), **kwargs)
-
-    @pytest.fixture(params=[0, 1, 2])
-    def get_cahgo2_qpt_ph_modes(self, request):
-        fc = ForceConstants.from_phonopy(
-            path=get_phonopy_path('CaHgO2', ''),
-            summary_name='mp-7041-20180417.yaml')
-        kwargs = self.get_multithreaded_kwargs(request.param)
-        return fc.calculate_qpoint_phonon_modes(get_test_qpts(), **kwargs)
+    tol_kwargs = {'sf_atol': 3e-3,
```
oerc0122, line 32: Why these tolerances? Magic numbers usually need comment.

tests_and_analysis/test/euphonic_test/test_install.py
```diff
@@ -11,7 +12,9 @@ def test_license_is_installed(self):
             license_data = fp.readlines()
         assert 'GNU GENERAL PUBLIC LICENSE' in license_data[0]
 
+    @pytest.mark.phonopy_reader  # Needs pyyaml
     def test_citation_cff_is_installed(self):
+        import yaml
```
oerc0122, line 17: Do you want a try-catch with import error here? Then you might be able to skip rather than fail?

tests_and_analysis/test/run_tests.py
```diff
@@ -1,9 +1,27 @@
...
-from typing import Tuple, List, Union
+
+
+def main():
+    test_dir, reports_dir = _get_test_and_reports_dir()
+
+    do_report_coverage, do_report_tests, tests, markers_to_run = \
+        _get_parsed_args(test_dir)
+
+    pytest_options: List[str] = _build_pytest_options(
```
oerc0122, line 17: Minor inconsistency in format, purely cosmetic.

tests_and_analysis/test/script_tests/test_powder_map.py
```diff
@@ -5,15 +5,21 @@
...
-import matplotlib.pyplot
 from packaging import version
 from scipy import __version__ as scipy_ver
 
 from tests_and_analysis.test.utils import get_data_path, get_castep_path, get_phonopy_path
 from tests_and_analysis.test.script_tests.utils import (
     get_script_test_data_path, get_current_plot_image_data, args_to_key)
-import euphonic.cli.powder_map
+
+pytestmark = pytest.mark.matplotlib
```
oerc0122, line 15: This code seems to be duplicated a lot. Would it be better to use a module var/func or class property to set this?

oerc0122 (approved):
Cosmetic suggestions, if it's all running fine, looks good to me.

tests_and_analysis/test/euphonic_test/test_calculate_structure_factor.py
rebeccafair: There is a comment, but stupidly its where these values are used rather than where they're defined. I've moved things around.

tests_and_analysis/test/euphonic_test/test_install.py
rebeccafair: Good idea, done.

tests_and_analysis/test/run_tests.py
rebeccafair: Fixed.

tests_and_analysis/test/script_tests/test_powder_map.py
rebeccafair: If imports were done elsewhere, they would need to be passed back to this file so they could be used, I'm not sure that could be done without it being a bit confusing and unconventional. Leaving as is for now.

Issue #201, opened 2022-01-18
Handle uneven bin widths in spectra


rebeccafair:
Some operations, such as broadening by convolution, are not strictly correct when applied to data with uneven bin widths (irregularly sampled data). We should ensure these are handled sensibly.

rebeccafair:
I worry that this approach could be a lot of work for a specific case, do people generally actually care about completely correct broadening along q in a bandstructure? For publication quality plots, maybe, but in that case they may want to create their own set of Spectrum2Ds over each direction, broaden them, then stick them together to make a single Spectrum2D, as I doubt the automatically generated bandstructure would be the one that they want anyway. Horace has 'spaghetti' plots, it would be interesting to know how the broadening is handled there, whether all the q-bins are the same or not. In any case I don't think this is a high priority issue at the moment.

ajjackson:
I don't think it would be _that_ much work, but agree that it is not really a priority. I'm happy to go ahead without it for now; if people are using these things then maybe it's worth refining.

The differences would mostly only be significant for _large_ q-broadening and I don't have much sense right now of what values are typical.

PR #203, opened 2022-01-20
201 uneven bin widths

rebeccafair:
Addresses #201.

Main points:
* I have added a `method={None, 'convolve}` kwarg to `Spectrum1D/1DCollection/2D.broaden`. If there are unequal bins and `method=None` it will raise a `ValueError`, and if `method='convolve'` it will warn but continue. Default is `method=None`
* I have assumed that the x and y directions are independent i.e. if there are unequal bins in the x-direction, you can still broaden along the y-direction and it wont throw an error or warn. I think this is correct?
* I have set the broadening method for `euphonic-intensity-map` for CL tools to be 'convolve'. I didn't think there was much point in putting a `--broadening-method` in the CL tools for now since we don't actually have any alternatives... this will still warn the user though. I didn't set it for `euphonic-dos` or `euphonic-powder-map` as I think they will always have equal bins with our  current implementation, and at least this will alert us if that changes.


ajjackson:
> - I have added a method={None, 'convolve} kwarg to Spectrum1D/1DCollection/2D.broaden. If there are unequal bins and method=None it will raise a ValueError, and if method='convolve' it will warn but continue. Default is method=None

This is API breakage, if code following the current rules would raise an error. There should be a deprecation / transition period before the next major version bump.

> - I have assumed that the x and y directions are independent i.e. if there are unequal bins in the x-direction, you can still broaden along the y-direction and it wont throw an error or warn. I think this is correct?

As long it's 1-D broadening, yes. It may fall apart if we bring in broadening with 2-D covariance matrices, but at this stage we don't know how that would look.

> - I have set the broadening method for euphonic-intensity-map for CL tools to be 'convolve'. I didn't think there was much point in putting a --broadening-method in the CL tools for now since we don't actually have any alternatives... this will still warn the user though. I didn't set it for euphonic-dos or euphonic-powder-map as I think they will always have equal bins with our current implementation, and at least this will alert us if that changes.

Sounds good

rebeccafair:
> > * I have added a method={None, 'convolve} kwarg to Spectrum1D/1DCollection/2D.broaden. If there are unequal bins and method=None it will raise a ValueError, and if method='convolve' it will warn but continue. Default is method=None
> 
> This is API breakage, if code following the current rules would raise an error. There should be a deprecation / transition period before the next major version bump.

Good point, now fixed.

ajjackson (approved):
From the changeset it looks like some quartz Debye-Waller reference data is being deleted. Is that related to these changes, or some incidental cleanup?

Otherwise this looks good to me and behaves as expected.

rebeccafair:
> From the changeset it looks like some quartz Debye-Waller reference data is being deleted. Is that related to these changes, or some incidental cleanup?

Yeah, incidental cleanup. I wasn't convinced that the tests were actually using all the spectrum files so I ran [`pytest-filecov`](https://pypi.org/project/pytest-filecov/0.1.2/), and to my surprise it revealed those files, which have been there a surprisingly long time. Scrictly removing them should be a separate branch but I didn't think it was worth the effort!

PR #204, opened 2022-02-01
Kinematic constraints for powder maps

### LINKED ISSUE #200: Kinematic constraints on powder maps [nice to have]
The powder maps tool is extremely useful but it would be nice to be able to apply the kinematics to the plots (i.e. mask out regions which cannot be accessed). This should (in my opinion) be done in a simple and generic way where we just ask for a lowest and highest 2theta and the fixed Ei or Ef. More complex geometries should remain the preserve of users.

The Q of a detector for a given Ki and Kf is 

Q=np.sqrt(ki**2+kf**2-2*np.abs(ki*kf)*np.cos(ang*np.pi/180))
while the Energy is 
E=2.07*(ki**2-kf**2)
assuming that Ei and Ef are in meV and Ki and Kf are in angstrom. 

If we were to simply set all points which lie outside this range to NaN (which matplotlib appears to deal nicely with) it would give a nicer comparison to the experimental Q-E maps.

ajjackson:
Closes #200 

So far:

- Implement a kinematic constraint method for Spectrum2D
- Add some CLI arguments to powder_map.py
- Replace a few np.max with np.nanmax so that colour-mapping doesn't
  stumble over the masked-out values

Requirements to close:

- [x] Unit tests for kinematic constraint function
- [x] Script tests for euphonic-powder-map
- [x] Documentation for Python API
- [x] Changelog update


ajjackson:
David has said he's happy with the results this is producing, so I think all that remains is testing/docs.

ajjackson:
Ok, some basic docs are there (including an image) and the argument names are slightly less bad.

I'm still somewhat interested in tweaking the mask behaviour to be a bit more generous re: including bins. This would make very narrow slices (e.g. for individual detector angles) more practical, but perhaps that is not very helpful unless combined with a sum to 1D. Which sounds like a separate package of work.

euphonic/spectra.py
```diff
@@ -1297,6 +1297,89 @@ def from_dict(cls: Type[T], d: Dict[str, Any]) -> T:
...
+        q_bounds = q_bounds.real
+
+        new_z_data = np.copy(self.z_data.magnitude)
+        mask = np.logical_or((self.get_bin_centres(bin_ax='x')[:, np.newaxis]
+                              < q_bounds[0][np.newaxis, :]),
+                             (self.get_bin_centres(bin_ax='x')[:, np.newaxis]
+                              > q_bounds[1][np.newaxis, :]))
+
+        new_z_data[mask] = float('nan')
+        new_z_data[:, self.get_bin_centres(bin_ax='y').to('meV') > e_i
```
rebeccafair, line 1357: I'm a bit confused how this works when `e_f` is fixed so `e_i` is an array?

euphonic/spectra.py
```diff
@@ -1297,6 +1297,89 @@ def from_dict(cls: Type[T], d: Dict[str, Any]) -> T:
...
+
+        # Momentum goes negative where final energy greater than incident
+        # energy; detect this as complex component and set extreme q-bounds to
+        # enforce conservation of energy
+        q_bounds = np.sqrt(k2_i + k2_f
+                           - 2 * cos_values[:, np.newaxis]
+                               * np.sqrt(k2_i * k2_f, dtype=complex)
+                           )
+        q_bounds.magnitude.T[np.any(q_bounds.imag, axis=0)] = [float('Inf'),
+                                                               float('-Inf')]
```
rebeccafair, line 1347: There aren't any tests where there are imaginary `q_bounds`, it would be good to have one

euphonic/spectra.py
```diff
@@ -1297,6 +1297,89 @@ def from_dict(cls: Type[T], d: Dict[str, Any]) -> T:
...
+
+
+def _get_abs_angle_range(angle_range: Tuple[float]) -> Tuple[float]:
+    """Process full detector angle range to get useful range in abs values"""
+    min_angle, max_angle = sorted(angle_range)
+
+    # Actual limits depend on |2θ|, so some shuffling is necessary to get
+    # absolute values in the right order, spanning the whole range.
+    if (min_angle < 0) and (max_angle <= 0):
+        min_angle, max_angle = -max_angle, -min_angle
```
rebeccafair, line 1375: There are no tests where both `min_angle` and `max_angle` are negative, so this line is never tested.

tests_and_analysis/test/euphonic_test/test_spectrum2d.py
```diff
@@ -295,6 +296,33 @@ def test_incorrect_unit_conversion(self, spectrum2d_file, attr,
...
+    def test_kinematic_constraints(self, kwargs,
+                                   spectrum2d_file, constrained_file):
+        spec2d = get_spectrum2d(spectrum2d_file)
+        ref_spec2d = get_spectrum2d(constrained_file)
+
+        constrained_spec2d = spec2d.apply_kinematic_constraints(**kwargs)
+        check_spectrum2d(constrained_spec2d, ref_spec2d, equal_nan=True)
+
+    @pytest.mark.parametrize('kwargs, expected',
+                             [({'e_i': 10, 'e_f': 20}, ValueError)])
```
rebeccafair, line 319: The case where neither `e_i` or `e_f` are set should also be tested

rebeccafair (changes requested):
Looks pretty good! Just a few things in the comments, and:

* Tests for the kinematic constraints in `euphonic-powder-map` need to be added
* Changelog needs to be updated
* If `apply_kinematic constraints` is to be a public function, there should be some docs on that, maybe another page in the 'Python API' section? For now we can just say it applies to (|Q|, E) maps, but maybe in the future it can be extended to apply to vectors and work on (Q, E) maps, although I don't know how difficult that would be.  At the very least it would require additional input of the vector q-points as the Spectrum2D only keeps the scalar distance between them.

mducle:
> For now we can just say it applies to (|Q|, E) maps, but maybe in the future it can be extended to apply to vectors and work on (Q, E) maps, although I don't know how difficult that would be. At the very least it would require additional input of the vector q-points as the Spectrum2D only keeps the scalar distance between them.

I don't think you need to know the vector **q** for each point in the `Spectrum2D` if you're restricting it to 2D (*Q_alpha*,E) maps where *alpha* is a particular direction in reciprocal space - you'd just need to know the vector *alpha*. The kinematic envelope would be a 3D parabolic surface in the 4D space and we would just be taking a 2D slice of the 4D space (and hence a 1D line through the 3D kinematic envelope - we'd just need to know the direction of the slice). 

(This is assuming that we approximate the detectors by (min,max) angular extents in (theta,phi) - giving us a spherical quadrilateral detectors; more realistic detector geometries would make the maths pretty much impossible I think; in Horace we just brute force it by loading a detector parameter file which lists every detector pixel then we calculate the (Q,E) position of every pixel to get the limits.)

You would also need to know the orientation of the crystal; and there is an addition complication in that in experiments we almost always combine measurements at multiple orientations, which means that the kinematic limits is a set (convex hull?) of multiple parabolic surfaces...

euphonic/cli/powder_map.py
```diff
@@ -156,7 +177,16 @@ def main(params: Optional[List[str]] = None) -> None:
                      if args.energy_broadening else None),
             shape=args.shape)
 
-    print(f"Plotting figure: max intensity {np.max(spectrum.z_data):~P}")
+    if args.angle_range is not None:
+        print("Applying kinematic constraints")
+        energy_unit = args.energy_unit
+        e_i = args.e_i * ureg(energy_unit) if (args.e_i is not None) else None
+        e_f = args.e_f * ureg(energy_unit) if (args.e_f is not None) else None
+        spectrum = spectrum.apply_kinematic_constraints(
```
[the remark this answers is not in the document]
ajjackson: Argparse will catch it, they are in an exclusive argument group together.

In the help page this is indicated as
```
[--e-incident E_I | --e-final E_F]
```

euphonic/cli/powder_map.py
```diff
@@ -52,6 +53,20 @@ def get_parser() -> ArgumentParser:
...
+
+    e_lims.add_argument('--e-incident', '--e-i', dest='e_i',
+                        type=float, default=None,
+                        help="Incident energy for direct-geometry constraints")
+    e_lims.add_argument('--e-final', '--e-f', dest='e_f',
+                        type=float, default=None,
+                        help="Final energy for indirect-geometry constraints")
+    kinematic.add_argument('--angle-range', nargs=2, type=float, default=None,
+                           dest='angle_range',
+                           help="Detector angle range")
```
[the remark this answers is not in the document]
ajjackson: I think it will be clear if we mention 2θ

euphonic/spectra.py
```diff
@@ -1297,6 +1297,89 @@ def from_dict(cls: Type[T], d: Dict[str, Any]) -> T:
...
+
+        Either e_i or e_f should be set, according to direct/instrument
+        geometry. The other values will be inferred, interpreting y_data as
+        energy transfer.
+
+        """
+
+        momentum2_to_energy = 0.5 * ureg('hbar^2 / neutron_mass').to('meV angstrom^2')
+
+        if (e_i is None) == (e_f is None):
```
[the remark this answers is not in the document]
ajjackson: Good old XNOR :smile:

euphonic/spectra.py
```diff
@@ -1297,6 +1297,89 @@ def from_dict(cls: Type[T], d: Dict[str, Any]) -> T:
...
+            raise ValueError("Only one of e_i and e_f should be set. "
```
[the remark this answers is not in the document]
ajjackson: Wouldn't hurt, clarity is good

euphonic/spectra.py
ajjackson: Good point. Just need one with a higher incident energy.

ajjackson:
Right, at that point the model complexity falls into "need to know a lot about the instrument" which we've generally said should fall on packages outside Euphonic.

euphonic/cli/powder_map.py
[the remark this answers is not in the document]
ajjackson: The bigger issue is the case where neither is set. I am changing the logic so that instead of checking whether angle_range has a value, angle_range is given a default (0, 180) and the constraints are enabled if e_i or e_f was set. This maps more closely to the underlying function, and gives a kind of "ideal spherical instrument" default.

ajjackson:
Here's an interesting quirk of the angle-range logic:

Angle range of 0-180
sweep1

Angle range of 0-270
sweep2

I think this was effectively converted to a range of 0-90. It's easy to see the problem, but a _really_ robust solution might need to be a bit smarter.

euphonic/spectra.py
```diff
@@ -1297,6 +1297,89 @@ def from_dict(cls: Type[T], d: Dict[str, Any]) -> T:
...
+        new_z_data[mask] = float('nan')
+        new_z_data[:, self.get_bin_centres(bin_ax='y').to('meV') > e_i
+                   ] = float('nan')
+
+        return type(self)(
+            np.copy(self.x_data.magnitude) * ureg(self.x_data_unit),
+            np.copy(self.y_data.magnitude) * ureg(self.y_data_unit),
+            new_z_data * ureg(self.z_data_unit),
+            copy.copy(self.x_tick_labels),
+            copy.copy(self.metadata))
```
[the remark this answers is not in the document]
ajjackson: Can't hurt

euphonic/spectra.py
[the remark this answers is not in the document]
ajjackson: It is two sentences, though! They could be joined be a (semi)colon and no parentheses, but I don't agree with starting a new sentence in parens before the end of this one.

euphonic/spectra.py
ajjackson: It turns out this line can be removed and nothing changes. It explicitly applies energy conservation, but the region is already taken out by momentum conservation (where the accessible range goes imaginary).

euphonic/cli/powder_map.py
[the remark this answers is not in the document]
rebeccafair: Seems a good solution to me! In that case we should make sure there is a `euphonic-powder-map` test with just `--e-i` or `--e-f` set with no angles.

euphonic/spectra.py
```diff
@@ -1297,6 +1297,89 @@ def from_dict(cls: Type[T], d: Dict[str, Any]) -> T:
                    x_tick_labels=d['x_tick_labels'],
                    metadata=d['metadata'])
 
+    def apply_kinematic_constraints(self: T,
```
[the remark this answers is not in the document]
rebeccafair: I can't think of any other cases off the top of my head, though I'm sure there must be some...

Perhaps we could put it in `spectra.py` for now, and if we end up having more turn spectra into a package and have a `spectra.util` module and import the functions into the `spectra` namespace? Although I don't know if having 2 files with the same name is a bad idea, even if they're in different packages.

euphonic/spectra.py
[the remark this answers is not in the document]
rebeccafair: Fair enough. Having a sentence only in parentheses looked weird to me, but I'm no grammar expert. Maybe the parentheses aren't needed at all, but I'll leave that up to you.

ajjackson:
Angle range problem analysis:
- the problem we are actually trying to solve is to find the minimum and maximum of the cosine function within the angle range.
- in "problem" cases, a turning point exists within the given range, introducing a minimum or maximum that does not lie at the start/end points. We know exactly where these points will be, but there is an infinite number of them: maxima at ..., -2π, 0, 2π, 4π... and minima at ...-π, π, 2π, ...

So for a reliable algorithm to identify the min/max:
- write down the cosines of the endpoints
- modulo the lower endpoint to the region 0-2pi and slide the higher endpoint by the same amount
- if the translated region includes 2π, append 1 to the set of cosine values
- if the translated region includes π, append -1 to the set of cosine values
- take the min and max of the set of cosine values

euphonic/spectra.py
ajjackson: Done! The same case checks for a nasty angle as well.

euphonic/spectra.py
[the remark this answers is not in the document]
ajjackson: Will be easy enough to manage migration later with an import and deprecation statement :+1:

ajjackson:
This test failure is baffling. The x-ticks are apparently different between oldlibs and new - possibly related to the way the odd-number q-limit is handled? The case is ` --e-f=15 --ebins 50 --q-max=15 --angle-range 20 230 --npts=10 --npts-min=10 --q-spacing=1`

When I manually run this with old an new libs, the complained missing value of 16 isn't even on the plot, in either case. So I guess this is some change in how Matplotlib handles xticks and range limits, but it's insignificant to the output.

Screenshot from 2022-03-23 10-48-22

ajjackson:
This discrepancy seems totally unimportant so let's dodge it by using an even-numbered q-range in the test case.

doc/source/powder-map-script.rst
```diff
@@ -49,6 +49,39 @@ To see all the command line options, run::
...
+is installed
+
+Kinematic constraints
+---------------------
+
+Inelastic neutron-scattering measurements have an accessible
+:math:`(\mathbf{q}, \omega)` range depending on the instrument
+geometry: either the incident energy is fixed (direct geometry) or the
+final energy is fixed (indirect geometry), and the scattering angles
+are determined by detector/reflector positions. The remaining degrees
```
mducle, line 66:
```suggestion
are determined by detector/analyzer positions. The remaining degrees
```

doc/source/spectra.rst
```diff
@@ -317,6 +317,66 @@ Plotting
...
+(e.g. by scattering from an "analyser" crystal).
+Conservation laws allow the overall energy and momentum
+transfer to be determined for a given scattering angle --- but this also
+means that the accessible :math:`(q, \omega)` range is restricted by
+these instrument parameters.
+
+The function :py:func:`apply_kinematic_constraints
+<euphonic.spectra.apply_kinematic_constraints>` applies these limits
+to a :ref:`Spectrum2D` object with appropriate dimensions (i.e. the x-
+and y-axes represent :math:`|q|` and :math:`\omega` respectively).
```
mducle, line 336:
```suggestion
to a powder averaged :ref:`Spectrum2D` object with appropriate dimensions
(i.e. the x-and y-axes represent :math:`|q|` and :math:`\omega` respectively).
```

ajjackson: Re-read it in context and you had that bang on. Cheers.

doc/source/spectra.rst
```diff
@@ -317,6 +317,74 @@ Plotting
...
+(e.g. by scattering from an "analyser" crystal).
+Conservation laws allow the overall energy and momentum
+transfer to be determined for a given scattering angle and crystal
+orientation. In powder measurements, the crystal orientation is not
+needed so the kinematic limits --- the accessible :math:`(q, \omega)`
+range --- determined by the conservation laws are given solely
+by these instrument parameters.
+
+The function :py:func:`apply_kinematic_constraints
+<euphonic.spectra.apply_kinematic_constraints>` applies these limits
```
rebeccafair, line 337: It would be good to show the function is `euphonic.spectra.apply_kinematic_constraints` otherwise it's not clear where to import it from!

euphonic/spectra.py
```diff
@@ -1297,6 +1298,119 @@ def from_dict(cls: Type[T], d: Dict[str, Any]) -> T:
...
+    Args:
+        spectrum: input 2-D spectrum, with |q| x_data and energy y_data
+        e_i: incident energy of direct-geometry spectrometer
+        e_f: final energy of indirect-geometry spectrometer
+        angle_range: min and max scattering angles (2θ) of detector bank in
+            degrees.
+
+    Returns:
+        Masked spectrum with inaccessible bins set to NaN in z_data.
+    """
```
rebeccafair, line 1326: This should be a Numpy-style docstring like other docstrings in Euphonic. And the """ should be on a line on its own with the text starting on a new line.

euphonic/spectra.py
```diff
@@ -1297,6 +1298,119 @@ def from_dict(cls: Type[T], d: Dict[str, Any]) -> T:
...
+    return Spectrum2D(
+        np.copy(spectrum.x_data.magnitude) * ureg(spectrum.x_data_unit),
+        np.copy(spectrum.y_data.magnitude) * ureg(spectrum.y_data_unit),
+        new_z_data * ureg(spectrum.z_data_unit),
+        copy.copy(spectrum.x_tick_labels),
+        copy.deepcopy(spectrum.metadata))
+
+
+def _get_cos_range(angle_range: Tuple[float]) -> Tuple[float]:
+    """Get max and min of cosine function over angle range
```
rebeccafair, line 1396: """ should be on its own line, start text on a new line

rebeccafair (changes requested):
Just some very minor things and I think we're done.

Good spot and fix on the 0, 270 angle ranges etc!

tests_and_analysis/test/euphonic_test/test_spectrum2d.py
```diff
@@ -431,3 +432,60 @@ def test_get_bin_centres_with_invalid_data_shape_raises_value_error(
...
+
+    @pytest.mark.parametrize('json_file, kwargs, expected',
+                             [('NaCl_band_yaml_dos_map.json',
+                               {'e_i': 10 * ureg('1/cm'),
+                                'e_f': 20 * ureg('1/cm')},
+                               ValueError),
+                              ('NaCl_band_yaml_dos_map.json', {}, ValueError),
+                              ('example_spectrum2d.json',
+                               {'e_i': 20 * ureg('1/cm')},
+                               ValueError)])
```
rebeccafair, line 486: This tests the case where y_data is invalid ('dimensionless') but not x_data. We could do with an example where x_data is in invalid units too, but there don't seem to be any in the current test data. Maybe change the x units of 'example_xbin_edges_spectrum2d.json' to something silly? I don't think that would have an adverse affect on any other tests.

tests_and_analysis/test/script_tests/test_powder_map.py
```diff
@@ -43,7 +43,11 @@
      *quick_calc_params],
     [graphite_fc_file, '--asr', *quick_calc_params],
     [graphite_fc_file, '--asr=realspace', '--dipole-parameter=0.75',
-     *quick_calc_params]]
+     *quick_calc_params],
+    [graphite_fc_file, '--e-i=15', '--e-max=20', '--ebins=50',
+     *quick_calc_params],
```
rebeccafair, line 48: To fully cover the `if args.e_i is not None and args.e_max is None:...` branches we should also have a test where `--e-i` is set without `--e-max` 😢

rebeccafair (changes requested):
Sorry, just checked the codecov

euphonic/spectra.py
```diff
@@ -1297,6 +1298,122 @@ def from_dict(cls: Type[T], d: Dict[str, Any]) -> T:
...
+    Either e_i or e_f should be set, according to direct/instrument
+    geometry. The other values will be inferred, interpreting y_data as
+    energy transfer.
+
+    Parameters
+    ----------
+    spectrum: input 2-D spectrum, with |q| x_data and energy y_data
+    e_i: incident energy of direct-geometry spectrometer
+    e_f: final energy of indirect-geometry spectrometer
+    angle_range: min and max scattering angles (2θ) of detector bank in degrees
```
rebeccafair, line 1323: If no type is specified, the description of each parameter should be indented on the following line, and there should be no `:` https://numpydoc.readthedocs.io/en/latest/format.html#parameters

rebeccafair (changes requested):
This is the last comment I promise...

PR #205, opened 2022-02-04
Enable saving as JSON from CL tools

rebeccafair:
Partially addresses #130

I decided to just do the simple thing for now and just add a `--save-json` option to the CL tools, as this is sorely needed by the `euphonic-powder-map` tool and has been requested for quite a while...

In doing so I found a strange bug, in that the values produced by `euphonic.cli.utils._get_tick_labels` are actually of type `List[Tuple[np.int32, str]]` rather than using plain `int`, so it was not JSON serialisable! This is due to the call to `np.where`, which produces an array of type `np.int64`. I could've probably just fixed this in that function, but thought that setting the `x_tick_labels` on an existing object is actually probably a common thing to do, especially with values produced by `np.where`. So instead I fixed it in the `Spectrum` object itself, and we actually weren't doing much type checking on x_tick_labels when we probably should be.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,6 +7,9 @@
     ``euphonic-intensity-map`` command-line tools can now read
     files that don't contain eigenvectors, if eigenvectors are
     not required for the chosen options.
+  - A new ``--save-json`` option is available for command-line tools
+    which produce plots, this will output the produced spectrum to
+    a Euphonic .json file.
   - There is now the option to use a fast, approximate variable-width broadening method when
     adaptively broadening dos:
 
--- a/doc/source/powder-map-script.rst
+++ b/doc/source/powder-map-script.rst
@@ -4,7 +4,12 @@
 euphonic-powder-map
 ======================
 
+.. contents:: :local:
+
 .. highlight:: bash
+
+Overview
+--------
 
 The ``euphonic-powder-map`` program can be used to sample
 spherically-averaged properties from force constants data over a range
@@ -35,6 +40,7 @@ q range with denser sampling, in THz and with the intensity widget disabled::
    :alt: 2D intensity map with |q| from 0.01 to 4.0 on the x-axis,
          and energy on the y axis, showing powder-averaged coherent
          inelastic neutron scattering intensities for NaCl.
+
 
 To see all the command line options, run::
 
@@ -72,6 +78,13 @@ Progress bars
 Sampling many q-points can be computationally expensive, so a progress
 bar will automatically be displayed if `tqdm <https://tqdm.github.io/>`_
 is installed
+
+Output to file
+--------------
+
+The ``--save-json`` option can be used to output the produced
+:ref:`Spectrum2D` object as a Euphonic .json file with a specified
+name for further use in Euphonic or other programs.
 
 Command Line Options
 --------------------
--- a/euphonic/cli/dispersion.py
+++ b/euphonic/cli/dispersion.py
@@ -52,6 +52,8 @@ def main(params: Optional[List[str]] = None) -> None:
     style = _compose_style(user_args=args,
                            base=[base_style])
 
+    if args.save_json:
+        spectrum.to_json_file(args.save_json)
     with matplotlib.style.context(style):
         _ = plot_1d(spectra,
                     ymin=args.e_min,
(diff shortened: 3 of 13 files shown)


ajjackson:
Looks clean and useful!

I'm slightly surprised that no code had to be removed/replaced to implement the new x_tick_labels property. I guess this is all handled by the constructor magic?

ajjackson:
I tried saving JSON from powder-map and replotting with `euphonic-intensity-map`, but get this error message. It looks like there is no support for Spectrum1D/Spectrum2D in _load_euphonic_json. Are we planning to add that?

```
(science) Arctopus:~/src/euphonic (cl_save_json)$ euphonic-intensity-map mp200.json
  File "/home/adam/.conda/envs/science/bin/euphonic-intensity-map", line 33, in <module>
    sys.exit(load_entry_point('euphonic', 'console_scripts', 'euphonic-intensity-map')())
  File "/home/adam/src/euphonic/euphonic/cli/intensity_map.py", line 25, in main
    data = load_data_from_file(args.filename, verbose=True,
  File "/home/adam/src/euphonic/euphonic/cli/utils.py", line 152, in load_data_from_file
    data = _load_euphonic_json(path, frequencies_only)
  File "/home/adam/src/euphonic/euphonic/cli/utils.py", line 62, in _load_euphonic_json
    raise ValueError("Could not identify Euphonic data in JSON file.")
ValueError: Could not identify Euphonic data in JSON file.***
```

rebeccafair:
> I tried saving JSON from powder-map and replotting with `euphonic-intensity-map`, but get this error message. It looks like there is no support for Spectrum1D/Spectrum2D in _load_euphonic_json. Are we planning to add that?
> 
> ```
> (science) Arctopus:~/src/euphonic (cl_save_json)$ euphonic-intensity-map mp200.json
> Traceback (most recent call last):
>   File "/home/adam/.conda/envs/science/bin/euphonic-intensity-map", line 33, in <module>
>     sys.exit(load_entry_point('euphonic', 'console_scripts', 'euphonic-intensity-map')())
>   File "/home/adam/src/euphonic/euphonic/cli/intensity_map.py", line 25, in main
>     data = load_data_from_file(args.filename, verbose=True,
>   File "/home/adam/src/euphonic/euphonic/cli/utils.py", line 152, in load_data_from_file
>     data = _load_euphonic_json(path, frequencies_only)
>   File "/home/adam/src/euphonic/euphonic/cli/utils.py", line 62, in _load_euphonic_json
>     raise ValueError("Could not identify Euphonic data in JSON file.")
> ValueError: Could not identify Euphonic data in JSON file.***
> ```

Yep! I did consider enabling reading of JSON, but once I started ran into a couple of things and decided I didn't have time to do it and thought it was important to just get JSON output in first.

Maybe replotting should be a separate tool - `euphonic-plot`? As part of the replotting I also thought we could allow broadening to be added, but I was thinking that if we allow reading of arbitrary `Spectra` the `--energy-broadening` and `--q-broadening` arguments don't make much sense, and it should instead be `--x-broadening` or `--y-broadening`, which would require some CL parser refactoring and... I started to feel like I was going down a rabbit hole 😅

ajjackson:
If it's expensive to generate a spectrum, the generating it once without broadening and plotting a few times with different broadening would be quite useful.

ajjackson:
I guess a `euphonic-plot` would be quite a small program, and could provide features such as x/y axis swapping.

But there will be a slight nuisance when it comes to creating a parser that does similar things to other programs, with different argument names, without redundancy 🤔

rebeccafair:
Indeed! And I just remembered we'd discussed storing some plot arguments in the spectrum metadata a while ago, but never decided exactly how to implement it, but this is something that could be quite useful for this plot tool too. It seems like a plotting CL tool should be a simple task but I'm feeling a lot of scope creep here...

ajjackson:
Yes, that is starting to sound like a separate PR. I suppose I will need to write a script to test this on in the mean time 😅

rebeccafair:
I guess I could try and put a simple one in this PR, basically just include the features already in the 'plotting' CL parser and nothing else? Maybe then it will become clearer what other features are needed.

ajjackson:
I don't see that it would invalidate the implementation in this PR, just extend it maybe. So perhaps it's better to put off and discuss in more detail.

ajjackson (approved):
Implementation seems clean.

Have tried saving files and viewing from Python interpreter, results seem consistent. Clobbering an existing filename works.

We can do more to make this user-friendly; currently CLI tools don't support the JSON files and from the Python interface you need to know in advance whether to use Spectrum1D/Collection/2D. (Maybe that's fine.)

But work on that can be driven by user need as it arrises; already these JSON dumps will be useful for creating composite plots and tinkering with broadening options.

PR #207, opened 2022-02-16
206 Brille sampling tool

### LINKED ISSUE #206: Error estimation for Brille interpolation
Building on #95  and playing with the implementation in #104 it would be useful to get some information about the error associated with linear interpolation.

I propose to achieve this by
- implementing 3-d quasirandom sampling of the Brillouin zone to obtain a set of low-symmetry reference points
- creating a command-line tool that compares linearly-interpolated points from Brille with "exact" Fourier-interpolated points from Euphonic

If appropriate this might lead to sampling recommendations or an automatic convergence scheme.

ajjackson:
Closes #206 

This introduces a tool for error estimation and analysis of Brille-interpolated results.

### DIFF
--- a/euphonic/brille.py
+++ b/euphonic/brille.py
@@ -131,6 +131,7 @@ class BrilleInterpolator:
     def from_force_constants(
             cls: Type[T], force_constants: ForceConstants,
             grid_type: str = 'trellis', n_grid_points: int = 1000,
+            grid_density: Optional[int] = None,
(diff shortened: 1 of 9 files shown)


rebeccafair:
I've added an option to specify the Brille grid density, and some plots comparing the structure factors and intensities. It would be good to have some feedback on whether the Brille grid density is a useful metric, which plots are useful or not, and any tweaks that could be made @ajjackson @mducle?

Issue #210, opened 2022-03-02
Problem reading Phonopy with 'au' units


rebeccafair:
[Some Phonopy calculators](https://phonopy.github.io/phonopy/interfaces.html) use 'au' as a unit meaning Bohr, unfortunately Pint interprets this to mean Astronomical Unit, leading to some interesting results...

Thanks @ajjackson for finding this.

alinelena:
replacing au with bohr in the .yaml seems to do the trick. in theory I suspect phonopy shall write it correctly. maybe euphonics can intercept au and change it to bohr.. but I still think phonopy shall write the proper unit.

PR #211, opened 2022-03-03
Allow read of Phonopy 'au' units

rebeccafair:
Addresses #210


ajjackson (approved):
This seems fine.

The implementation
```python
pu['force_constants'] = pu['force_constants'].replace('/', '/(') + ')'
```
does look a little fragile, e.g. it will give unbalanced parentheses if more than one '/' was present in the input. I don't see a good way around that, though; if Phonopy has standardised on this format with one division we have to work with it. And it is better to break on unexpected input than, say, to accept it and be off by many orders of magnitude 😆 .

I can now read the production data from Alin's collaborator. I'll approve now, but maybe we should give a few days before merging to see if we get any other sample data?

rebeccafair:
> The implementation
> 
> ```python
> pu['force_constants'] = pu['force_constants'].replace('/', '/(') + ')'
> ```
> 
> does look a little fragile

Yeah, I am slightly worried about that but I couldn't think of a better way of doing it without overcomplicating it... Hopefully we've found most of the possibilities now!

And yep I'll hold off on merging this for a bit

mducle:
Use regular expressions?

```python
import re
divs = re.findall(r'/([A-z\.\^0-9]*)', pu['force_constants'])
if divs:
    pu['force_constants']= pu['force_constants'].split('/')[0] + '/(' + '/('.join([d+')' for d in divs])
```

ajjackson:
That regex will still choke if it encounters parentheses.
With a `findall` regex of `r"/([^/]+)"` I can feed it something truly monstrous like

```
"Ry/Angstrom.(bohr/bohr/bohr)/eV"
```
and get `'Ry/(Angstrom.(Angstrom)/(bohr)/(bohr.bohr))/(eV)'` which looks ok.

Were there characters apart from `/` that you wanted to exclude from matches?

rebeccafair:
> ```
> "Ry/Angstrom.(bohr/bohr/bohr)/eV"
> ```
I wonder if we shouldn't be encouraging this kind of behaviour :satisfied:

The regex does seem more robust, I'm happy to go with that

ajjackson:
> I wonder if we shouldn't be encouraging this kind of behaviour laughing

I certainly don't intend to document it :see_no_evil:

ajjackson:
This works on the test case and on other live data we've seen so far. I suggest merging as-is, and we can open a new bug report if another file turns up that doesn't work.

Issue #212, opened 2022-03-16
Remove deprecations before 1.0


rebeccafair:
Any deprecated arguments or functions should be removed before a 1.0 release.

rebeccafair:
Addressed in #213, merged to master in #231

PR #213, opened 2022-03-18
Remove deprecations before 1.0

rebeccafair:
Addresses #212. This removes deprecated functions/arguments before a 1.0 release.

There is some 'deprecation helper' functionality that is no longer used, e.g. the `deprecated_arg` function in `euphonic.cli.utils._get_cli_parser` and `euphonic.util._deprecation_warn`. This could be removed but could be handy for future deprecations. If removed someone could recover it from the git history but only if they know it's there in the first place.

This is a draft for now awaiting changes from `horace-euphonic-interface` and `euphonic_sqw_models`.

**Update:**
[horace-euphonic-interface](https://github.com/pace-neutrons/horace-euphonic-interface/tree/test_remove_deprecations) and [euphonic_sqw_models](https://github.com/pace-neutrons/euphonic_sqw_models/tree/test_remove_deprecations) branches testing against this branch work without any changes to their code. They don't use any deprecated code so nothing to do here!


ajjackson:
Would it make sense to target a `release-v1` branch rather than `master` with this? Then there is an opportunity for other PRs to decide which API version they are targeting. (I'm particularly thinking of the CASTEP averaging issue, which will change results if on by default.)

CHANGELOG.rst
```diff
@@ -1,6 +1,22 @@
 `Unreleased <https://github.com/pace-neutrons/Euphonic/compare/v0.6.4...HEAD>`_
 ----------
 
+- Changes:
+
+  - The following deprecated features have been removed:
+
+    - The ``return_mode_widths`` argument in ``calculate_qpoint_phonon_modes/frequencies``
```
ajjackson, line 8: Should this be `ForceConstants.calculate_qpoint_phonon_modes() and ForceConstants.calculate_qpoint_frequencies()`? It's not obvious what functionality is affected.

CHANGELOG.rst
```diff
@@ -1,6 +1,22 @@
...
+      has been removed
+    - The ``eta_scale`` argument in ``calculate_qpoint_phonon_modes/frequencies``
+      has been removed
+    - The ``--weights`` command-line tool argument has been removed
+    - The ``x_label``, ``y_label``, ``y_min`` and ``y_max`` arguments to
+      ``plot_1d/2d`` have been removed
```
ajjackson, line 14: For the matplotlib options and ``--weights`` it might be more reassuring at a glance to make it clear that these were aliases and the functionality is still there.

ajjackson (commented):
This looks like a very welcome cleanup, [+37 -8,040] is a wonderful thing to see. (Although presumably dominated by clearing out unused test data?)

I'd like to clarify the release situation before approving a merge to master. Is it worth looking over the other high-profile Issues/PRs and deciding which others should be in v1 and how that affects the merge strategy?

rebeccafair:
> Would it make sense to target a `release-v1` branch rather than `master` with this? Then there is an opportunity for other PRs to decide which API version they are targeting. (I'm particularly thinking of the CASTEP averaging issue, which will change results if on by default.)
> 
> I suppose it doesn't matter if there will be no more release versions before then!
> 
> Mantid has its Conda euphonic dependency pinned to 0.6.* at the moment, so it _should_ be ok even though it uses `force_constants_from_file`.

I don't think there will be a release before then, but good idea just to make sure we get everything finalised before release, I also want to make a few changes to the CITATION.cff (PR incoming).

I think the CASTEP averaging issue is the only other thing that should go into 1.0, but it would have to be soon. Do we know exactly what needs to be done for this?

ajjackson:
PR is open https://github.com/pace-neutrons/Euphonic/pull/221.

I suggest the PR target master initially so we get the side-by-side comparison tests (with functionality hidden behind non-default options). Then a second PR targets v1.0_release, and either
- change it to on-by-default, or
- drop the old behaviour entirely, and legacy test data

CHANGELOG.rst
rebeccafair: Done!

ajjackson (approved):

PR #218, opened 2022-05-31
Detect repeated q-points in band-structure data

### LINKED ISSUE #202: Missing symmetry labels from Phonopy band structures
If I generate a band structure from Phonopy, and plot with euphonic-dispersion, e.g.

```python
phonopy phonopy.conf -c POSCAR-unitcell --band '0 0 0 0.5 0.5 0 0.5 0. 0.' --band-points 1001 --band-format hdf5 --band-labels 'Gamma Y V' --eigenvectors

euphonic-dispersion band.hdf5 -s band.png --reorder
```

the resulting plot correctly identifies symmetry labels for the first and last high-symmetry points, but misses the one(s) in the middle. I get the same behaviour with `euphonic-intensity-map`.

band

I think the direction_changed algorithm is failing to pick up the turning points because at special points the q-points are duplicated? (And so there is no direction change between them...) The answer could be to add another case to `_recip_space_labels` so that redundant points are always added to the shortlist for labelling.

ajjackson:
Fixes #202 

When importing QpointFrequencies/QpointPhononModes on a high-symmetry path it is possible that a repeated non-Gamma q-point will not be identified as a turning point. Conventionally these repeated points are used to indicate points-of-interest in band structure data (i.e. places to insert a vertical line and a label).

It seems harmless to check for these while looking for Gamma and turning points and add them to the "qpt_has_label" list, and will occasionally catch a missing point. Sometimes special points are marked in the middle of a high-symmetry line without direction change,

That being said, the example in #202 _does_ have a direction change so I don't quite understand why it isn't caught; maybe there is still a bug there to catch/fix?

- [x] Add a test case
- [x] Check the new behaviour isn't contradicted in the docs somewhere


rebeccafair:
> That being said, the example in #202 _does_ have a direction change so I don't quite understand why it isn't caught; maybe there is still a bug there to catch/fix?

Just looked into this, in `util.direction_changed`, the vector between q-points `delta` is calculated, which for repeated q-points is zero. To find the relative direction of these vectors the dot product `dot` of each `delta` with the next is calculated, which for the repeated q-points will also be zero! So the test to see if it is more than `tolerance` returns `False`. I can't immediately think of a good way around this, maybe it would be enough for now to state in the `direction_changed` docstring that it wont correctly determine direction changes for repeated points?

ajjackson:
The fix here, to explicitly include any repeated points, should catch it either way?

ajjackson:
I don't see anywhere in the docs we make promises about how these labels will be detected, except in the `get_qpoint_labels` docstring which I have just updated.

rebeccafair (approved):
Looks good! The CI doesn't seem to be running after the force push though...

We had this problem earlier with spinw, could try adding `synchronize` to the PR types? https://github.com/SpinW/spinw/commit/9782aed3fdea6546966001b7d92ceeea77f0c927 I'm happy to merge without it though, I'm sure they'll pass

ajjackson:
Interesting! I guess it's possible to manually start CI after force-push if needed?

rebeccafair:
Yeah, I forgot you could manually re-run with the run_ci label, but I think synchronize might actually be better (you might have to remove/add the label on each commit I think).

PR #220, opened 2022-06-01
Small doc updates

rebeccafair:
* Updates the equations in the docstrings to match the conventions in the Euphonic paper
* Other small miscellaneous updates for clarity

### DIFF
--- a/doc/source/force-constants.rst
+++ b/doc/source/force-constants.rst
@@ -24,14 +24,12 @@ The :py:attr:`ForceConstants.force_constants <euphonic.force_constants.ForceCons
   \frac{\partial^{2}E}{{\partial}u_{\kappa\alpha0}{\partial}u_{{\kappa}'{\alpha}'l}}
 
 This describes the change in total energy when atom :math:`\kappa` in unit cell :math:`0`` is displaced in direction :math:`\alpha` and atom :math:`\kappa\prime` in unit cell :math:`l` is displaced in direction :math:`\alpha\prime`.
-In Euphonic the force constants are stored in 'compact' form, which means that the minimum required information is stored.
-Using equivalent vectors, the force constants between every atom in the unit cell, and every atom in the supercell are all that is needed,
-so only :math:`N(3n)^2` values need to be stored rather than :math:`(3Nn)^2` as in the 'full' square force constants matrix,
-where N is the number of unit cells in the supercell, and n is the number of atoms in the unit cell.
+In Euphonic the force constants are stored in 'compact' form, with shape :math:`3Nn\times3n`, where :math:`N` is the number of cells in the supercell, and :math:`n` is the number of atoms in the unit cell.
+This form can be expanded as needed to a 'full' :math:`3Nn\times3Nn` matrix of supercell atom pairs by mapping to an equivalent vector from unit cell 0.
 
 **Shape and Indexing**
 
-The force constants matrix has shape ``(N, 3*n, 3*n)``, where N is the number of unit cells in the supercell, and n is the number of atoms in the unit cell.
+The force constants matrix has shape ``(N, 3*n, 3*n)``, where ``N`` is the number of unit cells in the supercell, and ``n`` is the number of atoms in the unit cell.
 The force constants index ``[l, a, b]``, where ``a = 3*kappa + alpha`` and ``b = 3*kappa_prime + alpha_prime``,
 describes the change in energy when atom ``kappa`` in unit cell ``0`` is displaced in direction ``alpha`` and atom ``kappa_prime`` in unit cell ``l`` is displaced in direction ``alpha_prime``.
 For example, ``ForceConstants.force_constants[5, 8, 1]`` is the change in energy when atom
--- a/doc/source/force-constants.rst
+++ b/doc/source/force-constants.rst
@@ -20,10 +20,10 @@ The :py:attr:`ForceConstants.force_constants <euphonic.force_constants.ForceCons
 
 .. math::
 
-  \phi_{\alpha, {\alpha}'}^{\kappa, {\kappa}'} =
-  \frac{\delta^{2}E}{{\delta}u_{\kappa,\alpha}{\delta}u_{{\kappa}',{\alpha}'}}
+  \Phi_{\alpha {\alpha}'}^{\kappa {\kappa}'}(0,l) =
+  \frac{\partial^{2}E}{{\partial}u_{\kappa\alpha0}{\partial}u_{{\kappa}'{\alpha}'l}}
 
-This describes the change in total energy when atom :math:`\kappa` is displaced in direction :math:`\alpha` and atom :math:`\kappa\prime` is displaced in direction :math:`\alpha\prime`.
+This describes the change in total energy when atom :math:`\kappa` in unit cell :math:`0`` is displaced in direction :math:`\alpha` and atom :math:`\kappa\prime` in unit cell :math:`l` is displaced in direction :math:`\alpha\prime`.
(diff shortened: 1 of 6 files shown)


doc/source/force-constants.rst
```diff
@@ -20,10 +20,10 @@ The :py:attr:`ForceConstants.force_constants <euphonic.force_constants.ForceCons
...
 .. math::
 
-  \phi_{\alpha, {\alpha}'}^{\kappa, {\kappa}'} =
-  \frac{\delta^{2}E}{{\delta}u_{\kappa,\alpha}{\delta}u_{{\kappa}',{\alpha}'}}
+  \Phi_{\alpha {\alpha}'}^{\kappa {\kappa}'}(0,l) =
+  \frac{\partial^{2}E}{{\partial}u_{\kappa\alpha0}{\partial}u_{{\kappa}'{\alpha}'l}}
 
-This describes the change in total energy when atom :math:`\kappa` is displaced in direction :math:`\alpha` and atom :math:`\kappa\prime` is displaced in direction :math:`\alpha\prime`.
+This describes the change in total energy when atom :math:`\kappa` in unit cell :math:`0`` is displaced in direction :math:`\alpha` and atom :math:`\kappa\prime` in unit cell :math:`l` is displaced in direction :math:`\alpha\prime`.
 In Euphonic the force constants are stored in 'compact' form, which means that the minimum required information is stored.
```
ajjackson, line 27: Not sure this is universally true; the information could be more compact if we used symmetry operations. The following sentence is also a bit long and hard to follow. We can probably explain this better!

Perhaps replacing both sentences with something like:

> This "compact" form of the force constants can be expanded as needed to a "full" 3Nx3N matrix of supercell atom pairs by mapping to an equivalent vector from unit cell 0.

rebeccafair: Done!

doc/source/force-constants.rst
```diff
@@ -24,14 +24,12 @@ The :py:attr:`ForceConstants.force_constants <euphonic.force_constants.ForceCons
   \frac{\partial^{2}E}{{\partial}u_{\kappa\alpha0}{\partial}u_{{\kappa}'{\alpha}'l}}
 
 This describes the change in total energy when atom :math:`\kappa` in unit cell :math:`0`` is displaced in direction :math:`\alpha` and atom :math:`\kappa\prime` in unit cell :math:`l` is displaced in direction :math:`\alpha\prime`.
-In Euphonic the force constants are stored in 'compact' form, which means that the minimum required information is stored.
-Using equivalent vectors, the force constants between every atom in the unit cell, and every atom in the supercell are all that is needed,
-so only :math:`N(3n)^2` values need to be stored rather than :math:`(3Nn)^2` as in the 'full' square force constants matrix,
-where N is the number of unit cells in the supercell, and n is the number of atoms in the unit cell.
+In Euphonic the force constants are stored in 'compact' form, with shape :math:`3Nn\times3n`, where :math:`N` is the number of cells in the supercell, and :math:`n` is the number of atoms in the unit cell.
```
ajjackson, line 27: This description is contradicted by the following paragraph.

euphonic/qpoint_phonon_modes.py
```diff
@@ -339,9 +339,9 @@ def calculate_debye_waller(
 
         .. math::
 
-          \\textbf{W}_{\\kappa} =
+          W^{\\alpha\\beta}_{\\kappa} =
           \\frac{\\hbar}{4M_{\\kappa}\\sum\\limits_{\mathbf{q}}{weight_\mathbf{q}}}
-          \\sum\\limits_{\mathbf{q}\\nu \in{BZ}}weight_\mathbf{q}\\frac{\\mathbf{e}_{\mathbf{q}\\nu\\kappa}\\mathbf{e}^{*}_{\mathbf{q}\\nu\\kappa}}
+          \\sum\\limits_{\mathbf{q}\\nu \in{BZ}}weight_\mathbf{q}\\frac{e_{\mathbf{q}\\nu\\kappa\\alpha}e^{*}_{\mathbf{q}\\nu\\kappa\\beta}}
           {\\omega_{\mathbf{q}\\nu}}
           coth(\\frac{\\hbar\\omega_{\mathbf{q}\\nu}}{2k_BT})
```
ajjackson, line 346: Could the `coth` be `\mathrm{coth}`? Trig functions are usually typeset upright.

I'm not wild about _weight_ but it is difficult to suggest a non-confusing alternative.

ajjackson (approved):
It all seems to render ok. As mentioned there seems to be a bit of contradiction/redundancy in the Force Constants section. That's not new, can approve these changes without touching it. But worth a look?

doc/source/force-constants.rst
rebeccafair: You mean about the shape of the matrix?

ajjackson: Right. Maybe instead of "with shape 3Nnx3n" it should be something like "with N(3n^2) elements"? Then we build anticipation for the next paragraph telling you the actual shape!

rebeccafair: Good idea, done!

euphonic/qpoint_phonon_modes.py
rebeccafair: Done!

euphonic/qpoint_phonon_modes.py
```diff
@@ -340,10 +340,10 @@ def calculate_debye_waller(
         .. math::
 
           W^{\\alpha\\beta}_{\\kappa} =
-          \\frac{\\hbar}{4M_{\\kappa}\\sum\\limits_{\mathbf{q}}{weight_\mathbf{q}}}
-          \\sum\\limits_{\mathbf{q}\\nu \in{BZ}}weight_\mathbf{q}\\frac{e_{\mathbf{q}\\nu\\kappa\\alpha}e^{*}_{\mathbf{q}\\nu\\kappa\\beta}}
+          \\frac{\\hbar}{4M_{\\kappa}\\sum\\limits_{\mathbf{q}}{\\mathrm{weight}_\mathbf{q}}}
+          \\sum\\limits_{\mathbf{q}\\nu \in{BZ}}\\mathrm{weight}_\mathbf{q}\\frac{e_{\mathbf{q}\\nu\\kappa\\alpha}e^{*}_{\mathbf{q}\\nu\\kappa\\beta}}
           {\\omega_{\mathbf{q}\\nu}}
-          coth(\\frac{\\hbar\\omega_{\mathbf{q}\\nu}}{2k_BT})
+          \\mathrm{coth}(\\frac{\\hbar\\omega_{\mathbf{q}\\nu}}{2k_BT})
```
ajjackson, line 346: Looks good! The text below is now a bit out of sync: need to update epsilon -> e [is that _really_ a good idea? We haven't update Debye-Waller to "exp" if we are going that way to match the paper], bold q in ω<sub>**q**ν</sub>, upright "weight" with bold q.

The coth section would look better (and match paper) with big parens:
```
\\mathrm{coth}\\left(\\frac{\\hbar\\omega_{\mathbf{q}\\nu}}{2k_BT}\\right)
```

rebeccafair:
Good spots above, I think I've updated almost everything now to match the paper (with the exception that I don't use the prime notation in the DW equation, I think in the context of the software its obvious that the q-points in this equation are not the same as the structure factor equation).

I'm also noting here that in the commits https://github.com/pace-neutrons/Euphonic/compare/e3b1b53...6df4248 I have changed the CI behaviour so it will only run on branches containing `master`, `release`, or branches that specifically have `run_ci` in their name. CI can also be disabled for certain PRs by adding the `no_ci` label (e.g. for doc only updates like this one).

ajjackson:
Looks good to me, happy for merge

PR #228, opened 2022-07-06
Update Python test/release versions

rebeccafair:
The upcoming `0.6.5` release of Euphonic will support Python versions `3.6-3.10`, this PR updates the CI and release scripts for those versions.

Regularly testing on Python 3.6, 3.7, 3.8, 3.9 and 3.10 is overkill and will make the CI take ages, so the tests for 3.8 and 3.9 will generally be skipped. If we run the CI explicitly via a workflow dispatch, tests for all versions will be run. We can do this just before a release to ensure everything works on all versions.

### DIFF
--- a/.github/workflows/build_upload_pypi_wheels.yml
+++ b/.github/workflows/build_upload_pypi_wheels.yml
@@ -8,11 +8,11 @@ jobs:
   build:
     strategy:
       matrix:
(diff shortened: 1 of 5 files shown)


ajjackson (approved):
It looks like the tests are running and skipping as expected.

We can't see here if "synchronize" is working correctly, but we can try it out for release soon and if it fails then we can fix it then.

PR #229, opened 2022-07-07
Fix PyPI wheel builds

rebeccafair:
Ready for review provided the PyPI builds in https://github.com/pace-neutrons/Euphonic/actions/runs/2630595376 pass (the Conda ones will probably still fail). Summary:

- Move from `RalfG/python-wheels-manylinux-build@v0.3.4-manylinux2010_x86_64` to `pypa/cibuildwheel@v2.8.0` action to create manylinux wheels. Originally the manylinux builds where failing due to the issue described in https://github.com/pypa/manylinux issue 1309, meaning that an unversioned wheel was being created from verisoneer.  `cibuildwheel` is better maintained so has already caught and fixed this issue, and hopefully will do with any future issues.
- Added a workflow dispatch option which wont try and upload the wheels to PyPI, so we can create wheels without having to create a release.
- Switch from the `setup-python` Github action to `setup-miniconda`, as `setup-python` doesn't have Python 3.6
- Use strings for version numbers, otherwise `3.10` is read as `3.1`
- Upgrade to checkout `v3`, might as well?
- Install llvm on Mac for clang compiler to compile Euphonic
- Fix typo in `release_tox`

### DIFF
--- a/.github/workflows/build_upload_pypi_wheels.yml
+++ b/.github/workflows/build_upload_pypi_wheels.yml
@@ -42,6 +42,8 @@ jobs:
         with:
           ref: v0.6.5
           fetch-depth: 0 # Ensure tags are fetched for versioning
+      - name: Fix git ownership for manylinux builds, see https://github.com/pypa/manylinux/issues/1309
+        run: git config --global --add safe.directory "*"
       - name: Setup Python ${{ matrix.python-version }} with Conda
         uses: conda-incubator/setup-miniconda@v2
         with:
--- a/.github/workflows/build_upload_pypi_wheels.yml
+++ b/.github/workflows/build_upload_pypi_wheels.yml
@@ -47,21 +47,21 @@ jobs:
         with:
           python-version: ${{ matrix.python-version }}
       - name: Update Python pip, wheel, and twine
+        shell: bash -l {0}
         run: |
           python -m pip install --upgrade pip wheel twine
-        shell: bash
       - name: Install llvm on Macos
         if: startsWith(matrix.os, 'macos')
         run: brew install llvm
       - name: Build Python wheel
         if: matrix.os != 'ubuntu-latest'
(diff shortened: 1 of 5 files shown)


ajjackson:
> Ready for review provided the PyPI builds in https://github.com/pace-neutrons/Euphonic/actions/runs/2630595376 pass (the Conda ones will probably still fail).

```
___________________________________ summary ____________________________________
  pypi-py36: commands succeeded
  pypi-py37: commands succeeded
  pypi-py38: commands succeeded
  pypi-py39: commands succeeded
  pypi-py310: commands succeeded
ERROR:   conda-py37: commands failed
ERROR:   conda-py38: commands failed
ERROR:   conda-py39: commands failed
ERROR:   conda-py310: commands failed
  pypisource-py36: commands succeeded
ERROR:   pypisource-py310: commands failed
Error: Process completed with exit code 1.
```

Well, you called that correctly! But I don't follow what's going on here? Why are we requesting Conda builds that fail, and why does that action fail while this PR passes?

rebeccafair:
> Well, you called that correctly! But I don't follow what's going on here? Why are we requesting Conda builds that fail, and why does that action fail while this PR passes?

So that job is run manually by a workflow dispatch where you input the version of Euphonic to install and test, and it just runs all the tests in the [`release_tox.ini`](https://github.com/pace-neutrons/Euphonic/blob/3516167ffa6eeda7345d9bdcba815d248f5f7d57/release_tox.ini) file, which includes the downloading and installing from both PyPI and Conda, the Conda 0.6.5 release isn't available yet which is why those fail (the install from PyPI source with 3.10 also fails but that's due to a typo in the `release_tox.ini` file which I've now fixed).

This PR passes because it's just running the regular Euphonic tests as usual, and is not linked to the workflow dispatch release test above.

ajjackson (approved):
Cool let's try it

PR #231, opened 2022-07-11
Merge v1.0 release branch into master

rebeccafair:
Checking nothing has been broken...


ajjackson:
Looking at the code coverage report:

Coverage lost:
- `euphonic.util._deprecation_warn()` I guess this is actually good news :sweat_smile: 

Looks like there are some things in `euphonic.cli.utils` too but the coverage website is currently complaining about too many requests? Anyway, I doubt it is anything to worry about.

I think we're fairly happy that v1 is imminent, so ok to merge.

rebeccafair:
> Coverage lost:
> 
> * `euphonic.util._deprecation_warn()` I guess this is actually good news 😅

Good spot!

> Looks like there are some things in `euphonic.cli.utils` too

Managed to have a look and we've only lost coverage on the `deprecated_arg` function, as expected!

PR #238, opened 2022-07-22
General cleanup

rebeccafair:
* Cleans up deprecations/warnings that arise when running tests
* Remove the performance benchmarking directory, the performance tests haven't been running in over a year, and it would take a bit of maintenance to get them going again. We can recover them in the future if we decide we want to run regular performance tests again.


rebeccafair:
CI has failed because Codacy coverage failed to upload, but all the actual tests have passed.

doc/source/cite.rst
```diff
@@ -12,9 +12,9 @@ or it can be read programatically as follows:
 
   import yaml
   import euphonic
-  from importlib_resources import open_text
+  from importlib_resources import files
 
-  with open_text(euphonic, 'CITATION.cff') as fp:
+  with open(files(euphonic).joinpath('CITATION.cff')) as fp:
```
ajjackson, line 17: Could the `joinpath()` be a `/`? It looks like `files()` returns something Pathlib-ish.

euphonic/force_constants.py
```diff
@@ -296,7 +296,7 @@ def calculate_qpoint_phonon_modes(
 
         .. math::
 
-          \sum_{\\kappa^\\prime \\alpha^\\prime}D_{\\alpha {\\alpha}^\\prime}^{\\kappa {\\kappa}^\\prime}(\\mathbf{q}) {e}_{\\mathbf{q}\\nu{\\kappa^\\prime}{\\alpha^\\prime}} =
+          \\sum_{\\kappa^\\prime \\alpha^\\prime}D_{\\alpha {\\alpha}^\\prime}^{\\kappa {\\kappa}^\\prime}(\\mathbf{q}) {e}_{\\mathbf{q}\\nu{\\kappa^\\prime}{\\alpha^\\prime}} =
```
ajjackson, line 299: Are you sure these double-escapes are needed? I think `.. math::` blocks automatically handle them. In the RST documentation the examples use single backslashes. https://docutils.sourceforge.io/docs/ref/rst/mathematics.html

Was there a rendering problem under some conditions?

tests_and_analysis/test/euphonic_test/test_qpoint_frequencies.py
```diff
@@ -528,7 +529,7 @@ def test_calculate_dos_with_0_inv_cm_bin_doesnt_raise_runtime_warn(self):
         qpt_freqs = get_qpt_freqs(
             'quartz', 'quartz_666_qpoint_frequencies.json')
         ebins = np.arange(0, 1300, 4)*ureg('1/cm')
-        with pytest.warns(None) as warn_record:
+        with warnings.catch_warnings(record=True) as warn_record:
```
ajjackson, line 532: Pytest docs suggest an alternative way of promoting warnings to errors: https://docs.pytest.org/en/7.1.x/how-to/capture-warnings.html#additional-use-cases-of-warnings-in-tests

```python
with warnings.catch_warnings():
    warnings.simplefilter("error")
    ...
```

Not clear if our way of doing it gives better information, probably not. (I think it will just tell us the number of warning but not what they were?)

ajjackson (commented):
A few queries

doc/source/cite.rst
rebeccafair: The [importlib_resources docs](https://importlib-resources.readthedocs.io/en/latest/using.html#packages-or-package-names) uses `joinpath` in their example, `/` works and is shorter (although as someone not used to that syntax its a little harder to read for me), but I guess there's no reason not to use it.

euphonic/force_constants.py
rebeccafair: There were no specific rendering issues, but when compiling the docs locally I was getting escape warnings e.g. `\s not recognised as an escape sequence` and if there are a lot of them that could hide other useful warnings

tests_and_analysis/test/euphonic_test/test_qpoint_frequencies.py
rebeccafair: Yeah, there seems to be a bit of a disagreement on what is the best way to do things, this replacement is the most like-for-like replacement, even though it's not what's listed in the docs. See https://github.com/pytest-dev/pytest issue 9404 and its first comment, this is also the drop in replacement implemented by Scipy, see https://github.com/scipy/scipy PR 15192. (not linking directly because I don't want to Spam their issues/PRs).

I guess where we assert that there is 1 warning we could also check the type of warning

doc/source/cite.rst
ajjackson: I'm not entirely used to `/` for paths either, but I'd quite like to _get_ used to it :smile:

euphonic/force_constants.py
ajjackson: That's weird, on Master branch with Linux and sphinx-build 3.5.4 I don't get any warning related to this. (It mostly complains about underline lengths in the changelog!)

tests_and_analysis/test/euphonic_test/test_qpoint_frequencies.py
ajjackson: With `pytest.warns`, as well as providing a warning subtype you can provide a pattern to match in the warning message. So where we _require_  a warning that's the tool for the job.

euphonic/force_constants.py
rebeccafair: Oh wait my mistake, the warnings are actually in the test outputs! When running `pytest -v tests_and_analysis\test\euphonic_test\test_force_constants.py` (Python 3.7 on Windows)

tests_and_analysis/test/euphonic_test/test_qpoint_frequencies.py
rebeccafair: Good point, I don't know why those were written with `None` in the first place...

euphonic/force_constants.py
ajjackson: I don't see it in the Windows CI logs though :thinking:

ajjackson: Adding to the confusion, multi-line strings seem to do some automatic escaping anyway:

```
In [100]: """ok there
     ...: is a break in
     ...: here. Or two. \s pls no complain"""
Out[100]: 'ok there\nis a break in\nhere. Or two. \\s pls no complain'
```

So it's not clear how pytest would ever see an invalid string to complain about.

doc/source/cite.rst
```diff
@@ -14,7 +14,7 @@ or it can be read programatically as follows:
   import euphonic
   from importlib_resources import files
 
-  with open(files(euphonic).joinpath('CITATION.cff')) as fp:
+  with open(files(euphonic)/'CITATION.cff') as fp:
```
ajjackson, line 17: I would find these more readable with whitespace around them; makes it a bit clearer an operation is going on.

euphonic/force_constants.py
rebeccafair: Ok looking at this https://stackoverflow.com/questions/46543194/does-r-stand-for-something-in-sphinx it looks like we have a few options:
* Leave everything double escaped (easiest option, a bit ugly but otherwise no harm done)
* Single escape everything, double escape things like `\r` (and anything else we find)
* Single escape everything, use raw string e.g. `r'this is a raw string'`. This seems to work but I'm not sure if it will give us issues elsewhere.

ajjackson: Raw string seems the nicest solution from the source code end. But I'm a bit confused as to why this would be system-specific, or if it is needed at all.

Maybe in some environments `\s` is interpreted as "space" and that inconsistency causes it to skip a stage of automatic double-escape?

I suppose we can keep the double-escapes in the PR for now, and migrate slowly to raw strings as things are added or maintained.

ajjackson (approved):

Issue #191, opened 2021-09-08
Investigate segfault on MacOS test runners


rebeccafair:
In very specific cases the `euphonic-dispersion`, `euphonic-intensity-map` and `euphonic-powder-map` tests have been found to fail on the Github actions MacOS runners. This was discovered in #189 , details in a94e159d06af0

Some commands that may cause a fault:
* `euphonic-intensity-map tests_and_analysis/test/data/castep_files/graphite/graphite.castep_bin -s test.png -w coherent --temperature 1000 --use-c` - with Python 3.6 this only segfaults when run via pytest on the MacOS runners
* `euphonic-powder-map tests_and_analysis/test/data/phonopy_files/NaCl_prim/phonopy_nacl.yaml -s test.png --npts 10 --npts-min 10 --q-spacing 1 -w coherent --temperature 1000 --use-c`

rebeccafair:
Looks like you can avoid the segfaults by using an MKL-linked Numpy (available via the default Conda channel). Apparently you can also [change the backend](https://conda-forge.org/docs/maintainer/knowledge_base.html#switching-blas-implementation) of the Numpy installed via the conda-forge channel to use MKL but I haven't been able to get this to work myself yet. What backend libraries are used with the Numpy in Mantid @ajjackson ?

ajjackson:
From a recent conda build on OSX:

```python
np.show_config()
blas_info:
    libraries = ['cblas', 'blas', 'cblas', 'blas']
    library_dirs = ['/Users/adam/mambaforge/envs/mantid-developer/lib']
    include_dirs = ['/Users/adam/mambaforge/envs/mantid-developer/include']
    language = c
    define_macros = [('HAVE_CBLAS', None)]
blas_opt_info:
    define_macros = [('NO_ATLAS_INFO', 1), ('HAVE_CBLAS', None)]
    libraries = ['cblas', 'blas', 'cblas', 'blas']
    library_dirs = ['/Users/adam/mambaforge/envs/mantid-developer/lib']
    include_dirs = ['/Users/adam/mambaforge/envs/mantid-developer/include']
    language = c
lapack_info:
    libraries = ['lapack', 'blas', 'lapack', 'blas']
    library_dirs = ['/Users/adam/mambaforge/envs/mantid-developer/lib']
    language = f77
lapack_opt_info:
    libraries = ['lapack', 'blas', 'lapack', 'blas', 'cblas', 'blas', 'cblas', 'blas']
[...]

PR #242, opened 2022-09-26
Add workaround for segfault with openblas

rebeccafair:
Temporary workaround for #191 

Should be merged after #243

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,6 +1,10 @@
 `Unreleased <https://github.com/pace-neutrons/Euphonic/compare/v1.0.0...HEAD>`_
 ----------
 
+- Bug fixes:
+
+  - Avoid occasional segmentation faults when using OpenBLAS, workaround for #191
+
 `v1.0.0 <https://github.com/pace-neutrons/Euphonic/compare/v0.6.5...v1.0.0>`_
 ----------
 
--- a/euphonic/force_constants.py
+++ b/euphonic/force_constants.py
@@ -605,13 +605,11 @@ class ForceConstants:
 
         rfreqs = np.zeros((n_rqpts, 3*n_atoms))
         if return_eigenvectors:
-            reigenvecs = np.zeros(
-                (n_rqpts, 3*n_atoms, n_atoms, 3), dtype=np.complex128)
+            n_reigenvecs = n_rqpts
         else:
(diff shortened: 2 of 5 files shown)


ajjackson (commented):
This looks like magic, let me see if I understand correctly:

`reigenvecs` is a (n_qpts, 3N, N, 3)` array where the q-points are indexed over a set of q-points that had duplicates removed.

This is created as an empty array (actually zeros?) and passed to the C extension as a destination to write the output. (In practice some in-place mutation will also happen.)

Only n_qpts of eigenvectors will actually be calculated, written to this array, and read back out. But for mysterious openblas reasons, we also need to have an extra row of zeros in this array. Making the array larger than the results to be iterated inside it? [Does this row actually get changed?]

The bug is known to be related to array size, but despite this we are confident that for any value of n_qpts, incrementing the size by one will avoid the segfault.

It's not obvious to me that adding this row wouldn't be equally likely to take us _to_ one of the "bad numbers", have we checked that by running one less q-point for a case where usually this fixes a segfault?

rebeccafair:
Yep your summary is correct, and it is sort of like magic...

> [Does this row actually get changed?]

I've checked the usual segfaulting 8419 q-points with 1 atom/unit cell (which with this change creates a 8420-length array) and the final row doesn't seem to get changed mysteriously, they all stay as zeros. I would've expected if there was some kind of overflow it would get changed, but it does seem to stop the fault at least...

> It's not obvious to me that adding this row wouldn't be equally likely to take us _to_ one of the "bad numbers", have we checked that by running one less q-point for a case where usually this fixes a segfault?

Yes this is a worry. I have checked 1 fewer q-point, and have been running a selection of q-points/natoms on my old laptop in the background and haven't seen it segfault yet, so it seems to at least not be any worse than what we have at the moment?

ajjackson:
So our hypothesis is that occasionally a dodgy openblas build will slightly overreach its indexing range while trying to claim memory; it doesn't actually use it, but we can protect ourselves from the resulting segfault with a bit of padding.

And in cases experienced _so far_ one row has been sufficient padding; if we _really_ understood the theoretical limit of padding required we'd be in a position to also fix the bug upstream!

Ok, seems to me that adding one row of padding is a reasonable safety measure and only costs a little bit of memory.

PR #243, opened 2022-09-26
Work around different tick labels with mpl 3.6.0

rebeccafair:
Temporary workaround for Matplotlib 3.6 which seems to add tick labels to invisible axes, rather than them being empty strings. Not sure if this intended behaviour and will be fixed soon or not, so just work around it for now. Visually the plots produced by the CL tools seem fine and despite having tick labels, the axes are still invisible.

Should be merged before #242


ajjackson (commented):
As the fix is essentially "if a tick is supposed to be empty, don't check it", the key question is:

Are there any non-invisible axes where we expect to see empty ticks in the tests? e.g. a hidden y scale for arbitrary intensity. I don't think there are at the moment.

I'm worried that this test will fail to catch the scenario where the ticks are numbered and somehow became visible. What property is keeping them hidden while the axis label is still displayed?

rebeccafair:
Yeah, I agree that was a bit of a hack.

I haven't been able to figure out what property of the axis makes it invisible, or why the labels have suddenly changed, it looks like the invisible axes for all mpl versions still have `AutoLocator` as their tick locator, so maybe the behaviour of that has changed? I can't find anything in the release notes though.

One thing we could do is explicitly set the xticks on the invisible axis to zero when we create it? `ax = fig.add_subplot(111, frameon=False, xticks=[])`? I don't like changing the behaviour of things just to pass tests, but I guess that's what we have to do when we're testing the plot details rather than using some kind of image based testing...

rebeccafair:
Ok, I think I've improved it, we now only read and check tick labels from visible axes (I have removed invisible axes ticks from the .json files) and we also now only check tick labels for `dispersion`, `intensity-map` as this is where we've set ticks explicitly, we no longer check `dos` or `powder-map` because we don't explicitly set them so can't ensure the `matplotlib` set defaults won't change between verisons.

tests_and_analysis/test/script_tests/utils.py
```diff
@@ -46,15 +46,18 @@ def get_fig_label_data(fig) -> Dict[str, Union[str, List[str]]]:
...
                   'title': fig._suptitle.get_text()}
 
-    # Get axis labels from whichever axis has them; collect all tick labels
+    # Get axis labels from whichever axis has them.
+    # Collect all tick labels from visible axes only,
+    # we don't care about invisible axis labels
     for ax in fig.axes:
-        ax_label_data = get_ax_label_data(ax)
+        if ax.get_frame_on():
+            ax_label_data = get_ax_label_data(ax)
```
ajjackson, line 54: Would it make sense to add something like:
```python
else:
    # Invisible axes should not have visible ticks or labels
    assert not any([tickline.get_visible() for tickline in ax.get_xticklines()])
    assert not any([tickline.get_visible() for tickline in ax.get_yticklines()])
    assert all(np.array([t.get_color() for t in ax.get_xticklabels()]) == 'none')
    assert all(np.array([t.get_color() for t in ax.get_yticklabels()]) == 'none')
```
?

rebeccafair: I guess there's no harm in it!

rebeccafair: I had to change it a bit to avoid line breaks but done :)

rebeccafair: Oh no, seems like `ax.get_x/yticklines().get_visible()` returns `True` for old versions of `matplotlib`, even though you can't see them... maybe we can skip this check?

ajjackson: Argh! Ok, a bit more experimentation suggests a cleaner overall strategy. If in the plotter we have:

```python
for ax_dir in ax.xaxis, ax.yaxis:
    ax_dir.set_major_locator(plt.NullLocator())
```

then there shouldn't _be_ any ticks at all on the invisible axis. Then instead of looking at properties, we can just

```python
for ticklines in ax.get_xticklines(), ax.get_yticklines():
    assert len(ticklines) == 0
```

and be a _lot_ more confident that they won't find some crazy way to gain a label and/or become visible.

rebeccafair: Ok, so if we use `NullLocator` for the y axis (or set `yticks=[]` in `add_subplot`), we get this weird result where the y label overlaps the ticks: 
image

I think that's because the `ylabel` is attached to the invisible axis, so if there are no tick labels on the invisible axis it puts the label close to the axis as if the visible ticks weren't there. I think we have 2 options:

* Set x-axis locator only to `NullLocator`, don't check y-axis ticks
* Set x and y-axis locators to `NullLocator`, set y-label on one of the visible axes instead (this should be fine as they both have `sharey=True`

rebeccafair: I've tried the second one in [9259f49](https://github.com/pace-neutrons/Euphonic/pull/243/commits/9259f49f9af6e4bc35354ea6adc782774d188758), fingers crossed...

rebeccafair: Wait, of course this would mess up the xlabels too, argggggh!

image

rebeccafair: > I suppose, then, that our best option is to
> 
> * not check for invisibility
> * wait for a user report and fix the problem when it occurs. Perhaps by then we can drop some MPL versions and check more consistently.

I think so, it's been a good lesson of don't touch plotting unless you really have to...

ajjackson (approved):

PR #245, opened 2022-10-18
Instrumental resolution broadening

ajjackson:
This implements energy-dependent resolution broadening with 1-D polynomial functions.

The idea is that while Euphonic does not contain detailed instrument models, users can obtain a polynomial fit to energy-dependent broadening (derived empirically or calculated using something like PyChop or McStas) and provide the polynomial coefficients instead of a single broadening "width". We could provide information in tutorials to help with this without overcomplicating the Euphonic library.

It re-uses the interpolated broadening machinery used for adaptive DOS broadening. As this resolution broadening is energy-dependent, it would normally be performed after other calculation and binning steps.

There are some obvious places for further improvement and consideration:

- Access from CLI (as above, I like the idea that we simply use the existing width parameter and allow it to be passed multiple values)
- Optimisation for use with multiple data series: currently the approximate broadening only works on 1D data. We can make a simple loop to fill out Spectrum1DCollection or Spectrum2D, but this would involve a bit of redundant recalculation of kernels. There is also some redundant histogram-binning overhead because the inner function doesn't know that the data is already binned.
- Lorentzian functions?
- Exact broadening as a reference for interpolated method?

But for now we should discuss the API. How much of this stuff should live in `euphonic.broadening` on `Spectrum` classes or elsewhere?

I also note a convention difference between `euphonic.broadening` which works with "width" in _sigma_ (standard deviation) and the outer functions which treat width as FWHM. It's quite possible users will have a mixture of both and the conversion is simple, but the internals could be more clear and consistent about this.


euphonic/broadening.py
```diff
@@ -1,14 +1,67 @@
...
+    bins
+        Energy bins for output spectrum
+    x
+        Data positions (to be binned)
+    width_polynomial
+        Coefficients of polynomial fit to energy/width relationship. Order
+        should be consistent with numpy.polyfit / numpy.polyval (i.e. from
+        largest exponent to smallest.) Width is expressed as standard deviation
+        (not FHWM); the conversion factor between these conventions is
+        :math:`FWHM = \sqrt{8 \ln 2} \sigma`
```
rebeccafair, line 38: It's too early for this kind of detail but noting it now so I don't forget, the Numpy docs now recommend the use of [numpy.polynomial](https://numpy.org/doc/stable/reference/routines.polynomials.package.html#module-numpy.polynomial), which unfortunately has the opposite convention with coefficients from [lowest to highest](https://numpy.org/doc/stable/reference/generated/numpy.polynomial.polynomial.Polynomial.html#numpy.polynomial.polynomial.Polynomial)...

rebeccafair:
> * Access from CLI (as above, I like the idea that we simply use the existing width parameter and allow it to be passed multiple values)

I also like this idea. It might be nice if we could reuse `Spectrum1D.broaden` and allow `width` to be both a `Quantity` or `List[Quantity]`. But we have to be careful this doesn't become an everything argument, are there other types of broadening we might want to add in the future?

Currently the functions take the widths and their units separately, which is nicer than having to input redundant `[1*ureg('meV'), 2*ureg('meV')]` etc., but it does clash with convention in the rest of Euphonic. I guess we could accept both a plain float (in which case assume units) and Quantity, or maybe just require Quantities for now and worry about this later.

> * Optimisation for use with multiple data series: currently the approximate broadening only works on 1D data. We can make a simple loop to fill out Spectrum1DCollection or Spectrum2D, but this would involve a bit of redundant recalculation of kernels. There is also some redundant histogram-binning overhead because the inner function doesn't know that the data is already binned.
> * Lorentzian functions?

I guess initially we could just implement the `Spectrum1DCollection` in a loop and do some quick profiling to see if it is an issue? 

With `Spectrum2D` we might need some more alterations to be correct, because really we should have a 2D kernel with appropriate widths in each dimension and convolve with that rather than doing a 1D convolution in each dimension (I think?). The Lorentzian broadening in Euphonic does 2 x 1D convolutions instead of a 2D convolution so I think its actually not completely correct.

Ignoring the 2D issue for now, we could probably add 1D Lorentzian kernels in `_width_interpolated_broadening` without too much trouble? But for now we could just error if a user tries to use polynomial and a Lorentzian shape.

> But for now we should discuss the API. How much of this stuff should live in `euphonic.broadening` on `Spectrum` classes or elsewhere?

I think `euphonic.broadening`  seems like the right place for `broaden_with_polynomial`. We could call that functionality from `Spectrum1D.broaden` if we accept a list in addition to a scalar width, provided that we don't think this argument isn't going to blow up and try to do too many things in the future

> I also note a convention difference between `euphonic.broadening` which works with "width" in _sigma_ (standard deviation) and the outer functions which treat width as FWHM. It's quite possible users will have a mixture of both and the conversion is simple, but the internals could be more clear and consistent about this.

Good point! I like the use of `width_convention`, if it's something that's too much for this PR we could add it to the `broaden` functions later?

mducle:
> > *  Access from CLI (as above, I like the idea that we simply use the existing width parameter and allow it to be passed multiple values)
>
> I also like this idea. It might be nice if we could reuse Spectrum1D.broaden and allow width to be both a Quantity or List[Quantity]. But we have to be careful this doesn't become an everything argument, are there other types of broadening we might want to add in the future?

Since in principle for neutron spectrometer resolutions the broadening width is a function of **Q** and *ω*, the most general width argument we can specify would be a function or lambda which takes `Qh`, `Qk`, `Ql`, and `E` as input and returns the width at that point. [Horace does this](https://github.com/pace-neutrons/Horace/blob/master/horace_core/disp2sqw.m#L61-L69) but I've never used this functionality, so I'm not sure if we should design for it - maybe the best thing is to restrict `width` here to be a scalar or list and to have another argument (e.g. `width_function`) for something more general if there is a demand for it.

One thing about the polynomial interpolation - should we consider how to specify the range/limits in which the interpolation is valid. The polynomial would only be an approximation to the resolution function within some range; but if you have e.g. modes which are very very high in energy, it might be outside the range of validity of the polynomial - can we assume that users would check this themselves? Or should we have _another_ parameter, e.g. `width_poly_limits` which would raise an error or warning if there are modes outside this...

ajjackson:
> One thing about the polynomial interpolation - should we consider how to specify the range/limits in which the interpolation is valid. 

Internally we could pass around [that new Polynomial object](https://numpy.org/doc/stable/reference/generated/numpy.polynomial.polynomial.Polynomial.html#numpy-polynomial-polynomial-polynomial) instead of the coefficient list. This has optional _domain_ and _window_ attributes which could correspond to those input range limits.

Already I find it necessary to apply a floor to output values that get too small, otherwise we have sub-bin peak widths that create increasingly tall peaks without a perceivable difference in width.

Arguably it's on the user to choose an appropriate energy range, anyway...

ajjackson:
> Since in principle for neutron spectrometer resolutions the broadening width is a function of Q and ω, the most general width argument we can specify would be a function or lambda which takes Qh, Qk, Ql, and E as input and returns the width at that point.

Even more general would be a 4x4 covariance matrix, but we aren't ready for that yet :sweat_smile:  We should also consider whether that sort of thing would live in Euphonic or another code... But anyway here we are broadening a pre-calculated spectrum, and Euphonic spectra only go up to 2D. We don't have that 4-D input coordinate at the point this is applied; such a routine would need to be part of the StructureFactor methods to calculate 1d averages and S(|**q**|,ω) maps.

> We could call that functionality from Spectrum1D.broaden if we accept a list in addition to a scalar width, provided that we don't think this argument isn't going to blow up and try to do too many things in the future

Right, I do worry that adding more and more options to `broaden()` will become unsustainable. Allowing the existing width argument to take a list, or even a function, is not too bad. Would we also need options to clarify which axes the function refers to? What about `width_unit`, `width_lower_limit`, `width_convention`, `adaptive_error`? I'm not fond of multi-mode functions with arguments that do nothing in most cases.

Re: `width_unit`, the reason it exists is that in a polynomial not all the coefficients have the same unit. E.g. the TOSCA resolution polynomial is  σ = 1e-7 cm<sup>-1</sup> E<sup>2</sup>/cm<sup>-2</sup> + 5e-3 E cm<sup>-1</sup>/cm<sup>-1</sup> + 2.5 cm<sup>-1</sup>. So it can't be stored correctly as a Quantity! But maybe internally something like Tuple(Polynomial, Quantity) would work where Quantity is a scale factor containing the unit - by which the input is divided and the output is multiplied.

ajjackson:
A neat benefit of working with Polynomial is that in principle one can also use Chebyshev. But in practice these resolution functions are so smooth that it will make no difference...

rebeccafair:
> Re: `width_unit`, the reason it exists is that in a polynomial not all the coefficients have the same unit. E.g. the TOSCA resolution polynomial is σ = 1e-7 cm-1 E2/cm-2 + 5e-3 E cm-1/cm-1 + 2.5 cm-1. So it can't be stored correctly as a Quantity! But maybe internally something like Tuple(Polynomial, Quantity) would work where Quantity is a scale factor containing the unit - by which the input is divided and the output is multiplied.

Ah of course, I hadn't thought of that. In that case perhaps it is better to have `broaden_with_polynomial` as a separate function. Should the default `width_convention` be FWHM like broaden?

ajjackson:
Yeah, FWHM is more consistent with Euphonic in general. (And PyChop, actually...) It's just that the usual TOSCA function is in STD and that's what I've been testing with :sweat_smile:

ajjackson:
> > Re: `width_unit`, the reason it exists is that in a polynomial not all the coefficients have the same unit. E.g. the TOSCA resolution polynomial is σ = 1e-7 cm-1 E2/cm-2 + 5e-3 E cm-1/cm-1 + 2.5 cm-1. So it can't be stored correctly as a Quantity! But maybe internally something like Tuple(Polynomial, Quantity) would work where Quantity is a scale factor containing the unit - by which the input is divided and the output is multiplied.
> 
> Ah of course, I hadn't thought of that. In that case perhaps it is better to have `broaden_with_polynomial` as a separate function. Should the default `width_convention` be FWHM like broaden?

Well, the other benefit to having a Spectrum-specific broadening tool is that is can handle the bin-width issue: spectral data is scaled so the overall intensity is unaffected by bin-width, whereas the lower-level broadening methods are designed to bin from raw data (which makes sense for an adaptive DOS). It's a bit of a headache keeping the units straight so nice to abstract that away.

It could be a pure function instead of a method. Maybe start with pure functions, then implement in command-line tools, then after a while we can review and decide what would make most sense as a method?

ajjackson:
Another API aspect I'm not sure about is limiting the functional to a Polynomial object. Although the type-hint says that, really the only features that matter are:

- it's a callable that can take a dimensionless array x and return a corresponding array y
- you can multiply the callable by a scale factor and this will impact the results

The former property could be satisfied by any function handle. The latter requires a suitable object and is e.g. also satisfied by the Chebyshev polynomial object. (For typical smooth resolution functions, the least-square Polynomial and minimax Chebyshev approaches will give very similar results, but maybe in some cases the Chebyshev truncation workflow is preferred?)

The second property is also not really _needed_, it can be handled by scaling the output instead of the object. Multiplying the coefficients is more computationally efficient, but I don't suppose that avoiding one array multiplication is really a big deal.

It's easy to foresee the utility of allowing an external, more complex, resolution function to be passed as a handle. I suppose a balance between YAGNI and future-proofing would be have `width_polynomial` as a named argument, which could later become optional by offering an alternative `width_function`.

ajjackson:
An example of a more generic interface to form a discussion basis:

```python
spectrum1d.broaden(x_width: Union[Quantity, Callable[[Quantity], Quantity]],
                   **kwargs) -> Spectrum1D

spectrum2d.broaden(x_width: Union[Quantity, Callable[[Quantity], Quantity]],
                   y_width: Union[Quantity, Callable[[Quantity], Quantity]],
                   **kwargs) -> Spectrum2D
```
which type-dispatches to fixed or variable-width implementations depending on inputs and also passes on the corresponding **kwargs.

Pros:
- the Tuple is gone, things are well-defined w.r.t. Quantity again
- concise syntax for interactive use, notebooks etc.
- consistent, backward-compatible and covers 2D combinations of fixed/variable
- although it doesn't support 2x2 covariance functions, it also doesn't _imply that it would_. Such a feature could be feasibly added to the same API as `xy_width: Callable[[Quantity, Quantity], Quantity]` where the callable output is a dimensioned 2x2 array and the inputs are (x, y) coordinates.
- As the Callable doesn't have to be a Polynomial, one can easily e.g. wrap a call to PyChop:
  ```python
  from pychop.Instruments import Instrument
  def my_resolution(energies: Quantity) -> Quantity:
      instrument = Instrument('MAPS', chopper='A', freq=400)
      return instrument.getResolution(energies.to('meV').magnitude,
                                      Ei_in=1000.) * ureg('meV')
  maps_spectrum = spectrum.broaden(y_width=my_resolution)

  ```

Cons:
- The user or calling function gains responsibility to provide a Callable that deals with Quantities correctly. This isn't too bad in practice; the one we use for the command-line tools would be something like `(lambda x: Polynomial(x.to(poly_unit) * poly_unit)`.
- We can no longer make a rather pretty and efficient transformation between FWHM and STD conventions by multiplying the Polynomial coefficients.
- A bunch of approximate-broadening-related keyword arguments become available even if you are just doing fixed-width broadening.
- No way to specify widths that vary as a function the _other_ axis
  - this could be achieved by also accepting Callable[[Quantity, Quantity], Quantity]. We would need a way to detect the difference between the two types of callable, but actually this is fairly easy: `len(inspect.signature(callable).parameters)`.
- What do we do about the existing `method` keyword? We're no closer to supporting irregular grids and it's not clear how this would interact with new modes.

mducle:
I think the proposed syntax looks good!

> * No way to specify widths that vary as a function the other axis 

Or you can add more keyword arguments - e.g. `x_width_[of|from]_xy` and `y_width_[of|from]_xy` but this might become a bit of a mouthful... (I'm not sure whether "of" or "from" makes more sense...

Or you can force users to use the `xy_width` keyword argument - this is essentially a diagonal covariance matrix right?

ajjackson:
> Or you can force users to use the `xy_width` keyword argument - this is essentially a diagonal covariance matrix right?

Right, if that _is_ available and people are trying to get that clever with their resolution functions then they might as well use this and load it with the desired off-diagonal terms.

rebeccafair:
I definitely prefer this new proposed interface, the more consistent use of `Quantity` makes me feel less uneasy and the PyChop example is very nice.

> Cons:
> 
> * The user or calling function gains responsibility to provide a Callable that deals with Quantities correctly. This isn't too bad in practice; the one we use for the command-line tools would be something like `(lambda x: Polynomial(x.to(poly_unit) * poly_unit)`.

Yeah, and as long as we document how to use it clearly, it shouldn't be too much trouble for users!

> * A bunch of approximate-broadening-related keyword arguments become available even if you are just doing fixed-width broadening.

I do worry about this, it's ok for now as we can just tell users to look at the docs for the `polynomial_broadening` function, but could get messy if we add more methods in the future. But I still think this way of doing this is better.

> * No way to specify widths that vary as a function the _other_ axis

`xy_width` sounds good to me!

> * What do we do about the existing `method` keyword? We're no closer to supporting irregular grids and it's not clear how this would interact with new modes.

I think it's still relevant, the 'fast approximate' polynomial broadening still uses a convolve in the end and we probably still want to emit the same warning for in the case of irregular grids, right? We could in the future evaluate a function for each bin (like in [calculate_dos](https://github.com/pace-neutrons/Euphonic/blob/master/euphonic/qpoint_frequencies.py#L202-L218) which would be slow, but at least more correct, and I don't think adding the polynomial broadening to `broaden` precludes this.

ajjackson:
Something I've noticed in `euphonic-dos` that needs fixing up:

there is a logical pathway such that
- if adaptive broadening and broadening width are both specified, the widths are multiplied together and applied
- if adaptive broadening and Lorentzian broadening are both specified, an error is raised because "Currently only Gaussian shape is supported with adaptive broadening"

There are two issues here:
- this isn't dimensionally consistent or correct. The width of the Gaussian function resulting from the convolution of two Gaussians is $\sigma_{1*2} =\sqrt{\sigma_1^2 + \sigma_2^2} $. Statistically this can be understood as _variances_ $\sigma^2$ being additive when normally-distributed errors are combined.
- It would never make sense to use a Lorentzian for adaptive DOS broadening as it has the wrong statistics for that job. However it is quite reasonable to apply Lorentzian broadening to represent some optical or scattering phenomenon on a distribution obtained by adaptive broadening. So when `shape='lorentz'` we should assume that the user means to apply Lorentzian broadening to the result of the (Gaussian) adaptive broadening. We aren't set up to interpolate the 2-D space of those function combinations, so this should be done in a separate step.

ajjackson:
Ah, looking at the help text I see that the existing behaviour is actually documented.

```
                eb_help = (
                    'If using fixed width broadening, the FWHM of broadening '
                    'on energy axis in ENERGY_UNIT (no broadening if '
                    'unspecified). If using adaptive broadening, this is a '
                    'scale factor multiplying the mode widths (no scale '
                    'factor applied if unspecified).')
```

I'm not sure that scaling the adaptive broadening is more useful than adding a fixed width to it, but I suppose that both are legitimate choices. Hmm, how best to handle this :thinking:

rebeccafair:
> Ah, looking at the help text I see that the existing behaviour is actually documented.
> 
> ```
>                 eb_help = (
>                     'If using fixed width broadening, the FWHM of broadening '
>                     'on energy axis in ENERGY_UNIT (no broadening if '
>                     'unspecified). If using adaptive broadening, this is a '
>                     'scale factor multiplying the mode widths (no scale '
>                     'factor applied if unspecified).')
> ```
> 
> I'm not sure that scaling the adaptive broadening is more useful than adding a fixed width to it, but I suppose that both are legitimate choices. Hmm, how best to handle this 🤔

I was just about to reply with this haha. The [conversion from mode gradients to widths](https://github.com/pace-neutrons/Euphonic/blob/master/euphonic/util.py#L266-L300) scales on the cell size and an estimate of the q-point spacing (which is based on the number of q-points due to symmetry weighting, I don't think this is completely correct but I think this is what CASTEP implements if I remember), which I wasn't sure was totally correct or would cover all cases so I wanted to make it easy to change this scaling so I put it in the `--eb` parameter. I didn't think of the case of someone doing adaptive broadening *then* doing some fixed with broadening after. Along the same line I didn't think of someone doing adaptive broadening *then* fixed width Lorentzian broadening.

ajjackson:
Yeah, when you think of the feature as "broadening" it seems weird to need two steps. But if you think of "improved DOS convergence" and "instrumental resolution" as different features it makes sense!

ajjackson:
I suppose the natural endpoint then would be to have separate `--adaptive-scale` and `--energy-broadening` arguments. But because `--energy-broadening` already exists, the path to transition is a bit ugly:

### Euphonic v1.x
`--adaptive-scale`: if using adaptive broadening, apply dimensionless scale factor
`--instrument-broadening`: fixed or energy-dependent broadening step, either integrated into adaptive Gaussian widths or as a separate step for Lorentzian.
`--energy-broadening`: Existing behaviour, dispatches to others depending on adaptive mode. Raises a warning recommending you be more specific and use one of the others.

### Euphonic v2.x
`--adaptive-scale`: as above
`--energy-broadening`: equivalent to v1.x `--instrument-broadening`
`--instrument-broadening`: aliased to `--energy-broadening`, raises DeprecationWarning

### Euphonic v3.x
`--adaptive-scale`: as above
`--energy-broadening`: as v2.x

rebeccafair:
This looks like a good path to me, or at least the best we're going to get!

ajjackson:
Ok, I'll crack on with that scheme. Is there somewhere we should be recording this kind of API
/deprecation roadmap?

rebeccafair:
Create some issues and slap a Euphonic 2.0/3.0 label on them? If we end up with lots we can think about organising it a bit more...

ajjackson:
Oddly, on my system the script_tests/test_dos `test_adaptive_and_lorentz_raises_value_error` hangs indefinitely rather than fail. Ah well, it seems to fail correctly on the CI and is about to come out because that combination is now supported.

ajjackson:
I notice there is some discrepancy between the Lorentzian broadening from the "direct" fixed-width implementation and the variable implementation. Nearly all of this can be accounted for by the truncation of the function in `spectra._distribution_1d`, which is currently set to 3*gamma. Lorentzians decay really slowly, I suggest pushing this up to something much larger.

lorentz-compare-extent-3

lorentz-compare-extent-16

The variable-width implementation sidesteps this problem by creating a kernel the size of the entire data range. This doesn't incur too much extra cost for the following FFT convolution, but could be overkill/inefficient for the direct implementation.

The remaining small difference can likely be attributed to convolution implementation details (e.g. boundary conditions).

ajjackson:
Oh wow, this function is even more hostile than I realised.

lorentz-log-log


Ok, I think our options are:
- go out to hundreds of FWHM to get the cutoff down to about 1e-6
- cut off a bit sooner and scale/shift the baseline so it hits zero neatly
- just use the whole x-axis

Tempted to take the last one, clearly Lorentzians can be cheap or approximate but not both

rebeccafair:
Oof, that is worse than I thought! I feel like we've seen this issue before with strange 'steps' in the 1D data at the edge of the kernel, but I haven't been able to find it in the issue tracker or elsewhere.

I agree using the whole energy range by default seems the most sensible.

ajjackson:
I had a little go at fitting a recursive filter and at low order the results don't look good enough. It might be possible at higher orders but even then extra work is needed to parametrise/interpolate for width. Would be an interesting project but for now I think we should just take the computational hit of exact direct evaluation. It's the best 'reference' method anyway.

index

freq_response

ajjackson:
I could see from that frequency response that a higher-order filter could be needed to get the required frequency cutoff. Out of curiosity I had another go and got some _much_ better IIR filters! For future reference, the tricks were:
- optimise in the poles/zeros domain and then convert to a stable cascaded-biquad filter
- numerical optimisation was very sensitive to starting values, needed to play with random seeds a bit

So it's not trouble-free and there's still the matter of parametrising/interpolating for width. But still, for future optimisation potential this is very encouraging.

filters

ajjackson:
Need to update unit-test Lorenzian broadened data for the new full-range calculation, as you can see the change seems pretty innocuous

comparison

ajjackson:
Before updating the DOS script test: here is the difference. It's quite significant, and clearly an improvement!

dos_diff

ajjackson:
On the other hand, this intensity difference in slices from 2-D looks wrong. Will investigate rather than blindly update the data... hurrah for unit tests!

map_diff

ajjackson:
Aha, I see the origin of the problem; the q-broadening width is very narrow. This leads to a greater-than-unity peak in the kernel, which would previously have been normalised away. Fundamentally this is a (very common) error in the way the kernel is constructed; the function is literally evaluated in each bin, rather than integrated across each bin to give an appropriate area. The trouble is that if we do it the "right" way it will look a bit different from a naive reference calculation!

*This problem is not restricted to Lorentzian functions*

Simple fix: numerically normalize the kernel. Arguably this just pushes the error into other bins, and will excessively flatten the tail. It is the approach used for the Scipy Gaussian broadening, so would be consistent - but the tail-flattening drawback seems more severe for a Lorentzian.

API-level fix: forbid very narrow broadening widths, disable broadening if width is set low enough to cause this problem

A marginally more expensive method that should produce formally correct histograms: evaluate
ql_8c86144a5fef9e38151114d19802f51a_l3
where *f* is the broadening function and *h* is the bin width.

I have an proof-of-concept which ends up being pretty cheap and concise, the main difference is that we want to work on an even-numbered input of bins rather than an odd-numbered set of bin-centers. What/where would be best to document this?

index

rebeccafair:
Very interesting! Thanks for looking into this so thoroughly. I would be tempted to go for direct evaluation + normalisation (and maybe warn on very low widths?) to be consistent with the Gaussian, but you're right about it still having issues. For small widths it seems to flatten the tail, but large widths (green line) it seems to overestimate the tail compared to the other methods. Would this be an issue for the widths used in most 'normal' use cases?

Maybe the more formally correct method would be best put in a broadening library (if the direct + normalisation method is sufficient for Euphonic for now)?

In terms of documenting this a short note in the docstring with some more info in https://github.com/pace-neutrons/pace-developers for now? If we end up with a broadening library the stuff in pace-developers could get moved to the library docs.

ajjackson:
With the normalised kernels the results have still changed a little bit because the range truncation was removed; this gives a larger normalisation constant and so slightly reduces the intensity.

map_diff

I think this is ok to update/replace in the test data?

ajjackson:
Unit test data updates: 1d case is indistinguishable
spec1d_lorentz_datachange

Quartz 2D case also small

spec2d_xy_lorentz_datachange

For LZO I clipped the plotting range (vmax) to exaggerate the effect and we can see that there is some subtle change to the peak intensity
spec2d_xy_lorentz_datachange_lzo

ajjackson:
2D demo with variable width functions:

broaden2d-input

broaden2d-output

ajjackson:
Ok @mducle I think this is techically there! Documentation-wise I could use a bit of perspective on what is best to elaborate where. There are a lot of places examples can go:

- docstrings
- Python doc pages for Spectrum1D and Spectrum2D
- CLI doc pages for euphonic-dos and euphonic-powder-map
- how-to guides

I think the two interesting examples to provide are
- known polynomial resolution function for TOSCA (currently I have included the Python version for Spectrum1D)
- using PyChop as a cheaper/faster/more-limited alternative to going down the Horace/Tobyfit route

What do you think is needed for this PR?

ajjackson:
@mducle  Would you be able to look at this? If you are busy maybe we can figure out some other combination of reviewers that covers the essentials.

ajjackson:
I have updated the parametrisation of kernel width from nominal error: a wider-ranging, more robust version has been developed and is included in a manuscript to be submitted soon (:crossed_fingers:).

As the existing cubic fit was used for test data and for adaptive broadening, at first we will allow them to exist in parallel and have adaptive broadening default to the cubic fit.

We should drop the inferior fit in the next major version; will create a milestoned Issue for that.

ajjackson:
Comparing the new 2D test data (chebyshev polynomial log-energy fit) to existing (cubic fit) the difference is reassuringly small.

euphonic_cheby_sanity_check

```python
from pathlib import Path

from euphonic import Spectrum2D
from euphonic.plot import plot_2d_to_axis

fig, axes = plt.subplots(ncols=2, nrows=2)

data_dir = Path.home() / 'src/euphonic/tests_and_analysis/test/data/spectrum2d'

spectra = {'Unbroadened': Spectrum2D.from_json_file(data_dir / 'synthetic_x.json'),
           'cubic': Spectrum2D.from_json_file(data_dir / 'synthetic_x_poly_broadened.json'),
           'cheby-log': Spectrum2D.from_json_file(data_dir / 'synthetic_x_poly_broadened_cheby.json')
           }

diff = Spectrum2D(spectra['cheby-log'].x_data,
                  spectra['cheby-log'].y_data,
                  spectra['cheby-log'].z_data - spectra['cubic'].z_data)

spectra['diff'] = diff

fig.suptitle(
    f"Maximum value in broadened data: {spectra['cubic'].z_data.max():.4f} \n"
    f"Maximum difference in broadened data: {diff.z_data.max():.3e}")

for ax, (label, spectrum) in zip(axes.flatten(), spectra.items()):
    plot_2d_to_axis(spectrum, ax)
    ax.set_title(label)

fig.tight_layout()
fig.savefig('euphonic_cheby_sanity_check.png')
```

ajjackson:
Small correction to error/width parametrisation breaks the corresponding test. The difference is actually negligible, so I will just update the reference file:

poly_2d_comparison

PR #246, opened 2022-10-19
Remove packaging imports

rebeccafair:
After installing Euphonic on an empty Python 3.9 conda environment I was getting an `ImportError` because packaging wasn't installed. I found that these imports in Euphonic are actually no longer used and are the last remaining references to the packaging library. Packaging is not in the Python standard library so we must remember to add it to dependencies if we're to use it again, it was omitted before. How we didn't notice this earlier I don't know, I guess we got lucky and other dependencies needed packaging so it was installed anyway.

### DIFF
--- a/euphonic/__init__.py
+++ b/euphonic/__init__.py
@@ -5,7 +5,6 @@ del get_versions
 import pint
 from pint import UnitRegistry
 from importlib_resources import files
-from packaging.version import parse as parse_version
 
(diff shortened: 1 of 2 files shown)


ajjackson (approved):
Seems fine!

I'd assume we were previously using it to check a numpy or scipy version and implement some workaround that got cleaned up with version 1.

Issue #256, opened 2022-11-16
Support MacOS 12


rebeccafair:
Currently Github actions `macos-latest` runs on MacOS 11, but MacOS 12 is also available. OpenMP might have been disabled by default in newer clang verisons provided with MacOS 12 https://mac.r-project.org/openmp/#do. We should check and fix this if so.

rebeccafair:
Seems ok, tests in #257 are passing. Guess we just wait for macos-latest to update naturally.

PR #257, opened 2022-11-22
Test on MacOS 12

rebeccafair:
#256


rebeccafair:
https://github.com/pace-neutrons/Euphonic/issues/256#issuecomment-1323479492

PR #259, opened 2023-01-03
Add paper citation to docs

rebeccafair:
I wasn't sure what citation style to use for this, so I've just made sure all the information is there. Alternatively, rather than adding to the citation docs page, we could update the CITATION.cff file so instead of having an overall DOI, we have multiple `identifiers`, one for the software itself, and one for the paper, as done here: https://github.com/citation-file-format/citation-file-format/blob/main/README.md#structure.

### DIFF
--- a/doc/source/cite.rst
+++ b/doc/source/cite.rst
@@ -19,3 +19,10 @@ or it can be read programatically as follows:
 
 The latest version of the citation file can also be found in Euphonic's code
 repository `here <https://github.com/pace-neutrons/Euphonic/blob/master/CITATION.cff>`_.
+
+Please also cite the following publication:
+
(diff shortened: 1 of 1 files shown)


ajjackson:
Maybe the best place to put it in the CITATION.cff is as `preferred-citation`
https://github.com/citation-file-format/citation-file-format/blob/main/schema-guide.md#credit-redirection

I don't think it hurts to include it in the docs as well; then we have one human-friendly request and one machine-friendly!

rebeccafair:
Ah good point, I didn't know about `preferred-citation`. I have now added it.

The paper won't show up on the DOI landing page (https://doi.org/10.5286/SOFTWARE/EUPHONIC) unless I make some changes, but I think that's fine as its just the landing page for the software?

ajjackson:
That's ok with me

PR #265, opened 2023-02-23
Allow 'color' to be passed to plot_1d

### LINKED ISSUE #264: TypeError: plot() got multiple values for keyword argument 'color'
`plot_1d` accepts extra kwargs which are automatically passed to `ax.plot`. However, if you specify `color` you get the following error:

```
TypeError: plot() got multiple values for keyword argument 'color'
```
When plotting we explicitly set the colour of each line so that in the case of multiple subplots (e.g. in a dispersion) the colour of the mode is consistent across subplots. However, we also blindly pass any extra kwargs to the plot function, so if you pass something we’ve already defined, such as color, it results in an error. It seems reasonable a user would want to set the same colour for all lines, so we should allow this.

rebeccafair:
Fixes #264

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,6 +1,11 @@
 `Unreleased <https://github.com/pace-neutrons/Euphonic/compare/v1.2.0...HEAD>`_
 ----------
 
+- Bug Fixes:
+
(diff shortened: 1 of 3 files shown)


mducle (approved):
Looks good and works for me!

Issue #270, opened 2023-05-15
Missing Py_INCREF(Py_None) in the C extension


i-loa:
**Bug report**
I am trying to use Euphonic in a program that models and fits a phonon model to experimental data, which can require 10s of thousands of calls to `ForceConstants.calculate_qpoint_phonon_modes(...)`. At some point, the program crashes with a core dump and `Fatal Python error: none_dealloc: deallocating None`. 

**Bug fix**
I believe the issue is that the function `calculate_phonons(...)` of the C extension in `_euphonic.c` returns a `Py_None` (line 303 as of today), but the reference counting is not done as required. Consequently, Python tries at some point to deallocate/garbage collect `Py_None`, which is not allowed and causes the crash.

Incrementing the `Py_None` reference counter before returning it fixes the problem:
```
    Py_INCREF(Py_None);
    return Py_None;
```

For further details see, e.g., https://www.oreilly.com/library/view/python-cookbook/0596001673/ch16s05.html

PR #271, opened 2023-05-15
Fix Py_None increment bug #270

mducle:
Change code to use the `Py_RETURN_NONE` macro to correctly increment the reference count to the `Py_None` object returned by the `calculate_phonons` C-extension.

### DIFF
--- a/.github/workflows/run_tests.yml
+++ b/.github/workflows/run_tests.yml
@@ -73,7 +73,8 @@ jobs:
         run: |
           python -m pip install --upgrade pip
           python -m pip install -r tests_and_analysis/ci_requirements.txt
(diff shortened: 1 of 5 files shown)


ajjackson:
I see we have the "spglib problem" where Numpy has to be installed first for the extensions to build. Maybe we should have that fall back to Python-only build in those cases. Do we document this requirement at the moment?

mducle:
No, it's not documented... should I add documentation or change so that we don't build the C-extensions on docs-builds?

ajjackson:
We should mention it in the installation section, but that can be a separate PR.

I don't really mind what we do for docs builds, the dependency shouldn't cause too much trouble.

ajjackson (approved):

Issue #278, opened 2023-06-19
Increase Python requirement to 3.8


ajjackson:
- Official end-of-life for Python 3.7 is June 27 2023
- Mantid is now built on at least Python 3.8 so this will not cause problems there
- We have started spotting places  in the code that it would be nice to use the [assignment expressions](https://peps.python.org/pep-0572/) (aka "the walrus operator")
- Some other Python 3.8+ features to consider using
  - positional-only parameters (probably not relevant, but could be used for more robust passing of **kwargs to matplotlib etc)
  - `@lru_cache` no longer requires an argument to decorate a function
  - `@functools.cached_property` looks handy
  - `@functools.singledispatchmethod` provides a cleaner handling of type-dispatched methods; this could be used to clean up e.g. the Spectrum objects that have a lot of type-dispatching.
  - `math.dist()` calculates Euclidean distance. Mostly we would use `np.linalg.norm` for this but it might be the better option if the data isn't already in numpy arrays.
  - `math.prod` - finally something like a lisp (* ...) for multiplication of an arbitrary number of arguments :grin: .
  - `multiprocessing.shared_memory` module (not that we should stop using C/OpenMP for "hot" parts of the code!)
    - This could be very useful for enumerating 2-phonon incoherent scattering combinations; currently Abins does this in serial
  - `typing.TypedDict` provides a proper way to type-hint heterogeneous dictionaries. Of course, a lot of those could be NamedTuple or dataclass, but for existing or JSON-imported dictionaries this is good.
  - `typing.Literal` to define a specific set of allowed values. We could be using this for e.g. 'gauss', 'lorentz' broadening options.

sf1919:
Mantid is moving to Python 3.10 at the moment ready for v6.8 release in October.

PR #279, opened 2023-06-22
Support newer Phonopy yaml versions

### LINKED ISSUE #276: Born charge / dielectric data keywords missed from some phonopy.yaml
I have a phonopy.yaml file with the section

```
nac:
  born_effective_charge:
  - # 1 ...
  ...
  dielectric_constant:
  - [  ...
  unit_conversion_factor: 14.4000
```

but Euphonic looks for the keys 'born', 'dielectric', 'nac_factor' and misses them.

Presumably the Phonopy output format has changed at some point; we should support these (better and more verbose!) keys as well.

ajjackson:
Fixes #276

Since version 2.18 Phonopy has moved some keywords into an `nac` block.

For now we retain compatibility with older versions.

A unit test is included, with data from a recent version of Phonopy (2.19.1).

I noticed this parser uses a lot of exception handling; explicit checks for key existence seem safer if that is what we are trying to determine. For now I just refactored the function being worked on; perhaps other parts can wait until we have the walrus, which is nice for this sort of thing. ( #278 )


euphonic/readers/phonopy.py
```diff
@@ -543,28 +543,38 @@ def _extract_summary(filename: str, fc_extract: bool = False
...
+            summary_dict['born'] = summary_object.get('born_effective_charge')
+            summary_dict['dielectric'] = summary_object.get(
+                'dielectric_constant')
+
+        # Cast arrays and drop keys if data not found
+        for key in ('nac_factor', 'born', 'dielectric'):
+            if summary_dict[key] is None:
+                del summary_dict[key]
+            elif key in ('born', 'dielectric'):
+                summary_dict[key] = np.array(summary_dict[key])
```
oerc0122, line 577: For the sake off saving a loop through 2 elements in a non-performance critical application, I find a second loop more explicit.
```suggestion

        for key in ('born', 'dielectric'):
            summary_dict[key] = np.array(summary_dict[key])
```

oerc0122 (approved):
Assuming the test data are OK.

euphonic/readers/phonopy.py
ajjackson: This will create array(None) if the keys were not present in the file.

ajjackson: Ah wait, actually we already deleted it so we will get a KeyError?

oerc0122: Adding
```
if key in summary_dict:
```
?
It was a suggestion as I find it describes what's going on better, definitely not required.

ajjackson: Like this?

```python
        # Drop NAC keys if no data was found
        for key in ('nac_factor', 'born', 'dielectric'):
            if summary_dict[key] is None:
                del summary_dict[key]

        for key in ('born', 'dielectric'):
            if key in summary_dict:
                summary_dict[key] = np.array(summary_dict[key])
```

ajjackson:
I'm quite puzzled by this test failure because
- it does superficially appear to be relevant as it relates to finding Born charge data
- but I can reproduce it on `master`
- but CI passed on `master` a few weeks ago

ajjackson:
I will fire off CI tests on `master` and see if they still pass... I have some conda environments where this test passes and some where it fails, and the one where it passes is too complicated to bisect :sweat_smile:

PR #284, opened 2023-08-31
Drop python 3.7 support, bump Scipy, Pint, h5py requirements

### LINKED ISSUE #278: Increase Python requirement to 3.8

ajjackson:
Closes #278 

This PR should focus on mechanics of version bump; subsequent PRs can introduce new goodies.

### DIFF
--- a/.github/workflows/build_upload_pypi_wheels.yml
+++ b/.github/workflows/build_upload_pypi_wheels.yml
@@ -10,7 +10,7 @@ jobs:
     strategy:
       matrix:
         os: [windows-latest, macos-latest, ubuntu-latest]
-        python-version: ['3.7', '3.8', '3.9', '3.10', '3.11']
+        python-version: ['3.8', '3.9', '3.10', '3.11']
         include:
           - os: windows-latest
             wheelname: win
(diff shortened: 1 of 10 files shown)


ajjackson:
Strange things happening on Windows seem to be resolved if we stay on 3.7 as the host version for tox. That's a bit silly but solves the immediate issue for minimal effort. Playing more with tox and github actìons may have to wait until next year!

oerc0122 (approved):

ajjackson:
Thanks for speedy review! I'll just call the CI manually to make sure 3.9 and 3.10 are happy before merging.

ajjackson:
Tests look fine :rocket:

PR #286, opened 2023-09-01
Try to get tox working on a newer Python version

ajjackson:
Would be good if Tox didn't need to run on Python 3.7, but we hit some mysterious issues with Windows runners.

Tried removing the login shell requirement from test runner, didn't help.

@mducle Edit: Problem was with anaconda default packages. Switching to `conda-forge` seems to solve the problems.
Whilst debugging the issues I used a lot of force-commits and messed up the history... so decided to junk it to leave just one commit. Old commits are [here](https://github.com/pace-neutrons/Euphonic/commits/5b03cac042a8bd337bcb20d1eaef166e35b4fbd2)


ajjackson:
Tests look legit, better do a workflow dispatch as well to be safe

https://github.com/pace-neutrons/Euphonic/actions/runs/6051778082

ajjackson:
All good. Thanks @mducle !

PR #287, opened 2023-09-01
Drop deprecated pkg_resources

ajjackson:
This line was causing some deprecation warnings. I think this is the right way to modernise it?

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -5,13 +5,15 @@
 
   - Python 3.7 is no longer supported
 
-    - Minimum version of scipy increased from 1.1 to 1.10
+  - Minimum version of scipy increased from 1.1 to 1.10
 
-      - This requires numpy >= 1.19.5
+    - This requires numpy >= 1.19.5
 
-    - Minimum version of pint increased from 0.10.1 to 0.19
+  - Minium version of matplotlib increased from 2.2.2 to 3.2.0
 
-    - Minimum version of h5py increaased form 2.8 to 2.10
+  - Minimum version of pint increased from 0.10.1 to 0.19
+
+  - Minimum version of h5py increaased form 2.8 to 2.10
 
 - Improvements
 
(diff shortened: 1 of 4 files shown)


ajjackson:
Oldlibs is failing, presumably there is a (Matplotlib?) version that doesn't like the Path objects and expects a string.

Could convert to string but then it will stay that way forever: tempted to narrow down the problematic version and apply tweak in that case only. Then it is more likely to be cleaned up later...

ajjackson:
After a bit of binary search, Matplotlib styles support Path objects from version 3.2 (March 2020). It doesn't seem unreasonable to bump the minimum requirement to this.

(Mantid is currently pinned to version 3.6)

ajjackson:
@oerc0122 @mducle Is there any reason to be concerned about bumping the Matplotlib dependency a bit? It's nice to do things with Path objects where possible :smile:

CHANGELOG.rst
```diff
@@ -5,13 +5,15 @@
...
   - Python 3.7 is no longer supported
 
-    - Minimum version of scipy increased from 1.1 to 1.10
+  - Minimum version of scipy increased from 1.1 to 1.10
 
-      - This requires numpy >= 1.19.5
+    - This requires numpy >= 1.19.5
 
-    - Minimum version of pint increased from 0.10.1 to 0.19
+  - Minium version of matplotlib increased from 2.2.2 to 3.2.0
```
oerc0122, line 12: Minimum

oerc0122 (changes requested):
Just a typo here.

Don't think there should be any concern about bumping.

mducle:
I don't think so, I think we just use fairly standard Matplotlib methods which haven't changed between version 2 and 3.

ajjackson:
Great, thanks :-) Will merge when tests pass again.

oerc0122:
Having taken a look at the docs, the main difference is that you no longer have Python 2 support (which I don't think is an issue at this point) and unless you're mucking around in the backend, the interfaces are much the same for standard calls.

oerc0122 (approved):

PR #290, opened 2024-01-19
Update versioneer

ajjackson:
SafeConfigParser has long been deprecated, and causes errors when used by versioneer with Python 3.12

This in turn is breaking a conda-forge build for 3.12 https://github.com/conda-forge/euphonic-feedstock/pull/31#issuecomment-1878588858


ajjackson:
Python 3.11 tests started breaking due to an API change in matplotlib. It's easily worked around, so do that here.

PR #296, opened 2024-04-16
Add error for old castep_bin files

### LINKED ISSUE #295: Cannot read Castep 16.11 bin file
The following is a valid `.castep_bin` file as far as I can tell, and has the required `FORCE_CON` field but euphonic gives the following error when trying to read it:

[cp-out.castep_bin.zip](https://github.com/pace-neutrons/Euphonic/files/14981322/cp-out.castep_bin.zip)

```
>>> fc = ForceConstants.from_castep("cp-out.castep_bin")
  File "<stdin>", line 1, in <module>
  File "C:\\mambaforge\envs\eu39\lib\site-packages\euphonic\force_constants.py", line 1833, in from_castep
    data = castep.read_interpolation_data(filename)
  File "C:\mambaforge\envs\eu39\lib\site-packages\euphonic\readers\castep.py", line 472, in read_interpolation_data
    cell_origins = np.reshape(
  File "C:\mambaforge\envs\eu39\lib\site-packages\numpy\core\fromnumeric.py", line 285, in reshape
    return _wrapfunc(a, 'reshape', newshape, order=order)
  File "C:\mambaforge\envs\eu39\lib\site-packages\numpy\core\fromnumeric.py", line 56, in _wrapfunc
    return _wrapit(obj, method, *args, **kwds)
  File "C:\mambaforge\envs\eu39\lib\site-packages\numpy\core\fromnumeric.py", line 45, in _wrapit
    result = getattr(asarray(obj), method)(*args, **kwds)
ValueError: cannot reshape array of size 1 into shape (1,3)
```

The file was generated by Castep 16.11 - possibly there was some change of format of the bin-files between Castep-16 and Castep-19 which was what euphonic was written against.

mducle:
Adds a more helpful error message if users try to use old castep_bin file (pre Castep-17.1).

An alternative fix would involve calculating the `supercell_origins` field. Some [code](https://github.com/pace-neutrons/Euphonic/blob/7608b14f7024ebb240c079ab336805e6a190b8f1/euphonic/readers/phonopy.py#L765-L768) used in the phonopy reader could possibly be re-used for this.

Fixes #295

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,9 +1,20 @@
 `Unreleased <https://github.com/pace-neutrons/Euphonic/compare/v1.3.1...HEAD>`_
 
+- Requirements
+
+  - ``packaging`` library added to dependencies.
+
 - Bug fixes
 
   - Fixed an error loading QpointPhononModes from JSON when there is a
     single q-point in the data
+
+- Improvements
+
+  - When loading ``.castep_bin`` files, explicitly check the CASTEP
+    version number and give a useful error message if this is < 17.1.
+    (These files are missing information about the unit cell origins,
+    and would previously cause an error with an unhelpful message.)
 
 - Maintenance
 
--- a/euphonic/readers/castep.py
+++ b/euphonic/readers/castep.py
@@ -469,8 +469,17 @@ def read_interpolation_data(
                     np.reshape(_read_entry(f, float_type),
                                (n_cells_in_sc, 3*n_atoms, 3*n_atoms)),
                     axes=[0, 2, 1]))
-                cell_origins = np.reshape(
-                    _read_entry(f, int_type), (n_cells_in_sc, 3))
+                try:
+                    cell_origins = np.reshape(
+                        _read_entry(f, int_type), (n_cells_in_sc, 3))
+                except ValueError:
+                    raise ValueError('Old castep file detected: '
+                        'Euphonic only supports post-Castep 17.1 files. '
+                        'Please rerun the calculation with a newer version '
+                        'of Castep with the original .cell file and a '
+                        '.castep file with a single line with the '
+                        '"continuation: <old.castep_bin>" keyword and '
+                        'use the new output .castep_bin file in Euphonic.')
                 _ = _read_entry(f, int_type)  # FC row not used
             elif header == b'BORN_CHGS':
                 born = np.reshape(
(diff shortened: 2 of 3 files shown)


ajjackson:
The `except ValueError` makes me a little nervous as this could catch an unrelated error and give the wrong message. Would

```python
cell_origins = _read_entry(f, int_type)
if isinstance(cell_origins, int):
    raise ValueError('Old castep file detected: '
                     'Euphonic only supports post-Castep 17.1 files. '
                     'Please rerun the calculation with a newer version '
                     'of Castep with the original .cell file and a '
                     '.castep file with a single line with the '
                     '"continuation: <old.castep_bin>" keyword and '
                     'use the new output .castep_bin file in Euphonic.')
cell_origins = np.reshape(cell_origins, (n_cells_in_sc, 3))
```

be a bit safer?

mducle:
Thanks Adam! I was going to try to look into the test failure - it doesn't seem to be related to this (I get the same errors on `master`) - but I guess that should probably go into a different PR...

krefson:
The dump files contain a version number, and I suggest checking against that.

The checkpoint version is encoded as a floating-point number as  ASCII in a character*10 field in the record immediately following BEGIN_PARAMETERS_DUMP.

The format change was introduced in CASTEP 17.x, so  testing for < 17.0 and reporting "not supported" should work as a robust  safeguard.

krefson:
If you really wanted to read dump files with version <= 16.x, then you will indeed need to recompute  the supercell_origins vector.  There is a short Fortran routine cell.f90 named `cell_generate_supercell_origins` which would be easy to transliterate into Python.

ajjackson:
I don't think we need to support CASTEP versions < 17, especially if an external workaround is available.

I'll implement the version number check, that seems a better way of handling things.

ajjackson:
Ok, I am using `packaging.version` to compare the version numbers. This adds a dependency, unfortunately, but

- it is quite a common one that replaces some things that were in standard library
- (e.g. it is also required by matplotlib, sphinx, black and ipython)
- it makes a rather ugly comparison operation very clean and easy 😅

ajjackson:
I think we should wait until #298 is merged and rebase this branch in order to get a clean error check before a final review/merge.

mducle (commented):
Looks good. Thanks for making the check much better!

In the interest of keeping a clean commit history maybe we ought to actually close this PR unmerged and create a new one with just the last commit (checking the Castep version in the `.castep_bin` file)?

ajjackson:
We can also achieve that from this PR with the "squash and merge" option. I'll add a changelog line first, though.

PR #303, opened 2024-07-02
Flatten q-point indices for compatibility with Numpy 2.0

ajjackson:
With Numpy 2.0, np.unique returns a 2-D array when applied to q-points. This in turn gets broadcast and adds an extra dimension to the frequency/mode arrays, causing all kinds of chaos.

Flattening the array here should be a no-op for Numpy <2 and fix the issue.

### DIFF
--- a/.github/workflows/build_upload_pypi_wheels.yml
+++ b/.github/workflows/build_upload_pypi_wheels.yml
@@ -14,7 +14,7 @@ jobs:
         include:
           - os: windows-latest
             wheelname: win
-          - os: macos-latest
+          - os: macos-12
             wheelname: macos
           - os: ubuntu-latest
             wheelname: manylinux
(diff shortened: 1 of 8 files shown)


ajjackson:
Mac test failures are Brille related, not much we can do about them here except ask for Brille build config to be updated https://github.com/brille/brille/issues/100

ajjackson:
Ok, after a lot of experimentation one way to get a Brille test working is to use an older Mac runner configuration.

Options include
- pin that until Brille supports a newer one https://github.com/brille/brille/issues/100
- use Mac-latest to test without Brille and Mac-12 to test with Brille
- grab an experimental updated Brille branch in the tests; once that is shown to work we may not have to wait long before it hits main Brille release anyway!

ajjackson:
Aha, here is the Numpy change. (It's a big changelog!) https://numpy.org/doc/stable/release/2.0.0-notes.html#np-unique-return-inverse-shape-for-multi-dimensional-inputs

PR #304, opened 2024-07-05
Test Brille fix

ajjackson:
Test updates to Brille build system https://github.com/brille/brille/issues/100

### DIFF
--- a/.github/workflows/build_upload_pypi_wheels.yml
+++ b/.github/workflows/build_upload_pypi_wheels.yml
@@ -14,7 +14,7 @@ jobs:
         include:
           - os: windows-latest
             wheelname: win
-          - os: macos-12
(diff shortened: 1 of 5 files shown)


ajjackson:
Interesting, the "all" test with all optional dependencies passes, but the brille-only test fails. Let's see if we can figure out which dependency drives this.

ajjackson:
Ah, I don't think it's dependency-related; that's just the only setup which runs the main Brille interface tests.

ajjackson:
This looks very promising, but the CI is being brought down by an unrelated deprecation warning! (There is a particular test that is supposed to pass without any warnings...) The issues seems to lie inside an upstream library (SeeK-path is annoying spglib) so for now I will suppress it.

PR #305, opened 2024-07-12
Spectrum refactoring, Spectrum2DCollection

ajjackson:
As part of a major refactor of Abins, we would like to use a Spectrum2DCollection object which behaves similarly to Spectrum1DCollection. This allows the code to largely treat 1D and 2D data interchangably, simplifying the logic and providing access to select/groupby "table-like" operations on the data.

This is implemented by factoring out the "collection"-specific methods to a "mixin" class. This gives a reasonable class heirarchy in which Spectrum1D, Spectrum2D, Spectrum1DCollection and Spectrum2DCollection are all children of Spectrum but not of each other.

In the process, we also found some limitations and inefficiencies in the existing Spectrum objects:
- iteration over Spectrum1DCollection is very slow due to redundant copying, checks and unit conversion
  - to allow cheaper iterations over metadata without creating new objects, this PR adds an `iter_metadata` method that iterates over metadata lines, merging the top-level and "line-data" parts on the fly. This is also a good building block to simplify some of the other metadata-handling methods.
- select() and group_by() do not properly handle metadata that is common to all rows (and so in the top-level of the metadata dict). This is made more robust in reimplementations of those methods.


euphonic/spectra.py
```diff
@@ -1656,6 +1780,149 @@ def from_dict(cls: Type[T], d: Dict[str, Any]) -> T:
...
+            x_tick_labels: Optional[Sequence[Tuple[int, str]]] = None,
+            metadata: Optional[Metadata] = None
+    ) -> None:
+        _check_constructor_inputs(
+            [z_data, x_tick_labels, metadata],
+            [Quantity, [list, type(None)], [dict, type(None)]],
+            [(-1, -1, -1), (), ()],
+            ['z_data', 'x_tick_labels', 'metadata'])
+        nx = z_data.shape[1]
+        ny = z_data.shape[2]
```
oerc0122, line 1838: Just to check, these shouldn't be `[0]` and `[1]`.

If not, would it be better as:
```python
_, nx, ny = z_data.shape
```

euphonic/spectra.py
```diff
@@ -1656,6 +1780,149 @@ def from_dict(cls: Type[T], d: Dict[str, Any]) -> T:
...
+            [Quantity, Quantity],
+            [[(nx,), (nx + 1,)], [(ny,), (ny + 1,)]],
+            ['x_data', 'y_data'])
+
+        self._set_data(x_data, 'x')
+        self._set_data(y_data, 'y')
+        self.x_tick_labels = x_tick_labels
+        self._set_data(z_data, 'z')
+        if metadata and 'line_data' in metadata.keys():
+            if len(metadata['line_data']) != len(z_data):
```
oerc0122, line 1850: Don't need `.keys`. if `metadata` is guaranteed dict would:
```python
if metadata.get("line_data", len(z_data)) != len(z_data):  # Assume true
```
work (though having written that it's shorter, but slightly misleading)?

Perhaps
```python
if (ld := metadata.get("line_data", None)) and ld != len(z_data):
```

euphonic/spectra.py
```diff
@@ -1656,6 +1780,149 @@ def from_dict(cls: Type[T], d: Dict[str, Any]) -> T:
...
+                raise ValueError(
+                    f'z_data contains {len(z_data)} spectra, but '
+                    f'metadata["line_data"] contains '
+                    f'{len(metadata["line_data"])} entries')
+        self.metadata = {} if metadata is None else metadata
```
oerc0122, line 1855: Do you prefer this way or
```suggestion
        self.metadata = metadata if metadata is not None else {}
```

euphonic/spectra.py
```diff
@@ -1656,6 +1780,149 @@ def from_dict(cls: Type[T], d: Dict[str, Any]) -> T:
...
+
+    def _split_by_indices(self, indices: Sequence[int] | np.ndarray
```
oerc0122, line 1857: Classmethod?

tests_and_analysis/test/data/spectrum2d/quartz_fuzzy_map_0.json
```diff
@@ -0,0 +1,267 @@
...
+        0.6666666666666666,
+        0.7777777777777777,
+        0.8888888888888888,
+        1.0
+    ],
+    "x_data_unit": "1 / angstrom",
+    "x_tick_labels": [
+        [
+            0,
+            "$\\Gamma$"
```
oerc0122, line 24: Does this load properly?

tests_and_analysis/test/euphonic_test/test_spectrum2dcollection.py
```diff
@@ -0,0 +1,214 @@
...
+def get_spectrum2dcollection(json_filename):
+    """Get Spectrum2DCollection reference data object"""
+    return Spectrum2DCollection.from_json_file(
+        get_spectrum2dcollection_path(json_filename))
+
+
+@pytest.fixture
+def quartz_fuzzy_collection() -> Spectrum2DCollection:
+    """Coarsely sampled quartz bands in a few directions"""
+    return get_spectrum2dcollection("quartz_fuzzy_map.json")
```
oerc0122, line 32: Do these accept `pathlib.Path`? Have you tested it?

tests_and_analysis/test/euphonic_test/test_spectrum2dcollection.py
```diff
@@ -0,0 +1,214 @@
...
+        """Construct Spectrum2DCollection with __init__()"""
+        n_x = 10
+        n_y = 20
+        n_z = 5
+
+        x_data = ureg.Quantity(np.linspace(0, 100, n_x), "1 / angstrom")
+        y_data = ureg.Quantity(np.linspace(0, 2000, n_y), "meV")
+        z_data = ureg.Quantity(np.random.random((n_z, n_x, n_y)), "1 / meV")
+
+        metadata = {"flavour": "chocolate",
```
oerc0122, line 108: Dang. I was hoping for a liquorice spectrum.

tests_and_analysis/test/euphonic_test/test_spectrum2dcollection.py
```diff
@@ -0,0 +1,214 @@
...
+                    "line_data": [{"index": i} for i in range(n_z)]}
+
+        x_tick_labels = [(0, "Start"), (n_x - 1, "END")]
```
oerc0122, line 111: Weird that "start" is title and "end" is upper.

euphonic/spectra.py
```diff
@@ -669,10 +674,387 @@ def broaden(self: T, x_width,
...
+        Raises
+        ------
+        ValueError
+            If no matching spectra are found
+        """
+        # Convert all items to sequences of possibilities
+        select_key_values = dict(
+            (key, (value,)) if isinstance(value, (int, str)) else (key, value)
+            for key, value in select_key_values.items()
+        )
```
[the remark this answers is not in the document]
oerc0122: Though with recent python (I think 3.10+, whichever added or on types) can do `isinstance(value, int | str)`

oerc0122 (commented):
Just a few code-style questions. Functionally, I'm trusting it's good.

euphonic/spectra.py
```diff
@@ -669,10 +674,387 @@ def broaden(self: T, x_width,
...
+        if isinstance(item, Integral):
+            return self._item_type(**init_kwargs)
+
+        return type(self)(**init_kwargs)
+
+    def _validate_item(self, item: Integral | slice | Sequence[Integral] | np.ndarray
+                       ) -> None:
+        """Raise Error if index has inappropriate typing/range"""
+        if isinstance(item, Integral):
+            return
```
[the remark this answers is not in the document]
ajjackson: I see what you mean. The issue is that the cases need to raise different exception types and messages, so the return type would need to be something like `Tuple[bool,Exception|None,str]`. The point of breaking out this method is to avoid clutter in the main function, so if we need _too_ much logic there it might as well all be inlined.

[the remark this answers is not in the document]
ajjackson: Perhaps it is better think of it as "raises, or doesn't" as the `None` is not used? Maybe a function name `_raise_error_if_invalid_item()` would be clearer?

euphonic/spectra.py
```diff
@@ -669,10 +674,387 @@ def broaden(self: T, x_width,
...
+    @overload
+    def _get_item_metadata(self, item: slice | Sequence[Integral] | np.ndarray
+                           ) -> Metadata:  # noqa: F811
+        """Get a metadata collection (may include line_data)"""
+
+    def _get_item_metadata(self, item):  # noqa: F811
+        """Produce appropriate metadata for __getitem__"""
+        metadata_lines = list(self.iter_metadata())
+
+        if isinstance(item, Integral):
```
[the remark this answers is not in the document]
ajjackson: I initially decided against it because of the special case for `len(item) == 1`, which doesn't fit.

But on reflection that special case is a bad idea and not indicated properly in the type  hint anyway. Let's see how much harm is really done by removing it...

euphonic/spectra.py
```diff
@@ -669,10 +674,387 @@ def broaden(self: T, x_width,
...
+        spectra are retained in the top level dictionary, any
+        others are put in the individual 'line_data' entries
+        """
+        return type(self).from_spectra([*self, *other])
+
+    def iter_metadata(self) -> Generator[OneLineData, None, None]:
+        """Iterate over metadata dicts of individual spectra from collection"""
+        common_metadata = dict(
+            (key, self.metadata[key])
+            for key in set(self.metadata.keys()) - {"line_data",})
```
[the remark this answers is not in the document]
ajjackson: No idea why it was written this way, probably an evolutionary artifact.

euphonic/spectra.py
```diff
@@ -669,10 +674,387 @@ def broaden(self: T, x_width,
...
+        if line_data is None:
+            line_data = itertools.repeat({}, len(self._get_raw_spectrum_data()))
+
+        for one_line_data in line_data:
+            yield common_metadata | one_line_data
+
+    def _select_indices(self, **select_key_values) -> list[int]:
+        required_metadata = select_key_values.items()
+        indices = [i for i, row in enumerate(self.iter_metadata())
+                   if required_metadata <= row.items()]
```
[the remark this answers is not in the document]
ajjackson: How about

```python
    def _select_indices(self, **select_key_values) -> list[int]:
        """Get indices of items that match metadata query

        The target key-value pairs are a subset of the matching data, e.g.

          self._select_indices(species="Na", weight="coherent")

        will match metadata rows

          {"species": "Na", "weight": "coherent"}

        and

          {"species": "Na", "weight": "coherent", "mass": "22.9898"}

        but not

          {"species": "Na"}

        or

          {"species": "K", "weight": "coherent"}
        """
        required_metadata = select_key_values.items()
        indices = [i for i, row in enumerate(self.iter_metadata())
                   if required_metadata <= row.items()]
        return indices
```

euphonic/spectra.py
```diff
@@ -1656,6 +1780,149 @@ def from_dict(cls: Type[T], d: Dict[str, Any]) -> T:
...
+    def __init__(
+            self, x_data: Quantity, y_data: Quantity, z_data: Quantity,
+            x_tick_labels: Optional[Sequence[Tuple[int, str]]] = None,
+            metadata: Optional[Metadata] = None
+    ) -> None:
+        _check_constructor_inputs(
+            [z_data, x_tick_labels, metadata],
+            [Quantity, [list, type(None)], [dict, type(None)]],
+            [(-1, -1, -1), (), ()],
+            ['z_data', 'x_tick_labels', 'metadata'])
```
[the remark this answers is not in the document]
ajjackson: Yeah that looks much better to me. I'm reluctant to touch it in the middle of this PR though, let's break that out to another Issue.

euphonic/spectra.py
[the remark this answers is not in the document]
oerc0122: Yeah, I get what you mean. I think `validate` is fine as a name for it, like I said it was this weird feeling that there was code-smell somewhere, but I couldn't quite pick it out.

I think maybe documenting the raises in the docstring as a block might help.

euphonic/spectra.py
oerc0122: Always just prefer to check as someone who has to switch languages regularly from 0-indexed and 1-indexed and back again.

euphonic/spectra.py
ajjackson: Seems a good idea to guarantee `metadata` first and simplify the logic a little.

euphonic/spectra.py
ajjackson: No strong preference really, I like that yours presents the "usual" case first though.

tests_and_analysis/test/euphonic_test/test_spectrum2dcollection.py
ajjackson: I do like putting odd things into tests of generic systems, to make sure we didn't accidentally make them more specific than expected.

tests_and_analysis/test/euphonic_test/test_spectrum2dcollection.py
ajjackson: I think this was intended to catch any unexpected case-normalisation as a side-effect of logic.

euphonic/spectra.py
```diff
@@ -669,10 +674,387 @@ def broaden(self: T, x_width,
         return new_spectrum
 
 
-LineData = Sequence[Dict[str, Union[str, int]]]
+OneLineData = Dict[str, Union[str, int]]
+LineData = Sequence[OneLineData]
+Metadata = Dict[str, Union[str, int, LineData]]
```
oerc0122, line 679: Presumably not possible to build `TypedDict`s

euphonic/spectra.py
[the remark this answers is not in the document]
ajjackson: Ok, I'll fill out the docstring properly

euphonic/spectra.py
```diff
@@ -669,10 +674,387 @@ def broaden(self: T, x_width,
...
+        spectra are retained in the top level dictionary, any
+        others are put in the individual 'line_data' entries
+        """
+        return type(self).from_spectra([*self, *other])
+
+    def iter_metadata(self) -> Generator[OneLineData, None, None]:
+        """Iterate over metadata dicts of individual spectra from collection"""
+        common_metadata = {
+            key: self.metadata[key]
+            for key in set(self.metadata.keys()) - {"line_data",}}
```
oerc0122, line 884: Looking at this again with not midnight eyes.
```suggestion
            for key in self.metadata
            if key != "line_data"
            }
```

euphonic/spectra.py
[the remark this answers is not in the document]
oerc0122: Actually, you've got `toolz` does this mean we could do this with a `mapvalues`?

euphonic/spectra.py
ajjackson: No, TypedDict is for when you know what the keys are. We have no opinion about which keys should have int values and which should have str values, so `Dict[str, str | int]` is exactly the right annotation.

euphonic/spectra.py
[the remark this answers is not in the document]
ajjackson: This isn't really an example, just an explanation of the intended behaviour (which hopefully makes it clear what the `<=` is going to do). A proper example would need more boilerplate showing how to construct `self` with different metadata.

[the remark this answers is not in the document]
oerc0122: Yeah, possibly best, particularly given the abstract class. Think I've just become a fan of doctest recently. Looks good!

euphonic/spectra.py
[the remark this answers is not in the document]
ajjackson: What do you think, more or less approachable?
```python
        from toolz.dicttoolz import valmap
        def ensure_sequence(value: int | str | Sequence[int | str]
                            ) -> Sequence[int | str]:
            return (value,) if isinstance(value, (int, str)) else value

        select_key_values = valmap(ensure_sequence, select_key_values)
```

euphonic/spectra.py
```diff
@@ -669,10 +674,387 @@ def broaden(self: T, x_width,
...
+        """
+        # This is for combining multiple separate spectrum metadata,
+        # they shouldn't have line_data
+        for metadata in all_metadata:
+            assert 'line_data' not in metadata.keys()
+
+        # Combine all common key/value pairs into new dict
+        combined_metadata = dict(
+            set(all_metadata[0].items()).intersection(
+                *[metadata.items() for metadata in all_metadata[1:]]))
```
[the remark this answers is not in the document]
ajjackson: I think it can be made a bit more eloquent with
```python
combined_metadata = dict(
    reduce(set.intersection,
           (set(metadata.items()) for metadata in all_metadata)))
```

Is that readable to you?

euphonic/spectra.py
```diff
@@ -669,10 +674,387 @@ def broaden(self: T, x_width,
...
+
+        return _obj_to_dict(self, attrs)
+
+    @classmethod
+    def from_dict(cls: Self, d: dict) -> Self:
+        """Initialise a Spectrum Collection object from dict"""
+        data_keys = list(f"{dim}_data" for dim in cls._bin_axes)
+        data_keys.append(cls._spectrum_data_name())
+
+        d = _process_dict(d,
```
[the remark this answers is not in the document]
ajjackson: I'm struggling for a good variable name here! Pandas uses `data` as the from_dict argument, but here that's very confusing because "data" is used in the other variable names to mean x_data, y_data, z_data...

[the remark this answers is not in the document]
ajjackson: (The other problem is that I used `d` for consistency with existing methods. If we rename all of them that's a break in public API. It should be handled with backward compatibility and deprecation warning.)

euphonic/spectra.py
```diff
@@ -1138,116 +1347,31 @@ def broaden(self: T,
...
         Returns
         -------
-        selected_spectra
-           A Spectrum1DCollection containing the selected spectra
-
-        Raises
-        ------
-        ValueError
-            If no matching spectra are found
+        spectrum_collection
```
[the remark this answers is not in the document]
ajjackson: I'm not sure why that would be helpful? The parameters and meanings are the same.

euphonic/spectra.py
ajjackson: This code is somewhat duplicated with Spectrum1DCollection, so I'll factor both out to the verbose but clear:

```python
    def _check_metadata(self) -> None:
        """Check self.metadata['line_data'] is consistent with collection size

        Raises
        ------
        ValueError
            Metadata contains 'line_data' of incorrect length
        
        """
        if 'line_data' in self.metadata:
            collection_size = len(self._get_raw_spectrum_data())
            n_lines = len(self.metadata['line_data'])

            if n_lines != collection_size:
                raise ValueError(
                    f'{self._spectrum_data_name()} contains {collection_size} '
                    f'spectra, but metadata["line_data"] contains '
                    f'{n_lines} entries')

```

euphonic/spectra.py
ajjackson: I don't see how that could work. 1D version needs access to x_data and y_data, 2D version needs access to x_data, y_data and z_data. Both are called by a common `split()` method from parent class.

tests_and_analysis/test/euphonic_test/test_spectrum2dcollection.py
ajjackson: Only `str` is promised/documented at the moment. But Path should work. I'll create an Issue to look at that.

euphonic/spectra.py
ajjackson: 🤦 

Ok, done but with two fewer linebreaks.

euphonic/spectra.py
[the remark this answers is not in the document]
oerc0122: Honestly, I like it? Maybe if `ensure_sequence` were moved into some general utilities module since it's a very common thing if such a thing exists.

euphonic/spectra.py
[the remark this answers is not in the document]
oerc0122: That makes more sense to me, yeah. Not sure whether it will to others though.

euphonic/spectra.py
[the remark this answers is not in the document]
oerc0122: Well, best not to break it if there's no real improvement.

euphonic/spectra.py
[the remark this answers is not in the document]
oerc0122: For me because it links to the actual implementation which may help. But that's probably down to personal taste.

euphonic/spectra.py
oerc0122: Yeah, that makes things clear for me.

euphonic/spectra.py
oerc0122: I just saw the `type(self)` which immediately makes me go "Classmethod?" possibly a bad reflex, but just to think about.

tests_and_analysis/test/euphonic_test/test_spectrum2dcollection.py
oerc0122: Makes sense!

euphonic/spectra.py
ajjackson: Here's another option with less repetition:
```python
        from toolz.dicttoolz import keyfilter
        common_metadata = keyfilter(lambda key: key != "line_data",
                                    self.metadata)
```

oerc0122: Here, I think you're accidentally replacing standard Python with toolz for the sake of it. 

I'm all for functional as you know, but I think here this is obfuscated to people who don't know toolz adding an extra setop, the more standard solution is better if only for familiarity. 

Plus a lambda and a function calls rather than a simple comprehension isn't reducing complexity for me.

ajjackson: Before I revert that, take a look at the new version of `_combine_metadata` I just pushed. In this case the `keyfilter` avoids nested comprehensions so I think the clarity gain is better.

But still perhaps the overall difference is too marginal to be worth it.

euphonic/spectra.py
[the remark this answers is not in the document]
ajjackson: Actually the special case was not so bad, it deals with an inconsistency in output of `itemgetter`. Instead we can use

```python
return _combine_metadata([metadata_lines[i] for i in item])
```
for all sequence cases.

I tried to use singledispatchmethod but it didn't go well:
- doesn't support Union types so Sequence[int] and np.ndarray have to be handled separately
- doesn't support Sequence[int] at all
- at one point I was getting weird abstract class errors that child classes were obliged to implement their own `singledispatchmethod` function.

As far as I can tell, then, the standard library `singledispatch` features are pretty useless 😞 
The Union types may have been fixed in a more recent Python version?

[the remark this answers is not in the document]
oerc0122: Yeah, way to to union types is:
```
x.register(int)
x.register(float)
def _(...)
```
But it doesn't support subscripted types which is a pain.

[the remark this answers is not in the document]
ajjackson: I tried it with bare `Sequence` but then I hit the "abstract method" weirdness.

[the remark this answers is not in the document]
ajjackson: Sadly abandoning this approach for now. I wish it worked.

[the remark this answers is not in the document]
oerc0122: Yeah. More important to get this merged!

euphonic/spectra.py
ajjackson: Getting a bit silly, but maybe this is part of the process of finding a good version:

```python
        # Put all other per-spectrum metadata in line_data
        from toolz.curried import map
        from toolz.functoolz import pipe

        keep_uncommon_metadata_items = keyfilter(
            lambda key: key not in common_metadata)

        line_data = pipe(all_metadata,
                         map(keep_uncommon_metadata_items),
                         list)

```

ajjackson: Taking away the Extremely Functional parts but keeping the named keyfilter function:

```python
        # Put all other per-spectrum metadata in line_data
        keep_uncommon_metadata_items = keyfilter(
            lambda key: key not in common_metadata)

        line_data = list(map(keep_uncommon_metadata_items, all_metadata))
```

oerc0122: I still think we're better off with:
```python
common_metadata = {
            key: self.metadata[key] for key in self.metadata if key != "line_data"
}
```

tests_and_analysis/test/data/spectrum2d/quartz_fuzzy_map_0.json
ajjackson: Looks good 😄 

Figure_1

oerc0122 (approved):

euphonic/spectra.py
```diff
@@ -669,10 +676,438 @@ def broaden(self: T, x_width,
...
+        From a sequence of metadata dictionaries, combines all common
+        key/value pairs into the top level of a metadata dictionary,
+        all unmatching key/value pairs are put into the 'line_data'
+        key, which is a list of metadata dicts for each element in
+        all_metadata
+        """
+        # This is for combining multiple separate spectrum metadata,
+        # they shouldn't have line_data
+        for metadata in all_metadata:
+            assert 'line_data' not in metadata.keys()
```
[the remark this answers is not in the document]
ajjackson: Indeed, I think I did this while playing with the rest of the function and then lost it in a revert.

Second form was considered: in this case, it uses more words to say the same thing in a more confusing order.

oerc0122 (approved):

PR #307, opened 2024-07-17
Drop Python 3.8, 3.9 and add Python 3.12

ajjackson:
Python 3.9 isn't _quite_ dead yet but supporting it on ARM Mac is quite annoying. 3.12 is very much out there in production now.

### DIFF
--- a/euphonic/readers/castep.py
+++ b/euphonic/readers/castep.py
@@ -111,7 +111,7 @@ def read_phonon_dos_data(
     _, idx = np.unique(atom_type, return_index=True)
     unique_types = atom_type[np.sort(idx)]
     for i, species in enumerate(unique_types):
-        dos_dict[species] = dos_data[:, i + 2]/dos_conv
+        dos_dict[str(species)] = dos_data[:, i + 2]/dos_conv
 
     return data_dict
 
--- a/euphonic/util.py
+++ b/euphonic/util.py
@@ -1,5 +1,6 @@
 from collections import OrderedDict
 from functools import reduce
+from importlib.resources import files
 import itertools
 import json
 import math
@@ -11,7 +12,6 @@ import warnings
 import numpy as np
 import seekpath
 from seekpath.hpkot import SymmetryDetectionError
-from importlib_resources import files
 from pint import UndefinedUnitError
 
 from euphonic import ureg, Quantity
--- a/euphonic/util.py
+++ b/euphonic/util.py
@@ -1,4 +1,3 @@
-from collections import OrderedDict
 from functools import reduce
 from importlib.resources import files
 import itertools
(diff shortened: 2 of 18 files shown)


ajjackson:
I have included a _little_ bit of type-hinting cleanup to take advantage of Python 3.10 features. More of this can be done in a separate PR, but doing some here will help convince me that we really aren't still running this on Python 3.8 somewhere in the test suite.

euphonic/readers/castep.py
oerc0122, line 1: Should these changes be part of this PR?

euphonic/util.py
```diff
@@ -52,7 +52,7 @@ def direction_changed(qpts: np.ndarray, tolerance: float = 5e-6
     return np.abs(np.abs(dot) - modq[1:]*modq[:-1]) > tolerance
 
 
-def is_gamma(qpt: np.ndarray) -> Union[bool, np.ndarray]:
+def is_gamma(qpt: np.ndarray) -> bool | np.ndarray:
```
oerc0122, line 55: So much nicer.

euphonic/util.py
```diff
@@ -444,8 +444,8 @@ def _cell_vectors_to_volume(cell_vectors: Quantity) -> Quantity:
 
 
 def _get_unique_elems_and_idx(
-        all_elems: Sequence[Tuple[Union[int, str], ...]]
-        ) -> 'OrderedDict[Tuple[Union[int, str], ...], np.ndarray]':
+        all_elems: Sequence[tuple[int | str, ...]]
+        ) -> 'OrderedDict[tuple[int | str, ...], np.ndarray]':
```
oerc0122, line 448: Does ordered dict need to be here if post 3.7?

release_tox.ini
```diff
@@ -36,10 +36,10 @@ commands_pre =
     --only-binary 'euphonic'
 commands = {[testenv]test_command}
 
-[testenv:pypi-py38-min]
+[testenv:pypi-py310-min]
 install_command = python -m pip install --force-reinstall {opts} {packages}
 deps =
-    numpy==1.19.5
+    numpy==1.21.3
```
oerc0122, line 42: The `docs/requirements.txt` still has 1.19.5 as its minimum version, is this fine to push up?

setup.py
```diff
@@ -139,16 +139,15 @@ def run_setup():
         include_package_data=True,
```
oerc0122, line 139: In setup.py you could *strictly* do a check against python version and set requirements based on that and pass through. Still not sure that's wise, but worth noting.

tests_and_analysis/test/run_tests.py
```diff
@@ -143,7 +142,7 @@ def run_tests(pytest_options: List[str], do_report_coverage: bool,
     pytest_options = ['--import-mode=append'] + pytest_options
 
     # Start recording coverage if requested
-    cov: Union[coverage.Coverage, None] = None
+    cov: coverage.Coverage | None = None
```
oerc0122, line 145: Is it still recommended this be expressed as `Optional`? Fine either way.

oerc0122 (approved):
All looks good, just one minor change I'm not sure should be in here and a couple of questions.

euphonic/readers/castep.py
ajjackson: It fixes a failure in the doc-tests, which only emerged now that the doc-tests are running on a newer Python/Numpy.

This _could_ be discussed in a separate issue but as it is a one-liner with seemingly no negative consequences it might as well be rolled in here! It might merit a CHANGELOG mention... actually I forgot to update the changelog anway 🤦

euphonic/util.py
ajjackson: OrderedDict is not deprecated and can still be useful for expressing that the order-conservation is important to an algorithm.

In this case it's not _very_ important, though, so can be cleaned up.

setup.py
ajjackson: I don't think it's necessary here. It's a useful tool when things go wrong, but for now let's assume that the packages we are using all define their requirements correctly and make the installer responsible for getting a compatible set.

If we hit trouble with a dependency's metadata, that's the time to introduce special version-dependent pins.

tests_and_analysis/test/run_tests.py
ajjackson: I prefer to only use `Optional` to indicate an optional _parameter_. Here we have a variable that, depending on code branches, may or may not be `None`; it's a different situation, where it is not so obvious what "Optional" would mean.

oerc0122 (approved):
Once Conda is no longer FUBAR, looks good.

ajjackson:
This won't hit conda-forge until the next release anyway, hopefully the fire will be out by then. Thanks for looking 😄

PR #311, opened 2024-07-26
Improve handling of spectral units

ajjackson:
It looks like spectral intensities are sometimes handled in energy units (such as when imported from CASTEP DOS). If these are converted to/from their proper (reciprocal energy) units, this can cause divide-by-zero errors as pint performs a wavenumber <-> wavelength conversion and takes the reciprocal of the spectral intensities, including empty bins.

Here we introduce another mechanism for dealing with spectral units: a few extra unit conversions in a special "reciprocal_spectroscopy" context (which comes an "rs" alias for convenience). This stops Pint from choking on the `1/(1/cm) <-> 1/meV` conversions.

The broadening system is made a bit more robust; previously the variable-width broadening assumed that things can always be converted to energy units.

Perhaps the most serious issue addressed here is the way the Spectrum data "getters" (hidden behind a `@property` interface) work.

The current spectrum getters create a unit conversion factor and then
multiply the raw array by it. There are two problems with this:

- Multiplying any iterator by a Quantity or Unit leads to Pint
  checking through all the data for consistency. This can be quite
  expensive, and is unnecessary for these trusted array objects. When
  iterating over spectra yielded from a Spectrum1DCollection, the cost
  can become quite noticeable.

- Some unit conversions in the spectroscopy context involve taking
  a reciprocal, e.g. conversion from cm -> 1/cm is allowed. But if
  this is done to a _factor_ the data itself will not be inverted. If
  we start with a Spectrum1D with x_data in wavenumber (1/cm) then

    ```python
    spectrum.x_data.to("cm")
    ```

  and

    ```python
    spectrum.x_data_unit = "cm"
    spectrum.x_data
    ```

  will not give the same result; the second method leaves the actual
  values untouched (as cm / (1/cm) = 1).

To further improve these conversions I included some technically-redundant energy-wavenumber conversions in the "reciprocal_spectroscopy" context. With these available, `.to(..., "rs")` conversions can avoid needlessly taking reciprocals of the data and raising divide-by-zero errors. That _should_ be more efficient, too, but it's not clear if that works in practice.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -24,10 +24,22 @@
 
     - Minimum version of threadpoolctl increased from 1.0 to 3.0.
 
-- Big fixes
+- Improvements
+
+  - A "reciprocal_spectroscopy" Pint context is made available in the
+    unit registry for tricky conversions between reciprocal
+    frequency/energy units. It is not active by default but can be
+    enabled with e.g.
+
+      (10 * ureg("1 / meV")).to("cm", "reciprocal_spectroscopy")
+
+- Bug fixes
 
   - Metadata strings from Castep-imported PDOS data are now converted
     from numpy strings to native Python strings.
+
+  - Spectra from CASTEP .phonon_dos files are now imported with units
+    of reciprocal energy (e.g. 1/meV)
 
 
 `v1.3.2 <https://github.com/pace-neutrons/Euphonic/compare/v1.3.1...v1.3.2>`_
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -33,6 +33,9 @@
 
       (10 * ureg("1 / meV")).to("cm", "reciprocal_spectroscopy")
 
+    This can also help to avoid divide-by-zero issues when performing
+    energy <-> wavenumber conversions.
+
 - Bug fixes
 
   - Metadata strings from Castep-imported PDOS data are now converted
@@ -40,6 +43,11 @@
 
   - Spectra from CASTEP .phonon_dos files are now imported with units
     of reciprocal energy (e.g. 1/meV)
+
+- Maintenance
+
+  - Cleared up unit-conversion-related warnings, de-cluttering the
+    expected test suite output.
 
 
 `v1.3.2 <https://github.com/pace-neutrons/Euphonic/compare/v1.3.1...v1.3.2>`_
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -24,6 +24,9 @@
 
     - Minimum version of threadpoolctl increased from 1.0 to 3.0.
 
+  - `toolz <https://toolz.readthedocs.io/en/latest/index.html>`_ is
+    added to the testing (tox) requirements
+
 - Improvements
 
   - A "reciprocal_spectroscopy" Pint context is made available in the
--- a/euphonic/validate.py
+++ b/euphonic/validate.py
@@ -82,7 +82,8 @@ def _check_unit_conversion(obj: object, attr_name: str, attr_value: Any,
     if hasattr(obj, attr_name):
         if attr_name in unit_attrs:
             try:
-                _ = ureg(getattr(obj, attr_name)).to(attr_value)
+                _ = ureg(getattr(obj, attr_name)).to(attr_value,
+                                                     "reciprocal_spectroscopy")
             except DimensionalityError:
                 raise ValueError((
                     f'"{attr_value}" is not a known dimensionally-consistent '
--- a/euphonic/validate.py
+++ b/euphonic/validate.py
@@ -82,8 +82,8 @@ def _check_unit_conversion(obj: object, attr_name: str, attr_value: Any,
     if hasattr(obj, attr_name):
         if attr_name in unit_attrs:
             try:
-                _ = ureg(getattr(obj, attr_name)).to(attr_value,
-                                                     "reciprocal_spectroscopy")
+                ureg(getattr(obj, attr_name)).ito(attr_value,
+                                                  "reciprocal_spectroscopy")
             except DimensionalityError:
                 raise ValueError((
                     f'"{attr_value}" is not a known dimensionally-consistent '
--- a/tests_and_analysis/test/euphonic_test/test_castep_reader.py
+++ b/tests_and_analysis/test/euphonic_test/test_castep_reader.py
@@ -4,6 +4,7 @@ import json
 import pytest
 import numpy as np
 import numpy.testing as npt
+from toolz.dicttoolz import valmap
 
 from euphonic.io import _from_json_dict
 from euphonic.util import direction_changed, mp_grid, get_qpoint_labels
@@ -43,14 +44,15 @@ class TestReadPhononDosData:
                     assert dct[exp_key] == exp_val
 
         # Convert ref data from cm to 1/meV to match current behaviour
-
         expected_dos_data['dos_unit'] = '1/meV'
 
-        for key, value in expected_dos_data['dos'].items():
-            expected_dos_data['dos'][key] = (ureg.Quantity(value, "cm")
-                                             .to("1 / meV", "reciprocal_spectroscopy")
-                                             .magnitude
-                                             .tolist())
+        def transform(dos: list[float]) -> list[float]:
+            return (ureg.Quantity(dos, "cm")
+                    .to("1 / meV", "reciprocal_spectroscopy")
+                    .magnitude
+                    .tolist())
+
+        expected_dos_data["dos"] = valmap(transform, expected_dos_data["dos"])
 
         check_dict(dos_data, expected_dos_data)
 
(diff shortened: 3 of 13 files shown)


ajjackson:
I performed a little speed test of conversions in different contexts. With this definition file:
```
@context reciprocal_spectroscopy = rs
    1 / [frequency] <-> 1 / [length]: 1 / value / speed_of_light
    1 / [energy] -> 1 / [frequency]: planck_constant * value
    1 / [frequency] -> 1 / [energy]: value / planck_constant
    [length] -> 1 / [energy]: value / planck_constant / speed_of_light
    1 / [energy] -> [length]: value * planck_constant * speed_of_light
    1 / [length] -> [energy]: value * planck_constant * speed_of_light
    [energy] -> 1 / [length]: value / planck_constant / speed_of_light
@end

@context bracketed_spectroscopy = bs
    1 / [frequency] <-> 1 / [length]: 1 / (value * speed_of_light)
    1 / [energy] -> 1 / [frequency]: planck_constant * value
    1 / [frequency] -> 1 / [energy]: value / planck_constant
    [length] -> 1 / [energy]: value / (planck_constant * speed_of_light)
    1 / [energy] -> [length]: value * (planck_constant * speed_of_light)
    1 / [length] -> [energy]: value * (planck_constant * speed_of_light)
    [energy] -> 1 / [length]: value / (planck_constant * speed_of_light)
@end
```
and this speed test code:
```python
import timeit

import numpy as np
from pint import UnitRegistry


ureg = UnitRegistry()
ureg.load_definitions("./definitions.txt")

sparse_data = np.random.random((10000, 200))
sparse_data[sparse_data > 0.5] = 0

sparse_spectrum = ureg.Quantity(sparse_data, "1/cm")

print("Default spectroscopy context:")
print(timeit.timeit("sparse_spectrum.to('meV', 'spectroscopy')", globals=globals(), number=1000))

print("Reciprocal spectroscopy context:")
print(timeit.timeit("sparse_spectrum.to('meV', 'reciprocal_spectroscopy')", globals=globals(), number=1000))

print("Reciprocal spectroscopy context with parenthesis groups:")
print(timeit.timeit("sparse_spectrum.to('meV', 'bracketed_spectroscopy')", globals=globals(), number=1000))
```

the output is

```
Default spectroscopy context:
/Users/adam.jackson/micromamba/envs/euphonic-np-test/lib/python3.10/site-packages/pint/facets/plain/quantity.py:1054: RuntimeWarning: divide by zero encountered in divide
  return self.__class__(other_magnitude / self._magnitude, 1 / self._units)
4.869501542067155
Reciprocal spectroscopy context:
3.7165777920745313
Reciprocal spectroscopy context with parenthesis groups:
2.6495610419660807
```

So we see that
- `.to()` is quite expensive
- there is almost a performance factor x2 gained by preparing appropriate definitions

Usually a factor x2 is not a big thing to scream about, but this is so slow that it is a credible bottleneck and worth sorting out.

That being said, things narrow a lot when replacing `.to()` with `.ito()` to convert in-place, so we should look for ways to benefit from that inside Euphonic/Abins:

```
Default spectroscopy context:
/Users/adam.jackson/micromamba/envs/euphonic-np-test/lib/python3.10/site-packages/pint/facets/plain/quantity.py:1054: RuntimeWarning: divide by zero encountered in divide
  return self.__class__(other_magnitude / self._magnitude, 1 / self._units)
0.04182825004681945
Reciprocal spectroscopy context:
0.03280979208648205
Reciprocal spectroscopy context with parenthesis groups:
0.03438079194165766
```

ajjackson:
I think it could be worth refactoring how spectra work; currently they store a numpy array and unit information, and dynamically construct a Quantity when properties are accessed. This is quite expensive, and it would be simpler to store a full Quantity. The current behaviour of "lazily" dealing with unit changes could be retained by deferring an `ito()` to the property getter.

Presumably some features work directly on the private array, and these would have to deal with an extra call to `magnitude`. It seems unlikely that this performance benefit outweight the downsides, but we can do some benchmarking if unsure.

Anyway, this can all be a separate PR!

ajjackson:
This could use another set of eyes before merging. In principle it is straightforward stuff, but messing around with unit conversions scares me a bit. Are there any additional sanity checks or stress tests we can try in order to make sure we didn't accidentally give permission to do anything wrong?

euphonic/validate.py
```diff
@@ -82,7 +82,8 @@ def _check_unit_conversion(obj: object, attr_name: str, attr_value: Any,
     if hasattr(obj, attr_name):
         if attr_name in unit_attrs:
             try:
-                _ = ureg(getattr(obj, attr_name)).to(attr_value)
+                _ = ureg(getattr(obj, attr_name)).to(attr_value,
```
oerc0122, line 85: Does this need to return?

tests_and_analysis/test/euphonic_test/test_castep_reader.py
```diff
@@ -41,12 +41,17 @@ def check_dict(dct, expected_dct):
...
+        # Convert ref data from cm to 1/meV to match current behaviour
+
+        expected_dos_data['dos_unit'] = '1/meV'
+
         for key, value in expected_dos_data['dos'].items():
-            expected_dos_data['dos'][key] = [x/dos_conv for x in value]
+            expected_dos_data['dos'][key] = (ureg.Quantity(value, "cm")
+                                             .to("1 / meV", "reciprocal_spectroscopy")
+                                             .magnitude
+                                             .tolist())
```
oerc0122, line 53: I mean, this is fine, but should it be a dict comprehension?

oerc0122 (approved):
All looks good if it works, just a couple of stylistic questions.

CHANGELOG.rst
```diff
@@ -24,11 +24,31 @@
 
     - Minimum version of threadpoolctl increased from 1.0 to 3.0.
 
-- Big fixes
```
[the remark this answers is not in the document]
ajjackson: 😆

euphonic/validate.py
ajjackson: By that, do you mean "why have the `_ = ...` or do you mean "this could be `ito` as we never use the result." ?

I think the `_ =` is just to clarify that this is being used for a side-effect. Do you find that distracting?

Performance seems fairly identical between `to` and `ito` here.

oerc0122: Yeah, I mean why the `_ =`. To me it reads as a bit confusing as though the functionality might change if it doesn't return, but maybe that's my matlab-destroyed brain talking where that happens all the time.

tests_and_analysis/test/euphonic_test/test_castep_reader.py
ajjackson: It could be, but
- performance difference is irrelevant, this is a test
- readability is already an issue here, dict comprehension would make it worse, I think?

```python
        expected_dos_data['dos'] = {
            key: ureg.Quantity(value, "cm")
                    .to("1 / meV", "reciprocal_spectroscopy")
                    .magnitude
                    .tolist()
            for key, value in expected_dos_data["dos"].items()
        }

```

I think the nicest option is 

```python
        from toolz.dicttoolz import valmap

        def transform(dos: list[float]) -> list[float]:
            return (ureg.Quantity(dos, "cm")
                    .to("1 / meV", "reciprocal_spectroscopy")
                    .magnitude
                    .tolist())

        expected_dos_data["dos"] = valmap(transform, expected_dos_data["dos"])
```

but toolz is not yet a Euphonic dependency. (I'd like to start using it for some other stuff, and Mantid team seems to be open to this. I suppose at this point it would just be a testing dependency.)

Without toolz there is

```python
        def transform(dos: list[float]) -> list[float]:
            return (ureg.Quantity(dos, "cm")
                    .to("1 / meV", "reciprocal_spectroscopy")
                    .magnitude
                    .tolist())

        expected_dos_data["dos"] = {
            key: transform(value)
            for key, value in expected_dos_data["dos"].items()}
```

or

```python
        def transform(dos_item: tuple[str, list[float]]) -> list[float]:
            key, dos = dos_item
            return key, (ureg.Quantity(dos, "cm")
                         .to("1 / meV", "reciprocal_spectroscopy")
                         .magnitude
                         .tolist())

        expected_dos_data["dos"] = dict(
            map(transform, expected_dos_data["dos"].items()))
```

euphonic/validate.py
ajjackson: As far as I know that can't happen in Python! But to make a tenuous Matlab connection, this pattern is more often seen in Jupyter notebooks or interactive terminal to suppress the return value from being printed (like including a `;` in Matlab).

tests_and_analysis/test/euphonic_test/test_castep_reader.py
oerc0122: Yeah, doesn't make much difference. Wish `toolz` become somewhat core like `itertools` and things. At least dict key/value maps.

ajjackson: Well, as I do ultimately intend to use it in the main Euphonic code, perhaps it's not a bad idea to add it to the test requirements now and see where else it can sneak into use.

PR #312, opened 2024-07-31
Fixing build/test failures

ajjackson:
Tests are broken overnight at build stage, despite no change to euphonic code. The setup.py build can no longer find Numpy, which has clearly been installed beforehand.

Could be a setuptools update? A lot of projects have been burned by removal of `setuptools.command.test` but we don't use that.


ajjackson:
Pinning to <72 has not helped. There is something interesting in the error message, though

```
Building wheels for collected packages: euphonic
  Building wheel for euphonic (pyproject.toml): started
  Building wheel for euphonic (pyproject.toml): finished with status 'error'
  error: subprocess-exited-with-error
  
  × Building wheel for euphonic (pyproject.toml) did not run successfully.
```

Why does this mention pyproject.toml when Euphonic does not use such a thing?

ajjackson:
Ok, pinning both setuptools and wheel versions seems to do the trick.

At some point it would be nice to rework the configuration to use pyproject.toml, and that might work with a newer setuptools. But that's a big job in its own right and not needed to get this working again for now.

PR #314, opened 2024-09-04
Update github action versions in test_release

ajjackson:
The latest version of download-artifact v4 includes an important security fix.

Version numbers of upload- and download- artifact are supposed to match.

Other actions are already using these versions but this one was overlooked.


ajjackson:
The test_release action seems to be loading and running correctly. The tests themselves are failing but that's due to incompatibility between the master-branch testing/build and the last release; this is expected.

PR #318, opened 2024-10-01
Implement web JSON export

### LINKED ISSUE #299: Write to phonon website JSON
Requested by ISIS beamline scientist Matthew Krzystyniak:

the [Phonon visualisation website](http://henriquemiranda.github.io/phononwebsite/) is a very useful tool for analysis of phonon dispersion. As input it requires a .json file with specific formatting; the data content is more or less the same as a QpointPhononModes.

While it is possible to generate these from Castep or Phonopy outputs with Sumo (using the writer from Pymatgen) this is a rather heavy dependency and not otherwise used much in a typical ISIS workflow.

It would be nice if Euphonic provided a simple command-line tool that converted force constants to these .json files, to aid visualisation during a phonon study. This could be implemented with
- a `QpointPhononModes.write_phonon_website_json(fp)` method in the Python API, and
- a simple way of calling this method from `euphonic-dispersion` or a new script.

ajjackson:
Closes #299 

Current status: it basically works but there is some funny business on the x-axis. The phonon website doesn't directly support breaks in the x-axis, so we should clean them up:
- [X] Merge the neighbouring symbols e.g. `["X", "Y"]` becomes `["X|Y"]`
- [X] Remove duplicate q-points at direction changes e.g. `["X", "X"]` becomes `["X"]`
- [X] use the line_breaks section to restart the line
- [X] Remove the empty section of "distance"


ajjackson:
I think the mass-weighting may also need tweaking, the acoustic modes don't quite look right!

ajjackson:
It looks like the last point of each segment is dropped: by "touching" segments we get correct-looking discontinuities except at the LO-TO-split Gamma

Screenshot 2024-10-11 at 17 49 29

```
232047   ▽ line_breaks: (4) [[…], […], […], […]]
232048     ▽ [0]: (2) [0, 127]
232049         [0]: 0
232050         [1]: 127
232052     ▽ [1]: (2) [127, 275]
232053         [0]: 127
232054         [1]: 275
232056     ▽ [2]: (2) [275, 299]
232057         [0]: 275
232058         [1]: 299
232060     ▽ [3]: (2) [299, 323]
232061         [0]: 299
232062       ▶ [1]: 323
```

ajjackson:
Screenshot 2024-10-11 at 18 05 26
Results finally look as they should 🎉 

Just a bit of cleanup needed to put things in sensible places with good names.

euphonic/qpoint_phonon_modes.py
```diff
@@ -716,3 +746,171 @@ def from_phonopy(cls: Type[T], path: str = '.',
...
+        """
+
+        with open(output_file, 'w') as fd:
+            json.dump(self._to_phonon_website_dict(name=name,
+                                                   x_tick_labels=x_tick_labels),
+                      fd)
+
+    @staticmethod
+    def _crystal_website_data(crystal: Crystal) -> dict[str, Any]:
+        elements = [
```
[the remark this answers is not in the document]
ajjackson: No, Euphonic ships its own cross section data.

This package _is_ available on conda-forge and provides its own neutron cross-section data so it might be interesting to incorporate. I think it's overkill to add that dependency for this single lookup of element symbols, though.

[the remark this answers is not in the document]
oerc0122: Yeah. Not for this PR for sure, but if you were using it, this would be a place to get around redefining things.

euphonic/qpoint_phonon_modes.py
```diff
@@ -716,3 +746,171 @@ def from_phonopy(cls: Type[T], path: str = '.',
...
+          -->
+
+          [(1, "X"), (4, "A"), (7, "Y|Z")]
+
+        """
+        labels = dict(x_tick_labels)
+
+        for index in sorted(labels):
+            if index - 1 in labels:
+                if labels.get(index - 1) != labels.get(index):
```
oerc0122, line 870: Realise what this is doing but the logic here makes me unsure about precisely how it is doing it.

euphonic/qpoint_phonon_modes.py
```diff
@@ -716,3 +746,171 @@ def from_phonopy(cls: Type[T], path: str = '.',
...
+        return breakpoints.tolist()
+
+    @staticmethod
+    def _expand_duplicates(distances: np.ndarray, pad_fraction = 0.001) -> list[int]:
+        diff = np.diff(distances)
+        pad = np.median(diff) * pad_fraction
+
+        duplicates = np.where(diff == 0.)[0] + 1
+        for duplicate in reversed(duplicates):
+            distances[duplicate:] += (distances[duplicate - 1] - distances[duplicate] + pad)
```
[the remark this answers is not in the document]
ajjackson: Good point! Would be a lot easier to read

euphonic/qpoint_phonon_modes.py
```diff
@@ -716,3 +746,171 @@ def from_phonopy(cls: Type[T], path: str = '.',
...
+        median = np.median(diff)
+        breakpoints = np.where((diff / median) > btol)[0] + 1
+
+        for breakpoint in reversed(breakpoints):
+            distances[breakpoint:] -= (distances[breakpoint] - distances[breakpoint - 1])
+
+        return breakpoints.tolist()
+
+    @staticmethod
+    def _expand_duplicates(distances: np.ndarray, pad_fraction = 0.001) -> list[int]:
```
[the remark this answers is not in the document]
ajjackson: Noted. It may also be the case that the expansion is unnecessary, now that the other issues have been fixed.

euphonic/cli/utils.py
```diff
@@ -934,3 +934,9 @@ def _compose_style(
 
     style.append(explicit_args)
     return style
+
+def _get_title(filename: str, title: str = '') -> str:
+    """Get a plot title: either user-provided string, or from filename"""
+    if title:
+        return title
+    return pathlib.Path(filename).stem
```
oerc0122, line 942: Warrant using a ternary if? 

Would it be better to use `None`? Is there a reason a user might want to have an empty title?

oerc0122 (commented):
Like the restructure. Much neater.

euphonic/qpoint_phonon_modes.py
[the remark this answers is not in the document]
ajjackson: It turns out this distance manipulation can be removed entirely! I though it was a necessary kludge for the plot rendering, but now things seem to work fine without it.

euphonic/qpoint_phonon_modes.py
ajjackson: Not really sure how to clarify, this looks like over-explaining:
```python
def _combine_neighbouring_labels(x_tick_labels: XTickLabels) -> XTickLabels:
    """Merge neighbouring labels in x_tick_label list

    If labels are the same, only keep one.

    If labels are different, join with |

    e.g.::

      [(1, "X"), (2, "X"), (4, "A"), (7, "Y"), (8, "Z")]

      -->

      [(1, "X"), (4, "A"), (7, "Y|Z")]

    """
    labels = dict(x_tick_labels)

    for index in sorted(labels):
        if index - 1 in labels:
            if labels.get(index - 1) != labels.get(index):
                # Neighbouring labels are different: merge second into first
                labels[index - 1] = f"{labels[index - 1]}|{labels[index]}"
            # Remove second label of pair
            del labels[index]
    return sorted(labels.items())
```

oerc0122: I think it's more why 
```python
if index - 1 in labels:
```
 defines them as "the same". That's the logic I can't follow, that only implies the neighbouring point still exists. If the *value* were the same rather than the key, then I could see removing the second label pair. But every time?

ajjackson: I'd expect it to produce `[(1, "X|Y"), (3, "X")]` which isn't brilliant but is ok given that this was a terrible input.

ajjackson: I'll see if I can rewrite more obviously using `pairwise`

oerc0122: Ok, so that isn't what I expected. I was thinking it would leave things unchanged. But I guess now that I look again, it makes more sense. 

I think it's actually ok, but I think I would add a third part to the `if`: neighbours (same), neighbours (different), non-neighbours.

euphonic/cli/utils.py
oerc0122: Ok, I would say that's a different issue and if this is the behaviour elsewhere that's fine. May want to make an issue even if it just sits there.

ajjackson: The fix required fewer changes than I expected, let's give it a try

euphonic/qpoint_phonon_modes.py
ajjackson: How about this, which does a slightly different but acceptable thing `[(3, "X|Y|X")]` and might be more readable:

```python
def _combine_neighbouring_labels(x_tick_labels: XTickLabels) -> XTickLabels:
    """Merge neighbouring labels in x_tick_label list

    If labels are the same, only keep one.

    If labels are different, join with |

    e.g.::

      [(1, "X"), (2, "X"), (4, "A"), (7, "Y"), (8, "Z")]

      -->

      [(1, "X"), (4, "A"), (7, "Y|Z")]

    """
    labels = dict(x_tick_labels)

    for (index, next_index) in pairwise(sorted(labels)):

        if index + 1 == next_index:  # Labels are neighbours

            if labels[index] != labels[next_index]:  # Combine differing labels
                labels[next_index] = f"{labels[index]}|{labels[next_index]}"

            del labels[index]  # Drop redundant label

    return sorted(labels.items())

```

ajjackson:
Not sure what's going on with the test failure; I can reproduce it locally but only within `tox`. Should get there in the end...

euphonic/qpoint_phonon_modes.py
oerc0122: Much clearer to me about the intent and the function.

ajjackson:
... got there in the end.

euphonic/cli/utils.py
```diff
@@ -671,7 +671,7 @@ def __call__(self, parser, args, values, option_string=None):
         section.add_argument(
             '-s', '--save-to', dest='save_to', default=None,
             help='Save resulting plot to a file with this name')
-        section.add_argument('--title', type=str, default='',
+        section.add_argument('--title', type=str, default=None,
```
oerc0122, line 674: Just to double check `default` bypasses `type`. I think it does, but you've checked this works.

euphonic/writers/phonon_website.py
```diff
@@ -0,0 +1,218 @@
+"""Export to JSON for phonon visualisation website"""
```
oerc0122, line 1: I know it's throughout the document, but the one place I think the link to "phonon visualisation website" should be would be here. I understand that others may come along in which case I would recommend:
```
"""Export to JSON for phonon visualisation websites:

- website 1
...
"""
```

euphonic/writers/phonon_website.py
```diff
@@ -0,0 +1,218 @@
...
+
+    A list of identified breakpoints is returned
+
+    """
+    diff = np.diff(distances)
+    median = np.median(diff)
+    breakpoints = np.where((diff / median) > btol)[0] + 1
+
+    for breakpoint in reversed(breakpoints):
+        distances[breakpoint:] -= (distances[breakpoint]
```
oerc0122, line 129: Not using `diff`? It this because it's the backwards diff?

euphonic/writers/phonon_website.py
```diff
@@ -0,0 +1,218 @@
...
+
+def _find_duplicates(distances: np.ndarray) -> list[int]:
+    """Identify breakpoints where a q-point is repeated in list
+
+    Return the higher index of two points: this works more nicely with phonon
+    website interpretation of line break points.
+    """
+
+    diff = np.diff(distances)
+    duplicates = np.where(diff == 0.)[0] + 1
```
oerc0122, line 143: Don't compare `float`s. I would use [`isclose`](https://numpy.org/doc/2.0/reference/generated/numpy.isclose.html)

euphonic/writers/phonon_website.py
```diff
@@ -0,0 +1,218 @@
...
+    qpts = modes.qpts
+    eigenvectors = modes.eigenvectors
+
+    abscissa = _calc_abscissa(modes.crystal.reciprocal_cell(), qpts)
+
+    duplicates = _find_duplicates(abscissa)
+    breakpoints = _remove_breaks(abscissa)
+
+    breakpoints = sorted(set([0] + duplicates + breakpoints + [len(abscissa)]))
+    line_breaks = [(start, end) for start, end in pairwise(breakpoints)]
```
oerc0122, line 192: Is this not `list(pairwise(breakpoints))`?

oerc0122 (commented):
Couple of queries. Looks functionally good.

euphonic/cli/utils.py
ajjackson: I've used this pattern many times before so it _seems_ to work sensibly.

euphonic/writers/phonon_website.py
```diff
@@ -0,0 +1,218 @@
...
+    qpoints: list[list[float]]
+    distances: list[float]  # Cumulative distance from first q-point
+    eigenvalues: list[float]  # in cm-1
+    vectors: list[list[list[tuple[ComplexPair, ComplexPair, ComplexPair]]]]
+    line_breaks: list[tuple[int, int]]
+
+
+def write_phonon_website_json(
+    modes: QpointPhononModes,
+    output_file: str = "phonons.json",
```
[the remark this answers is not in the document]
ajjackson: I need to do a wider cleanup to support Path in euphonic consistently. But now that this is in its own module I don't mind the "inconsistency" of having a Path option here.

euphonic/writers/phonon_website.py
ajjackson:
```python
    for breakpoint in reversed(breakpoints):
        distances[breakpoint:] -= diff[breakpoint - 1]
```
seems to work correctly

euphonic/writers/phonon_website.py
ajjackson: The float comparison _should_ really work in this context, but it does seem a good habit to avoid it.

euphonic/writers/phonon_website.py
ajjackson: Yes, some numerical offsets have been changed/removed over time 😆

oerc0122 (approved):
Once it's passing and you're happy, think this is ready to go!

ajjackson:
It still needs a couple of proper tests!

PR #321, opened 2024-10-22
Use pytest-xdist for all tests within tox

ajjackson:
The pytest-xdist plugin allows pytest to distribute the tests to "workers". Here we have quite an aggressive configuration:

- use as many workers as there are available CPUs.

  - This should be beneficial as we have quite a large number of short/fast tests to spread over workers.

- when a worker has run out of jobs, allow it to "steal" remaining jobs from other workers

  - This should be beneficial as there are a few tests which are much slower, and we don't want them to run in series on the same worker while others sit idle.

We might need to do some funny business and let certain tests run independently... But so far this works pretty well on my laptop.

### DIFF
--- a/tests_and_analysis/tox_requirements.txt
+++ b/tests_and_analysis/tox_requirements.txt
@@ -1,6 +1,6 @@
 mock
 pytest~=7.0
-coverage
+pytest-cov
 pytest-mock
 pytest-lazy-fixture
 pytest-xdist
--- a/tests_and_analysis/tox_requirements.txt
+++ b/tests_and_analysis/tox_requirements.txt
@@ -3,6 +3,7 @@ pytest~=7.0
 coverage
 pytest-mock
 pytest-lazy-fixture
+pytest-xdist
 pytest-xvfb
 python-slugify
 toolz
--- a/tests_and_analysis/test/run_tests.py
+++ b/tests_and_analysis/test/run_tests.py
@@ -113,18 +113,27 @@ def _build_pytest_options(reports_dir: str, do_report_tests: bool,
     options: list[str] = [tests]
     # Add reporting of test results
     if do_report_tests:
+        options.append("--cov=euphonic")
+        options.append("--cov-append")
+
         # We may have multiple reports, so get a unique filename
         filename_prefix = "junit_report"
         filenum = int(time.time())
+
         junit_xml_filepath = os.path.join(
             reports_dir, f"{filename_prefix}_{filenum}.xml")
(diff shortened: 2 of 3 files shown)


ajjackson:
It looks like some funny business is needed to get xdist and coverage to play together nicely 😢

ajjackson:
It looks like the `pytest-cov` plugin exists, in part, to make that work. But necessitates some other fiddling. It should make things cleaner in the end...

ajjackson:
I'm seeing some improvement here but it is a lot less spectacular than when running locally (M3 macbook with 12 cores).

ajjackson:
Despite seemingly generating and uploading a proper coverage xml file, something is going wrong with the coverage process.

Screenshot 2024-10-23 at 10 00 41

ajjackson:
It could be related to
```
[2024-10-23T08:48:35.375Z] ['error'] There was an error running the uploader: Error uploading to [https://codecov.io:](https://codecov.io/) Error: There was an error fetching the storage URL during POST: 429 - {"message":"Rate limit reached. Please upload with the Codecov repository upload token to resolve issue. Expected time to availability: 169s."}
```
or
```
test (ubuntu-latest)
The following actions use a deprecated Node.js version and will be forced to run on node20: codecov/codecov-action@v3. For more info: https://github.blog/changelog/2024-03-07-github-actions-all-actions-will-run-on-node20-instead-of-node16-by-default/
```

tests_and_analysis/tox_requirements.txt
```diff
@@ -1,8 +1,9 @@
 mock
 pytest~=7.0
-coverage
+pytest-cov
```
oerc0122, line 3: I presume this is the switched module you were talking about.

oerc0122 (approved):
All looks sensible if it works. Is this in conjunction with or in contrast to #322 ?

ajjackson:
> All looks sensible if it works. Is this in conjunction with or in contrast to #322 ?

In contrast. It looks like for multi-python-version testing the tox parallelism is better, but for a quick local run of the test suite without building `xdist` can be quite nice.

Perhaps it is worth keeping the `run_tests.py --parallel` option from this PR, but not use it in the Github actions? Then again, it's not a lot of trouble to call `pytest` directly in those cases.

ajjackson:
Closing in favour of `tox --parallel` #322 ; it doesn't make sense to combine them when Github runners only have four cores. Consider revisiting if we start running CI on more parallel cores.

PR #322, opened 2024-10-23
run_tests workflow: use parallel tox

ajjackson:
Alternative to #321 : parallelism at the `tox` level

Let's see if we get a more impressive speedup by distributing the tox environments over separate processors.

Depends on #324

### DIFF
--- a/.github/workflows/run_tests.yml
+++ b/.github/workflows/run_tests.yml
@@ -72,12 +72,16 @@ jobs:
         if: github.event_name != 'workflow_dispatch'
         env:
           TOX_SKIP_ENV: '.*?(py311).*?'
+          TOX_PARALLEL_NO_SPINNER: 1
         shell: bash -l {0}
-        run: python -m tox
+        run: python -m tox --parallel
+
       - name: Run tests, workflow dispatch so test all Python versions
         if: github.event_name == 'workflow_dispatch'
+        env:
+          TOX_PARALLEL_NO_SPINNER: 1
         shell: bash -l {0}
-        run: python -m tox
+        run: python -m tox --parallel
 
       - name: Upload test results
         if: always()


ajjackson:
Well, in the initial naive attempt the Mac performance is great (5mins) and the Win/Linux runs have build problems, probably related to race conditions while the parallel process build Euphonic binaries.

ajjackson:
I suspect that migration to pyproject.toml and tox 4 would solve this problem by better-isolating the builds. But a workaround to try before then is to build the tox environments in series and then run them in parallel.

ajjackson:
This is odd, Windows is failing to build even in series now?
```
   D:\a\Euphonic\Euphonic\c\_euphonic.c : fatal error C1083: Cannot open compiler generated file: 'D:\a\Euphonic\Euphonic\build\temp.win-amd64-cpython-310\Release\c\_euphonic.obj': Permission denied
  error: command 'C:\\Program Files\\Microsoft Visual Studio\\2022\\Enterprise\\VC\\Tools\\MSVC\\14.41.34120\\bin\\HostX86\\x64\\cl.exe' failed with exit code 1
  ```

ajjackson:
The problem is that Euphonic is installed in `pre_command`, not `install_command`, and `pre_command` doesn't run under `--notest`.

ajjackson:
I'm seeing some elaborate workaround options, but propose that this is put on hold until the project is migrated to a better-isolated build process with pyproject.toml, and the test-env framework migrated to tox 4. It may be that under those conditions _all we need to do_ is add `--parallel`.

@oerc0122 has taken an initial look at migration to pyproject.toml and hit a few (likely surmountable) problems, so we should assume it will take some effort to finish that.

ajjackson:
Have tried reimplementing this on #324

ajjackson:
One Windows test failure; a few values in an array on Python 3.12. Not sure how to attack that 😞

oerc0122:
Is it due to list order? Or is it actually a failure? 

Are the tests wholly independent or are they writing to files. Temp files struggle on Windows systems due to how Windows resolves and updates its filesystem, could that be an issue?

ajjackson:
Oh no, it passed

ajjackson:
Windows also looking happy now that I've re-run including Python 3.11

oerc0122 (approved):

ajjackson:
Presumably we don't expect this will cause trouble for tox 4 migration?

oerc0122:
> Presumably we don't expect this will cause trouble for tox 4 migration? #333

Should be fine, may have to change Tox4 call in the CI ([see here](https://tox.wiki/en/latest/cli_interface.html#tox-run-parallel-(p))) but that's trivial and I think this has priority pre-release.

Issue #323, opened 2024-10-30
Overhaul configuration/build system


ajjackson:
The Python world is generally moving towards isolated builds with Pyproject.toml specifications; Euphonic currently uses a setup.py file with a mixture of metadata and C compiling details.

@oerc0122 has suggested that Meson could be a useful build tool. As it explicitly supports building against OpenMP and Numpy, we might be able to eliminate some of the logic related to these. Conveniently, it can be pip-installed and so listed as a build-time dependency in the configuration.

ajjackson:
This has been done in #324

PR #324, opened 2024-10-31
Pyproject toml

oerc0122:
- Migration to `pyproject.toml` with `meson`.
- Migrated away from `versioneer` to custom `version.py` script.
- Migrated CI workflows to use [`cibuildwheel`](https://pypi.org/project/cibuildwheel/)
- Move utilities related to build into `<root>/build_utils`.
- Updated testing workflows to add appropriate compilers. 

Addresses #323


ajjackson:
This is looking really promising! It's nice having the compilation stuff in the meson config file and then a fairly clean toml for main project.

We may have to ditch versioneer; will be worth seeing what other meson projects are using?

ajjackson:
Have a look at this branch:
- I rebased on master (only tweak needed really was to add toolz to pyproject.toml as it was in the desynced setup.py)
- Hacked out versioneer
- Added a lot of boilerplate to meson.build so that it includes all they Python source files

The last part seems to be essential but looks pretty ugly. I think the preferred pattern is to cascade it into more meson.build files and load them with the `subdir` directive?

Anyway it seems to build and `euphonic-dos` runs so that's a good start.

.github/workflows/run_tests.yml
```diff
@@ -17,37 +17,39 @@ jobs:
...
-        if: startsWith(matrix.os.name, 'macos')
+        if: startsWith(matrix.os, 'macos')
         run: |
           brew install llvm
-          echo CC="${{ matrix.os.llvm }}/bin/clang" >> $GITHUB_ENV
-          echo LDFLAGS="-L${{ matrix.os.llvm }}/lib $LDFLAGS" >> $GITHUB_ENV
-          echo CPPFLAGS="-I${{ matrix.os.llvm }}/include $CPPFLAGS" >> $GITHUB_ENV
+          echo CC="${HOMEBREW_PREFIX}/opt/llvm/bin/clang" >> $GITHUB_ENV
+          echo LDFLAGS="-L${HOMEBREW_PREFIX}/opt/llvm/lib $LDFLAGS" >> $GITHUB_ENV
+          echo CPPFLAGS="-I${HOMEBREW_PREFIX}/opt/llvm/include $CPPFLAGS" >> $GITHUB_ENV
```
oerc0122, line 44: Could use `env:` here to inject the constants for cleanliness?

ajjackson: Is there a way to do matrux-dependent env? That would be ideal

ajjackson: Looks like it is possible to put some logic in env: but it has to be done with && / || operators https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/evaluate-expressions-in-workflows-and-actions

oerc0122: How about just:
```sh
          LLVM_DIR=${HOMEBREW_PREFIX}/opt/llvm
          echo CC="${LLVM_DIR}/bin/clang" >> $GITHUB_ENV
          echo LDFLAGS="-L${LLVM_DIR}/lib $LDFLAGS" >> $GITHUB_ENV
          echo CPPFLAGS="-I${LLVM_DIR}/include $CPPFLAGS" >> $GITHUB_ENV
```

ajjackson: I went for a blend of these 😄

meson.build
oerc0122, line 1: Do we need more comments/docs in this or are you happy as is?

ajjackson:
Yes, it looks like it.

The remaining action is to update the other (release build and test) workflows to get their environment variables in order too.

meson.build
ajjackson: It seems self-explanatory, unless there are any "gotchas" that should be highlighted

.github/workflows/build_upload_pypi_wheels.yml
```diff
@@ -38,14 +38,29 @@ jobs:
...
+        run: |
+          echo "CC=${MSVC_BIN}\cl.exe" >> $GITHUB_ENV
+          echo "CC_LD=${MSVC_BIN}\link.exe" >> $GITHUB_ENV
+      - name: Update Python pip, build, wheel, and twine
+        shell: bash -l {0}
+        run: |
+          python -m pip install --upgrade pip build wheel twine
       - name: Build Python wheel
 
         if: matrix.os != 'ubuntu-latest'
```
oerc0122, line 66: Is there any reason why `cibuildwheel` isn't being used on other platforms?

ajjackson: Excellent question. I never looked at this part closely, assumed it was just a tool that handles the "manylinux" complications.

ajjackson: I note that the CIBW_REPAIR_WHEEL_COMMAND hook is intended for things like `delvewheel`, if we end up needing those particular Windows shenanigans.

ajjackson:
Currently the CIBW tests are "installing the wheel" but failing to import `_euphonic`. Oddly, the configuration has
```          CIBW_TEST_COMMAND: pytest {package}/tests_and_analysis/test```

But in the output
```
       + mkdir -p /tmp/tmp.fqvqo48S0J/test_cwd
      + sh -c 'pytest /project/tests_and_analysis/test'
```

So "project" and "package" seem to be the same?

**Comment by ajjackson** [comment]
Perhaps the next troubleshooting step is to skip the tests, upload some wheels as Github artifacts and try installing them locally.

**Comment by ajjackson** [comment]
Looks like run_tests is happy again, does that help us with the wheel build situation?

**Comment by oerc0122** [comment]
> Looks like run_tests is happy again, does that help us with the wheel build situation?

Unfortunately that's only because I've reverted back to the hardcoded way of finding the MSVC

**Comment by ajjackson** [comment]
Hmm, so the environment is reliable on run_tests but not in the cibuildwheel environment?

Both of those use miniconda to install Python. But not for anything else? Maybe we can lose Conda to simplify the environment a bit.

File: build_utils/version.py [hunk 3]
```diff
@@ -0,0 +1,70 @@
...
+
+Provides 3 means of updating:
+
+- python version.py
+
+  Print computed version to stdout
+
+- python version.py --dump
+
+  Update version in `euponic/version.py`
```
ajjackson, line 11:
```suggestion
  Update version in `euphonic/version.py`
```

ajjackson:
Some reflections on the overall process that got us here:
- This was quite painful and took more time than expected
  - None of the problems can be blamed on pyproject.toml / PEP517 / PEP518; they do actually simplify things. The trouble is finding the correct configuration for the "simple" way of doing things...
- We had hoped that by using (python-)meson the platform-dependent code would be pushed upstream and we wouldn't need to worry about it any more. In practice what happened is that platform-dependent fiddling was moved from the setup.py to the Github runners. That _should_ be good news as it means we are less tied to particular build environments and e.g. a Mac user should now be able to build against non-Homebrew OpenMP if they can obtain it. 
- But this necessitated another round of work on the Github Actions. Troubleshooting Github Actions is a bit soul-destroying because it can take a lot of cursing and iteration to get to the correct "simple, obvious-looking" result.
- There were some gotchas with the Github-provided Windows runners:
  - after activating the msvc environment, we got an inconsistent `cl.exe` and `link.exe` on the PATH, because a "Git" directory is given priority and ships with GNU link.exe
  - the solution is to use absolute paths, but then we discovered that different runners can have slightly different MSVC versions leading to flaky builds/tests. We have ended up with some compromise code that makes a few assumptions and tries to locate a reasonable MSVC installation before setting CC and CC_LD to absolute paths. It will need updating when the runners are significantly changed.
- I'm puzzled that we had to dig so much and invent so much process/machinery to handle the task of versioning sdist and wheel built with python-meson. It seems like something "everyone" needs to do, yet the information is scattered around a bit. As far as I can see, "everyone" has either a) something simpler/less-capable than this (i.e. no "dirty" versioning) or b) something even more complicated (that suits their complicated project.) Perhaps it's worth a blog post or something to help people (i.e. our future selves) navigate our solution and the surrounding issues; but I suspect neither of us wants to look at this again for a while.
- Making PRs from forks is usually a pretty painless development workflow, but when applied to Github Actions it is _awful_. PRs would not trigger test runs; we had to post links to action runs on the other repository; discrepancies in the git tags gave misleading results.

In future if contributing CI tweaks from a fork I would make a PR _to another branch on the fork_ for purposes of discussion and automatic CI.

mducle:
> Troubleshooting Github Actions is a bit soul-destroying

Yes, totally agree... I found that using [tmate-actions](https://github.com/mxschmitt/action-tmate) was helpful but it's still quite a painful process...

ajjackson (approved):
At this stage:
- test CI (here) and package-building CI (on the fork) look as expected
- test_release.yml hasn't been updated but that can be a separate PR. It will probably need further revision when the release is actually made...

I think it's a good time to merge to master and see what happens.

ajjackson:
Ah, one thing to look at: we currently document an option to

```
pip install --install-option="--python-only" euphonic
```

https://euphonic.readthedocs.io/en/stable/installation.html#installing-euphonic-without-the-c-extension

I don't think that syntax works any more?

oerc0122:
The command is now:
```sh
pip install -Csetup-args="-Dpython_only=true" euphonic
```
Not sure if that'll work from PyPI though

ajjackson:
That doc update looks great! And the command behaves as expected on my laptop.

PR #328, opened 2024-11-18
Faster indexing/iteration for spectrum collections

ajjackson:
Iterating on SpectrumCollection objects is currently quite expensive, which limits the feasibility of processing them with functional-programming-style filters, pipelines etc. Currently in Abins prototypes I am using workarounds that access or construct raw numpy arrays. It would be nice if that wasn't necessary!

So far we just have a proof of concept: instead of constructing with the relatively slow `__init__()` method, we can use `__new__()` and bypass the safety checks. As we are constructing from known-good data this _should_ be ok...

### DIFF
--- a/euphonic/spectra.py
+++ b/euphonic/spectra.py
@@ -739,15 +739,9 @@ class SpectrumCollectionMixin(ABC):
     def _get_internal_spectrum_data_unit(self) -> str:
         return getattr(self, f"_internal_{self._spectrum_data_name()}_unit")
 
-    def _get_bin_kwargs(self) -> Dict[str, Quantity]:
-        """Get constructor args for bin axes from current data
-
-        e.g. for Spectrum2DCollection this is
-
-            {"x_data": self.x_data, "y_data": self.y_data}
-        """
-        return {f"{axis}_data": getattr(self, f"{axis}_data")
-                for axis in self._bin_axes}
+    @property
+    def _bin_axis_names(self) -> list[str]:
+        return [f"{axis}_data" for axis in self._bin_axes]
 
     @classmethod
     def _get_item_data(cls, item: Spectrum) -> Quantity:
@@ -780,8 +774,12 @@ class SpectrumCollectionMixin(ABC):
             np.sum(self._get_raw_spectrum_data(), axis=0),
             units=self._get_internal_spectrum_data_unit()
         ).to(self._get_spectrum_data_unit())
+
+        bin_kwargs = {axis_name: getattr(self, axis_name)
+                      for axis_name in self._bin_axis_names}
+
         return self._item_type(
-            **self._get_bin_kwargs(),
+            **bin_kwargs,
             **{self._spectrum_data_name(): summed_s_data},
             x_tick_labels=copy.copy(self.x_tick_labels),
             metadata=metadata
@@ -828,6 +826,20 @@ class SpectrumCollectionMixin(ABC):
         else:
             spectrum = self.__new__(type(self))
 
+        self._set_item_data(spectrum, item)
+
+        spectrum.x_tick_labels = self.x_tick_labels
+        spectrum.metadata = self._get_item_metadata(item)
+
+        return spectrum
+
+    def _set_item_data(
+            self,
+            spectrum: Spectrum,
+            item: Union[Integral, slice, Sequence[Integral], np.ndarray]
+    ) -> None:
+        """Write axis and spectrum data to a new Spectrum"""
+
         for axis in self._bin_axes:
             setattr(spectrum, f"_{axis}_data",
                     getattr(self, f"_{axis}_data").copy())
@@ -843,10 +855,7 @@ class SpectrumCollectionMixin(ABC):
         setattr(spectrum, f"{self._spectrum_data_name()}_unit",
                 self._get_spectrum_data_unit())
 
-        spectrum.x_tick_labels = self.x_tick_labels
-        spectrum.metadata = self._get_item_metadata(item)
 
-        return spectrum
 
     def _validate_item(self, item: Integral | slice | Sequence[Integral] | np.ndarray
                        ) -> None:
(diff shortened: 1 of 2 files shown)


ajjackson:
This seems very promising, on _euphonic/tests_and_analysis/test/data/spectrum1dcollection/La2Zr2O7_666_coh_pdos.json_ I am seeing a better-than-10x speedup on `list(spectrum)`.

euphonic/spectra.py
```diff
@@ -821,16 +819,44 @@ def __getitem__(
...
+        __new__() from the parent SpectrumCollection.
+        """
+
+        for axis in self._bin_axes:
+            setattr(spectrum, f"_{axis}_data",
+                    getattr(self, f"_{axis}_data").copy())
+            setattr(spectrum, f"_internal_{axis}_data_unit",
+                    getattr(self, f"_internal_{axis}_data_unit"))
+            setattr(spectrum, f"{axis}_data_unit",
+                    getattr(self, f"{axis}_data_unit"))
```
oerc0122, line 852:
```suggestion
            for prop in ("_{}_data", "_internal_{}_data_unit", "{}_data_unit"):
                curr_prop = prop.format(axis)
                setattr(spectrum, curr_prop, getattr(self, curr_prop).copy())
```
?

euphonic/spectra.py
```diff
@@ -821,16 +819,44 @@ def __getitem__(
...
+
+        setattr(spectrum, self._spectrum_raw_data_name(),
+                self._get_raw_spectrum_data()[item, :].copy())
+        setattr(spectrum, f"_internal_{self._spectrum_data_name()}_unit",
+                self._get_internal_spectrum_data_unit())
+        setattr(spectrum, f"{self._spectrum_data_name()}_unit",
+                self._get_spectrum_data_unit())
```
oerc0122, line 859: Is the first one right? The fact it uses `spectrum_raw_data_name` rather than `spectrum_data_name` makes me concerned this will get visually confusing.

ajjackson: Copying the raw data is the _whole point_ of doing this. When we access the actual "spectrum data" (i.e. `y_data` in 1D or `z_data` in 2D) it does some expensive unit-handling. Here we just copy strings and numpy arrays.

oerc0122: I don't object to that, I just think that when reading the code quickly, it'll be easy to miss that that's a different structure.

euphonic/spectra.py
oerc0122: It was mostly just an "is this more legible" thing anyway. It's fine as is, just a little visually noisy.

ajjackson: Maybe! Looks like `copy` on str is a no-op as it should be:

```python
In [6]: def string_assign():
   ...:     for _ in range(10000):
   ...:         b = x
   ...:

In [7]: def string_copy():
   ...:     for _ in range(10000):
   ...:         b = copy(x)
   ...:

In [8]: %timeit string_assign
4.04 ns ± 0.0335 ns per loop (mean ± std. dev. of 7 runs, 100,000,000 loops each)

In [9]: %timeit string_copy
4.03 ns ± 0.0177 ns per loop (mean ± std. dev. of 7 runs, 100,000,000 loops each)
```

euphonic/spectra.py
ajjackson: Perhaps `self._raw_spectrum_data_name` would parse better than `_spectrum_raw_data_name`. And it's more consistent with `_get_raw_spectrum_data()`.

oerc0122: That would certainly help. 

I don't know whether a blank line might also just visually separate that as being different to the ones below. I know they're doing something as a block, but I just get the subtle feeling that if I were modifying this code, I'd skip past that difference.

euphonic/spectra.py
ajjackson: The overhead on an array seems tolerable too. (It goes through a couple of lines of logic to find the `__copy__` method then delegates... https://github.com/python/cpython/blob/3.13/Lib/copy.py)

```python
In [12]: dat = np.random.random((1000, 1000))

In [13]: def numpy_copy():
    ...:     for _ in range(10000):
    ...:         b = dat.copy()
    ...:

In [14]: def copy_numpy():
    ...:     for _ in range(10000):
    ...:         b = copy(dat)
    ...:

In [15]: %timeit numpy_copy()
3.1 s ± 64.1 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [16]: %timeit copy_numpy()
3.15 s ± 131 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [17]:
```

euphonic/spectra.py
```diff
@@ -718,36 +718,30 @@ def _spectrum_data_name(cls) -> str:
         return f"{cls._spectrum_axis}_data"
 
     @classmethod
-    def _spectrum_raw_data_name(cls) -> str:
+    def _raw_spectrum_data_name(cls) -> str:
```
oerc0122, line 721: Looks good. At some point I can see these wanting to become `property` s, but not for this PR.

oerc0122 (approved):

euphonic/spectra.py
ajjackson: Unfortunately it is no longer allowed to make a `property` that is also a `classmethod`.

oerc0122: Given an instance has the properties of the class does that come into play? Nevermind then. Regardless, like I say, not for this PR. Let's get this merged when the tests are happy!

ajjackson: It seems useful/appropriate that this information can be accessed _before_ constructing the instance, as it is related to the `__init__` signature.

ajjackson: It doesn't actually look very hard to make a custom descriptor for similar behaviour. I _assume_ there is a reason it isn't actively encouraged 🤔  (or routinely imported from some library)

https://stackoverflow.com/questions/76378373/what-are-some-alternatives-to-using-classmethod-and-property-decorators-together

PR #329, opened 2024-11-19
Try dropping Conda from test runner

ajjackson:
It doesn't look like tox.ini/tox-conda is actually using conda to install dependencies (by calling conda_deps) so maybe we don't need this.

(Presumably it was added to manage installation of multiple Python versions, but the setup-python action can handle that now.)


ajjackson:
This has been incorporated into #324

PR #332, opened 2024-11-22
Migrate to tox 4

oerc0122:
- Migrate `tox.ini` to tox 4
- Remove custom install from tox.ini
- Add test for no C extension.
- Rename `phonopy_reader` to `phonopy-reader` for tox support (everywhere needed)
- Add verbose flag to tox's `pip install` command


ajjackson:
If switching the extra name from `phonopy_reader` to `phonopy-reader` works, please make sure to update the references in import error message and installation docs

ajjackson:
Shall we defer this to post-release, so that things are roughly in sync with #333 ?

oerc0122:
> Shall we defer this to post-release, so that things are roughly in sync with #333 ?

As we discussed in-person the other week, I think this is for the best.

ajjackson:
This is looking pretty good except that, bafflingly, the `platform` constraint is only respected for *testenv:py310-minrequirements-linux*. *py310-no-c* (platform = linux: linux) and *py310-minrequirements-mac* (platform = macos: darwin) are running on every platform.

PR #333, opened 2024-11-26
Use tox 4 for release testing

oerc0122:
Not backwards compatible with prior releases, requires #324 release.

Cannot test without #324 release.

### DIFF
--- a/build_utils/release_tox.ini
+++ b/build_utils/release_tox.ini
@@ -1,41 +1,41 @@
 [tox]
 # Use conda to set up the python environments to run in
-requires = tox-conda
+requires = tox>=4
 # The python environments to run the tests in
 envlist = pypi-py310-min,conda-py310-old-np,{pypi,conda}-{py310,py311,py312},pypisource-{py310,py312}
-# Skip the execution of setup.py as we do it with the correct version in commands_pre below
-skipsdist = True
+skip_install = True
 
 [testenv]
 changedir = tests_and_analysis/test
 test_command = python {toxinidir}/../tests_and_analysis/test/run_tests.py --report
 requirements_dir = {toxinidir}/../tests_and_analysis
+passenv =
+    EUPHONIC_VERSION
+    CC_LD
+extras =
+    test
+    matplotlib
+    phonopy-reader
+    brille
 
 # Test PyPI source distribution
 [testenv:pypisource-{py310,py312}]
 passenv = CC CC_LD LDFLAGS CPPFLAGS
 install_command = python -m pip install {opts} {packages}
-deps =
-    numpy
-    spglib  # Not interested here in whether source builds of spglib work
-    -r{[testenv]requirements_dir}/tox_requirements.txt
 commands_pre =
     python -m pip install \
     --force-reinstall \
-    'euphonic[matplotlib,phonopy_reader,brille]=={env:EUPHONIC_VERSION}' \
+    'euphonic=={env:EUPHONIC_VERSION}' \
     --no-binary 'euphonic'
 commands = {[testenv]test_command}
 
 
 [testenv:pypi-{py310,py311,py312}]
 install_command = python -m pip install {opts} {packages}
-deps =
-    numpy
-    -r{[testenv]requirements_dir}/tox_requirements.txt
 commands_pre =
     python -m pip install \
     --force-reinstall \
-    'euphonic[matplotlib,phonopy_reader,brille]=={env:EUPHONIC_VERSION}' \
+    'euphonic=={env:EUPHONIC_VERSION}' \
     --only-binary 'euphonic'
 commands = {[testenv]test_command}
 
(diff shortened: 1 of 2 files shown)


ajjackson:
I wonder if this ultimately needs to be the PR that breaks release testing into separate PyPI and Conda tests: tox and conda don't play together very nicely without the tox-conda package which is not compatible with tox 4. Specifically, our tox setup needs access to interpreters for multiple python versions, and conda only wants to give you one of those at a time. We may need to switch conda environments at the workflow level to test those builds.

build_utils/release_tox.ini
```diff
@@ -45,28 +45,12 @@ platform =
     linux: linux
     windows: win32
 deps =
-    numpy==1.24.0
+    numpy==1.21.3
```
ajjackson, line 48: 1.24.0 is the minimum according to pyproject.toml

build_utils/release_tox.ini
```diff
@@ -93,10 +80,12 @@ install_command = conda install {packages}
 conda_channels =
     conda-forge
     default
-conda_deps =
-    --file={[testenv]requirements_dir}/tox_requirements.txt
+deps = numpy==1.22
```
ajjackson, line 83: Can bump this to 1.24 to test v1.4.0

ajjackson:
This is looking good, only Conda tests are failing now.

build_utils/release_tox.ini
```diff
@@ -1,74 +1,49 @@
...
-skipsdist = True
+skip_install = True
 
 [testenv]
 changedir = tests_and_analysis/test
 test_command = python {toxinidir}/../tests_and_analysis/test/run_tests.py --report
 requirements_dir = {toxinidir}/../tests_and_analysis
+passenv =
+    EUPHONIC_VERSION
+    CC_LD
```
oerc0122, line 14: Duplicate `CC_LD`? Also, should we stick to one style? Newline or comma

build_utils/release_tox.ini
```diff
@@ -1,74 +1,49 @@
...
-deps =
-    numpy
-    spglib  # Not interested here in whether source builds of spglib work
-    -r{[testenv]requirements_dir}/tox_requirements.txt
+passenv = CC, CC_LD, LDFLAGS, CPPFLAGS
 commands_pre =
     python -m pip install \
     --force-reinstall \
-    'euphonic[matplotlib,phonopy_reader,brille]=={env:EUPHONIC_VERSION}' \
+    'euphonic{env:EXTRAS}=={env:EUPHONIC_VERSION}' \
```
oerc0122, line 24: Can we sub in-place using tox vars rather than environment here?

build_utils/release_tox.ini
```diff
@@ -1,74 +1,49 @@
...
+    numpy==1.21.3
 commands_pre =
     python -m pip install --force-reinstall \
-        -r{[testenv]requirements_dir}/minimum_euphonic_requirements_mac.txt
-    python -m pip install --force-reinstall \
-        -r{[testenv]requirements_dir}/tox_requirements.txt
+        -r{toxinidir}/tests_and_analysis/minimum_euphonic_requirements.txt
     python -m pip install \
-    'euphonic[matplotlib,phonopy_reader,brille]=={env:EUPHONIC_VERSION}' \
+    'euphonic{env:EXTRAS}=={env:EUPHONIC_VERSION}' \
```
oerc0122, line 44: I still think we should try the constraints file approach, may branch and play with that at some point.

oerc0122 (commented):
These are mostly due to me I suspect, but rather than interfere with your work on the branch

build_utils/release_tox.ini
ajjackson: Didn't seem to be allowed but I might have the syntax figured out now. One issue is that the name "extras" is reserved and we don't actually want a newline-separated list. So it has to be something else like "extras_str" 😞

build_utils/release_tox.ini
ajjackson: Agree, would prefer to separate from this PR as it applies to a wider range of test actions.

oerc0122:
Have you given this tox a test with a manual run or workflow trigger?

ajjackson:
Have just run it and looks good: https://github.com/pace-neutrons/Euphonic/actions/runs/12356249062

oerc0122:
> Have just run it and looks good: https://github.com/pace-neutrons/Euphonic/actions/runs/12356249062

```
  conda-py310: FAIL code 1 (0.65 seconds)
  conda-py311: FAIL code 1 (0.64 seconds)
  conda-py312: FAIL code 1 (0.65 seconds)
```

Task failed successfully?

ajjackson:
Separate action is better because these are different releases, typically built and tested on different days. It may even be best not to use tox for the conda one.

PR #336, opened 2024-12-06
Update release tests for new framework

ajjackson:
Current status:
- All conda tests failing (this is fine, the conda-forge package is not done yet)
- Ubuntu pypi-py310-min fails because it can't build spglib. The `ModuleNotFoundError: No module named 'distutils.msvccompiler'` error seems to be related to numpy/distutils versions.
- Windows pypisource-py310 and pypisource-py312 fail to build the Euphonic package, probably an environment/Meson issue
- macos-latest pypi-py310-min fails because it can't build spglib or h5py. I think there are no ARM wheels for the old versions and we haven't set up an appropriate environment to build them (e.g. with libhdf5 an old C compiler)
- macos-13 pypi-py310-min fails because it can't build spglib

### DIFF
--- a/build_utils/release_tox.ini
+++ b/build_utils/release_tox.ini
@@ -8,14 +8,15 @@ skipsdist = True
 
 [testenv]
 changedir = tests_and_analysis/test
-test_command = python run_tests.py --report
+test_command = python {toxinidir}/../tests_and_analysis/run_tests.py --report
+requirements_dir = {toxinidir}/../tests_and_analysis
 
 # Test PyPI source distribution
 [testenv:pypisource-{py310,py312}]
 install_command = python -m pip install {opts} {packages}
 deps =
     numpy
-    -r{toxinidir}/tests_and_analysis/tox_requirements.txt
+    -r{[testenv]requirements_dir}/tox_requirements.txt
 commands_pre =
     python -m pip install \
     --force-reinstall \
@@ -28,7 +29,7 @@ commands = {[testenv]test_command}
 install_command = python -m pip install {opts} {packages}
 deps =
     numpy
-    -r{toxinidir}/tests_and_analysis/tox_requirements.txt
+    -r{[testenv]requirements_dir}/tox_requirements.txt
 commands_pre =
     python -m pip install \
     --force-reinstall \
@@ -42,9 +43,9 @@ deps =
     numpy==1.21.3
 commands_pre =
     python -m pip install --force-reinstall \
-        -r{toxinidir}/tests_and_analysis/minimum_euphonic_requirements.txt
+        -r{[testenv]requirements_dir}/minimum_euphonic_requirements.txt
     python -m pip install --force-reinstall \
-        -r{toxinidir}/tests_and_analysis/tox_requirements.txt
+        -r{[testenv]requirements_dir}/tox_requirements.txt
     python -m pip install \
     'euphonic[matplotlib,phonopy_reader,brille]=={env:EUPHONIC_VERSION}' \
     --only-binary 'euphonic'
@@ -57,7 +58,7 @@ conda_channels =
     conda-forge
     default
 conda_deps =
-    --file={toxinidir}/tests_and_analysis/tox_requirements.txt
+    --file={[testenv]requirements_dir}/tox_requirements.txt
 commands_pre =
     conda install -c conda-forge euphonic={env:EUPHONIC_VERSION} matplotlib-base pyyaml h5py
 # Brille not available on conda
@@ -72,7 +73,7 @@ conda_channels =
     conda-forge
     default
 conda_deps =
-    --file={toxinidir}/tests_and_analysis/tox_requirements.txt
+    --file={[testenv]requirements_dir}/tox_requirements.txt
 commands_pre =
     conda install numpy=1.22
     conda install -c conda-forge euphonic={env:EUPHONIC_VERSION} matplotlib-base pyyaml h5py
--- a/build_utils/release_tox.ini
+++ b/build_utils/release_tox.ini
@@ -8,7 +8,7 @@ skipsdist = True
 
 [testenv]
 changedir = tests_and_analysis/test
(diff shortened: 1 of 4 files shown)


ajjackson:
Ok, the only thing failing now is the windows source build. It's something to do with build_utils/version.py https://github.com/pace-neutrons/Euphonic/actions/runs/12207029014/job/34057691946

@oerc0122 any thoughts on how to get more information about where this is going wrong?

The main difference between this environment and the cibuildwheel test is that this environment is likely conda-based (using tox-conda).

oerc0122:
Is it anything to do with:
```
ERROR: could not install deps [--file=D:\a\Euphonic\Euphonic\build_utils/../tests_and_analysis/tox_requirements.txt]; v = InvocationError("C:/Miniconda/Scripts/conda.exe install --quiet --yes -p conda-py310-old-np --channel conda-forge --channel default python=3.10 '--file=D:\\a\\Euphonic\\Euphonic\\build_utils/../tests_and_analysis/tox_requirements.txt'", 1)

[...]

Solving environment: ...working... failed

PackagesNotFoundError: The following packages are not available from current channels:

  - setuptools==71.1.0
```

The error isn't too helpful:
```
`D:\a\Euphonic\Euphonic\build_utils\.tox\pypisource-py312\python.EXE build_utils/version.py` failed with status 1.
```
But as it's an abort the only things that might fail are
```python
version_file.read_text
version_file.write_text
```
with `FileNotFound`
or
```python
        dist_version_file.parent.mkdir(parents=True, exist_ok=True)
```
with permissions issues?

ajjackson:
Maybe a local path issue where build_utils isn't in CWD?

ajjackson:
Hang on, should a source build even be running version.py? Where would it find build_utils?

ajjackson:
> Is it anything to do with:
> 
> ```
> ERROR: could not install deps [--file=D:\a\Euphonic\Euphonic\build_utils/../tests_and_analysis/tox_requirements.txt]; v = InvocationError("C:/Miniconda/Scripts/conda.exe install --quiet --yes -p conda-py310-old-np --channel conda-forge --channel default python=3.10 '--file=D:\\a\\Euphonic\\Euphonic\\build_utils/../tests_and_analysis/tox_requirements.txt'", 1)
> 
> [...]
> 
> Solving environment: ...working... failed
> 
> PackagesNotFoundError: The following packages are not available from current channels:
> 
>   - setuptools==71.1.0
> ```

I don't think this is the Windows source build issue; we also see it on the successful ubuntu run. This error is appearing on the Conda tests, where it tries to conda-install from a requirements.txt what has mostly been tested for pip. It looks like there isn't a conda-forge build for this version (but there are 71.0 and 72.0).

Another reason to separate pypi and conda tests, perhaps.

ajjackson:
> Hang on, should a source build even be running version.py? Where would it find build_utils?

Hmm, I don't like the answer to this question:

the new builds are including _everything_ in the source distribution, including tests, github workflows and test data. It unzips to 170M.

oerc0122:
> > Hang on, should a source build even be running version.py? Where would it find build_utils?
> 
> Hmm, I don't like the answer to this question:
> 
> the new builds are including _everything_ in the source distribution, including tests, github workflows and test data. It unzips to 170MB.

We can exclude those. I do believe that for a source dist the build-tools should be included.

ajjackson:
Including version.py makes sense, including release_tox and release.py is a bit odd. Maybe "local" build tools and CI build tools should be separated.

Anyway, I suggest that we _first_ try to fix the test to work on all platforms, and if that goes smoothly then follow up with a smaller build as a "bugfix" version.

ajjackson:
So:
- the Windows sdist failure is genuine
- some other stuff is fixed here
- Conda tests will likely need more work but I don't see the point in making a 1.4.0 Conda release

I think this is ready to go?

ajjackson:
v1.4.1.post1 release should have fixed sdist problems 🤞 

Tests running on that release here: https://github.com/pace-neutrons/Euphonic/actions/runs/12301522384

ajjackson:
And after rebase: https://github.com/pace-neutrons/Euphonic/actions/runs/12301569044

ajjackson:
PyPI source tests look good now, and unit tests are still passing.

Conda tests will probably fail but let's examine that once the new conda release is out.

Something funny is going on with duplicate unit test runs for push and pull_request, maybe I'll look at that separately.

build_utils/release_tox.ini
```diff
@@ -28,23 +31,42 @@ commands = {[testenv]test_command}
...
+platform =
+    linux: linux
+    windows: win32
 deps =
-    numpy==1.21.3
+    numpy==1.24.0
 commands_pre =
     python -m pip install --force-reinstall \
-        -r{toxinidir}/tests_and_analysis/minimum_euphonic_requirements.txt
+        -r{[testenv]requirements_dir}/minimum_euphonic_requirements.txt
```
oerc0122, line 51: The one question is whether to make this a constraints rather than a requirements file. Constraints may (should?) affect the resolver of the setup-requires in the pyproject.toml meaning you don't need to pre-install numpy. 

I think though that this becomes a later problem after e.g. #332

oerc0122 (approved):

build_utils/release_tox.ini
ajjackson: That does seem like a better match to what we are trying to do, yes. It still requires things to be update in two places, though. I wonder if the cleanest option is to replace the pip install with `uv install --resolution=lowest` and automatically detect the lowest supported version.

It might be worth bumping the minimum supported h5py to get rid of the platform-dependent complication...

ajjackson:
> Something funny is going on with duplicate unit test runs for push and pull_request, maybe I'll look at that separately.

Ah, I the the problem: tests are triggered on push to a branch with "release" in the name. Release branches shouldn't always be PRs, so this is handy. (And in this case the name was incidental.)

Ok, no need to take action at this moment. We *could* make the pattern a bit stricter, such as `release**` instead of `**release**`, so that a branch like this one doesn't trigger it.

PR #337, opened 2024-12-09
Fix and test sdist builds; reduce sdist size

ajjackson:
Currently the release-building action tests the binary wheels (as part of cibuildwheel workflow) but not the source distribution.

The source-build release test is failing on (only?) windows so perhaps something is not quite right. Here is a workflow to look into that.


ajjackson:
@oerc0122 Looks like the problem is a bug in version.py; it will error out if any of the proposed `git` names is unavailable. (i.e. on Mac/Linux if *git* is unavailable, on Windows if any of *git*, *git.cmd* or *git.exe* is unavailable.)

Currently the sdist is not tested in the package-building workflow. Properly testing this sort of thing is tricky:
- we can't really use a Github action that doesn't have git installed, and hiding it from the PATH would be quite fiddly. We get some coverage by testing on Windows where the multiple versions come into play, but the "no git" case seems impractical. Perhaps using a container?
- the tests _should_ not be in the sdist, so testing from a bare sdist install needs to _also_ have a copy of the repo available and make sure it isn't exposed during testing. The best way to do that is probably with *another* tox setup 😭 

The new action in this PR was mostly intended for troubleshooting and doesn't actually run tests. Maybe it should be migrated to a manual-dispatch workflow that targets a particular git commit and checks if the Ubuntu sdist -> Windows tests loop works, to be used when tweaking the build process.

Would it be very irresponsible to just roll out the version.py fix from this branch into a new release, without creating a bunch of new CI?

oerc0122:
> @oerc0122 Looks like the problem is a bug in version.py; it will error out if any of the proposed `git` names is unavailable. (i.e. on Mac/Linux if _git_ is unavailable, on Windows if any of _git_, _git.cmd_ or _git.exe_ is unavailable.)
> 
> Currently the sdist is not tested in the package-building workflow. Properly testing this sort of thing is tricky:
> 
> * we can't really use a Github action that doesn't have git installed, and hiding it from the PATH would be quite fiddly. We get some coverage by testing on Windows where the multiple versions come into play, but the "no git" case seems impractical. Perhaps using a container?
> * the tests _should_ not be in the sdist, so testing from a bare sdist install needs to _also_ have a copy of the repo available and make sure it isn't exposed during testing. The best way to do that is probably with _another_ tox setup 😭
> 
> The new action in this PR was mostly intended for troubleshooting and doesn't actually run tests. Maybe it should be migrated to a manual-dispatch workflow that targets a particular git commit and checks if the Ubuntu sdist -> Windows tests loop works, to be used when tweaking the build process.
> 
> Would it be very irresponsible to just roll out the version.py fix from this branch into a new release, without creating a bunch of new CI?

Well, now I feel like an idiot. I think I originally had a `FileNotFound` `except` and presumed `subprocess` would never `raise` it and removed it. 

I think that excluding the files from the sdist is easy enough and I think you're allowed to pass a directory to checkout, so we can probably do a bare checkout in the CI and call the `run_tests.py` to use the installed sdist version.

In terms of irresponsibility? It needs fixing. We just did a major build-system overhaul, we knew from the start things like this were liable to happen and would need attention. We did the best we could. I thought, though, that `cibuildwheel` would test the sdist for us, so I'm surprised we didn't catch this in that.

ajjackson:
The cibuildwheel docs suggest making the sdist in a separate step to the wheels, and having upload depend on both. That doesn't seem a bad idea, and would let some sdist tests run at the same time as other builds are being put together.

https://cibuildwheel.pypa.io/en/stable/deliver-to-pypi/#github-actions

ajjackson:
Here is the CI run that _nearly_ pushes to PyPI (so lets us see what files it would have used...) https://github.com/pace-neutrons/Euphonic/actions/runs/12253702919/job/34183074026?pr=337

I've now restored the usual triggers.

ajjackson:
Not sure how this was created. I don't _think_ I did...

Screenshot 2024-12-10 at 12 04 41

oerc0122:
> [@ajjackson](/ajjackson) [ajjackson](/ajjackson) [temporarily deployed](https://github.com/pace-neutrons/Euphonic/actions/runs/12253277209/job/34182097921) to pypi [December 10, 2024 09:43](#event-15597801622) — with  [GitHub Actions](https://github.com/apps/github-actions) Inactive

Mine now says: 
> ajjackson temporarily deployed to pypi 2 hours ago — with  GitHub Actions Inactive

ajjackson:
I don't see a weird release or files on PyPI, anyway 🤷‍♂️

.github/workflows/build_upload_pypi_wheels.yml
```diff
@@ -111,9 +137,67 @@ jobs:
...
+          ls -R dist/
+
+      - name: Update pip
+        shell: bash -l {0}
+        run: |
+          python -m pip install --upgrade pip
+
+      - name: Install from sdist
+        shell: bash -l {0}
+        run: python -m pip install $(find dist -name 'euphonic-*.tar.gz')[matplotlib,phonopy_reader,brille,test]
```
oerc0122, line 184: Unlikely to be multiple, but what happens if there are?

oerc0122 (commented):
Do you want to update the sdist excludes here?

.github/workflows/build_upload_pypi_wheels.yml
ajjackson: If they are different we should get a Pip resolution error

ajjackson: We could have something like

```bash
SDIST=$(find dist -name 'euphonic-*.tar.gz')
if test (echo $SDIST | wc -w) -ne 1
  then echo "Found $SDIST : not a single sdist file"; exit 1
fi
python -m pip install ${SDIST}[matplotlib,phonopy_reader,brille,test]
```

but I'm not sure it would be a lot safer.

oerc0122 (approved):

ajjackson:
It looks like #338 will make some further changes to the workflow, as cibuildwheel doesn't do well at testing its builds from sdist (as opposed to git checkout).

We could merge that into this branch first, or merge this branch first and retarget #338 . Any preference?

oerc0122:
> It looks like #338 will make some further changes to the workflow, as cibuildwheel doesn't do well at testing its builds from sdist (as opposed to git checkout).
> 
> We could merge that into this branch first, or merge this branch first and retarget #338 . Any preference?

It'll automagically retarget so it doesn't make too much difference, but I would probably say merge #338 as this is really one job to `master`.

ajjackson:
Ok, I think remaining tasks here are:

- [ ] strip out the docs as well
- [ ] update changelog again
- [ ] One last run of the workflow 🤞 
- [ ] Final sanity check from @oerc0122

ajjackson:
Pipeline looks good, if you're happy we can merge this and I'll have a crack at a release-number-updating manual action. (They need bumping anyway, may as well have a go at setting it up.)

oerc0122 (approved):
All looks sane to me, let's hope it works.

PR #338, opened 2024-12-10
Reduce source distribution size

ajjackson:
Depends on #337 

meson-python uses git-archive to create the sdist. Some of euphonic's test files are on the large side and we don't need to include them with every copy of the source.

This does have the downside that git-archive can no longer be used to produce a copy of the source with tests, but we weren't using that anyway.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,2 +1,3 @@
-# Added by versioneer
-euphonic/_version.py export-subst
+.github/** export-ignore
+tests_and_analysis/** export-ignore
+tox.ini export-ignore
--- a/.github/workflows/build_upload_pypi_wheels.yml
+++ b/.github/workflows/build_upload_pypi_wheels.yml
@@ -6,29 +6,7 @@ on:
   workflow_dispatch:
 
 jobs:
-  build-sdist:
-    name: Build sdist
-    runs-on: ubuntu-latest
-
-    steps:
-      - uses: actions/checkout@v4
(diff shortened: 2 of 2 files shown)


ajjackson:
Interesting, the tests _were_ running from the sdist!
https://github.com/pace-neutrons/Euphonic/actions/runs/12259358677

ajjackson:
Some very puzzling stuff going on here: the wheel tests are failing only on Linux, where it claims to be unable to find the run_tests.py. The file is clearly visible from a ls -R in the previous workflow step, and is accessed via an absolute path.

Screenshot 2024-12-10 at 16 28 00
Screenshot 2024-12-10 at 16 28 54

ajjackson:
@oerc0122 

Ahhh... I see the problem. The linux builds are running in a (manylinux) container, so they don't have access to the filesystem.

In that case, I think the options are

a) build Linux (and other platforms?) from the git checkout and not the sdist; or
b) create our own isolated wheel-testing workflow step that uses the wheel as input.

b) offers a tiny bit more purity for a lot of extra work. Let's do a)

ajjackson:
All green, finally 😮‍💨

https://github.com/pace-neutrons/Euphonic/actions/runs/12261330267

The sdist is still almost 10x the wheel size at 1MB. Probably due to docs/images; may as well strip those out as well?

oerc0122 (approved):

Issue #326, opened 2024-11-15
Create a release Workflow that integrates version-numbering and runs tests before pushing


ajjackson:
Discussion with @oerc0122 

The existing release process for Euphonic is at https://github.com/pace-neutrons/pace-developers/blob/master/euphonic/development/02_releases.md

- There are some manual steps that could be removed/consolidated
- The test_release action always fails at first while waiting for Conda. It can be a bit unclear what this is for and perhaps should be split into test_PyPI and test_conda workflows

As we move to a meson-based build system and rework the corresponding actions, it might be wise to have a single action which:

- Updates version numbers in euphonic.version, CITATION.cff
- Renames the "Unreleased" CHANGELOG section to reflect new version number and creates a new "Unreleased" section
- Commit these changes to master
- Builds wheels with new version number (which cannot be read from git tag at this point) and tests from wheels
- IF builds and tests are successful, create new tag, push to GitHub
- Create GitHub release from CHANGELOG (Github is quite forgiving about editing this later)
- Push wheels to PyPI

mducle:
I think the "Push wheels to PyPI" should be a separate manual action because you can only do it once (PyPI will not allow you to modify a release), so there's a high chance something will go wrong even if you make it through all the stages of the action successfully... I'd stop the automatic release action at the GitHub release stage, then manually test the wheels created in the GitHub release and if they were good then upload it to PyPI...

The system I used with `libpymcr` is to have an [action](https://github.com/pace-neutrons/libpymcr/blob/main/.github/workflows/release.yml#L18) run if there is `RELEASE` in the PR title and the PR is merged. This creates a GitHub release, then it runs the normal pipeline (which runs the test and creates the wheels) but the normal pipeline then [detects](https://github.com/pace-neutrons/libpymcr/blob/main/.github/workflows/main.yml#L184) it is a release and uploads the wheels to the release when it succeeds. The main probably with this workflow is that you have to merge the PR to trigger the release so if the release action has a bug / fails somehow you have to create a new PR. So, you could do something like this but manually triggered using a `workflow_dispatch` which takes the version number as an [input](https://github.blog/changelog/2021-11-10-github-actions-input-types-for-manual-workflows/).

ajjackson:
Our initial thought was also to be wary of the push to PyPI step, but...

Once we've run a test suite on the already-built wheels, what other reassurance are we waiting for? What is the likely failure scenario that doesn't kill the Action first?

ajjackson:
[...]

PR #340, opened 2024-12-11
Version number action

ajjackson:
Developing a github action that sets the version number in all the relevant places, for use in the release process.

This is part of #326

### DIFF
--- a/.github/workflows/build_upload_pypi_wheels.yml
+++ b/.github/workflows/build_upload_pypi_wheels.yml
@@ -33,8 +33,10 @@ jobs:
       fail-fast: false
     runs-on: ${{ matrix.os }}
     steps:
-      - name: Checkout project (for test files)
+      - name: Checkout project
         uses: actions/checkout@v4
+        with:
+          fetch-tags: true
 
       - name: Setup Python ${{ matrix.python-version }}
         uses: actions/setup-python@v5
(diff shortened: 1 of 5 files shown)


ajjackson:
This looks good now, we have successful runs when only the tag is updated
https://github.com/pace-neutrons/Euphonic/actions/runs/12284726381

and when both tag and files are updated
https://github.com/pace-neutrons/Euphonic/actions/runs/12284756168

And the tags are showing up in the list of repository tags. (I suppose they should be removed as these are not really official release candidates? But not much harm done if no release packages are made from them.)

ajjackson:
Ah, I see part of the problem: while rebasing I moved this history away from the tagged commits. That explains why it is seeing v1.4.0 tags 😅

ajjackson:
Ok, with a new tag the build names all look good again.

I'd like to add a step that validates the version number format and adds a "v" if missing.

ajjackson:
It seems to work! If I try to tag with "bad-version-tag" the run fails quickly:

Screenshot 2024-12-12 at 09 49 03

Whereas a slightly inconsistent format "1.4.1-rc1" gets reformatted to "v1.4.1rc1"
and this version is used for files and tag:

Screenshot 2024-12-12 at 09 59 06

ajjackson:
Unit tests are running here, not clear why they don't show up in the checklist.
https://github.com/pace-neutrons/Euphonic/actions/runs/12294025878

PR #348, opened 2024-12-18
Refactor test matrix: most logic at top level, call reusable workflow

### LINKED ISSUE #341: Drop macos-13 from run_tests.yml
We should keep running mac-os tests in the build/release workflows, but automated PR tests would benefit from dropping the slowest platform.

Perhaps some of the dependency combinations can also be moved to "workflow-dispatch-only" like Python 3.11?

ajjackson:
Based on https://stackoverflow.com/a/73953210

Closes #341

### DIFF
--- a/.github/workflows/run_tests.yml
+++ b/.github/workflows/run_tests.yml
@@ -20,17 +20,17 @@ jobs:
         config:
         - os: ubuntu-latest
           if: always()
-          python-versions: ${{((github.event_name == 'workflow_dispatch') && ['3.10', '3.11', '3.12'] || ['3.10', '3.12'])}}
+          python-versions: ${{((github.event_name == 'workflow_dispatch') && '[''3.10'', ''3.11'', ''3.12'']' || '[''3.10'', ''3.12'']')}}
           tox-skip-env: ${{((github.event_name == 'workflow_dispatch') && '' || '.*?(py311).*?')}}
           coverage: true
         - os: windows-latest
           if: always()
-          python-versions: ${{((github.event_name == 'workflow_dispatch') && ['3.10', '3.11', '3.12'] || ['3.10', '3.12'])}}
+          python-versions: ${{((github.event_name == 'workflow_dispatch') && '[''3.10'', ''3.11'', ''3.12'']' || '[''3.10'', ''3.12'']')}}
           tox-skip-env: ${{((github.event_name == 'workflow_dispatch') && '' || '.*?(py311).*?')}}
           coverage: false
         - os: macos-latest
           if: always()
-          python-versions: ${{((github.event_name == 'workflow_dispatch') && ['3.10', '3.11', '3.12'] || ['3.10', '3.12'])}}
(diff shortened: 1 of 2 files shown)


ajjackson:
I'm not sure why the codacy coverage is failing to upload; I haven't changed anything related to api keys or secrets. Maybe it doesn't like the upload coming from a sub-workflow?

ajjackson:
Aha, this is because secrets are not automatically passed to reusable workflows. https://docs.github.com/en/actions/sharing-automations/reusing-workflows#passing-inputs-and-secrets-to-a-reusable-workflow

Here I pass the codacy token explicitly, but another option is to use
`secrets: inherit` which will pass everything.

ajjackson:
Passing explicitly gave an error message that this one hadn't been defined in the workflow. Let's try "inherit", after all.

ajjackson:
All green, now to try manual dispatch

ajjackson:
I hit this "random" failure on Windows:
> PermissionError: [WinError 32] The process cannot access the file because it is being used by another process: 'D:\\a\\Euphonic\\Euphonic\\tests_and_analysis\\test\\reports\\coverage_1734543545.xml'

Seems like a rare collision in practice. But an unnecessary one because the Windows tests could be running without any coverage checks!

.github/workflows/run_tests.yml
```diff
@@ -17,86 +17,33 @@ jobs:
        (github.event_name == 'pull_request' && !contains(github.event.pull_request.labels.*.name, 'no_ci'))
     strategy:
       matrix:
-        os: [ubuntu-latest, windows-latest, macos-latest, macos-13]
+        config:
+        - os: ubuntu-latest
+          if: true
```
oerc0122, line 22: Dislike the tag `if`. Doesn't tell me what it's meant to do.

oerc0122 (approved):
Looks good, only issue is with the vague var `if`.

.github/workflows/run_tests.yml
ajjackson: How about "run"? I guess I could use "skip" and invert all the logic, but it makes it a bit harder to see "ubuntu-latest, true, true, true" as the "do everything" case.

ajjackson: `noskip: false` is a lot of negation to handle 😆

ajjackson: Ok I am going with "skip"

ajjackson:
Here's a random failure from coverage, having trouble with its sqlite database. I wonder if this is some kind of tox --parallel race condition that can be addressed as part of #352 

https://github.com/pace-neutrons/Euphonic/actions/runs/12410831358/job/34647267908

ajjackson:
Manual dispatch here is all green, I think this is good to merge?

https://github.com/pace-neutrons/Euphonic/actions/runs/12410831358

oerc0122 (approved):
All looks sensible. LGTM.

 To add salt though, I'd probably have gone with `to_run` and avoided the negation.

PR #349, opened 2024-12-18
Use constraints file in tox

### LINKED ISSUE #345: Experiment with constraints file or `uv`
In order to run the tests with the minimum requirements it may be possible to use a [constraints file](https://pip.pypa.io/en/stable/user_guide/#constraints-files) rather than a requirements file. 

The constraints file is a standard way of specifying the version constraints on a particular build which may allow the min-requirements tests to require no specialised commands beside specifying the constraints file in the `extras`.

The alternative would be an overall migration to the [`uv` build system](https://docs.astral.sh/uv/) or [`pip-tools`](https://pypi.org/project/pip-tools/2.0.1/) which provide the ability to build minimum requirements sets.

oerc0122:
Use constraints file in tox min-requirements. Also attempt to simplify testing commands.

Closes #345

### DIFF
--- a/build_utils/pypi_release_tox.ini
+++ b/build_utils/pypi_release_tox.ini
@@ -7,13 +7,14 @@ skip_install = True
 
 [testenv]
 changedir = tests_and_analysis/test
-test_command = python {toxinidir}/../tests_and_analysis/test/run_tests.py --report
-requirements_dir = {toxinidir}/../tests_and_analysis
 passenv =
     EUPHONIC_VERSION
 # The name "extras" is reserved and has its own format with newlines:
 # we need a token to inject into custom commands_pre, so call it extras_str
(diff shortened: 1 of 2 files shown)


ajjackson:
Looks like it's all passing again, now. _Hopefully_ nothing to worry about 😅

oerc0122:
> Looks like it's all passing again, now. _Hopefully_ nothing to worry about 😅

There's always more to worry about.

ajjackson (approved):
This all looks very clean and DRY, good stuff!

I've set a manual dispatch test run going here, in case there is any funny business on Python 3.11 or Intel Mac https://github.com/pace-neutrons/Euphonic/actions/runs/12439069788

Assuming all goes well with that I think this is ready.

PR #350, opened 2024-12-18
Separate release test workflows for PyPI and Conda-forge

### LINKED ISSUE #346: Split release tests into separate PyPI, Conda-forge

ajjackson:
Closes #346


ajjackson:
I suppose one benefit of having a `test_release` that calls both of these workflows would be that it already exists on `master` and gives a way to run these while working on them...

ajjackson:
Things seem to be running nicely now https://github.com/pace-neutrons/Euphonic/actions/runs/12401144969

Some concerning Windows failures but I don't think that's the workflow's fault?

ajjackson:
Hmm, the windows test failures are numerical differences of 0.005%. I'm not thrilled about them, but that's not a "cancel Christmas"-scale problem in my book. We can look at it next year and see if there is some sensitivity to Numpy versions/builds. It will be easier with this test split in place.

ajjackson:
Here's a random failure from coverage, having trouble with its sqlite database. I wonder if this is some kind of tox --parallel race condition that can be addressed as part of #352 

https://github.com/pace-neutrons/Euphonic/actions/runs/12410831358/job/34647267908

oerc0122 (approved):
Looks good. The #349 changes should slot right in, or be unnecessary. We'll find out once this is in.

.github/workflows/test_release_conda.yml
```diff
@@ -0,0 +1,68 @@
...
+      version:
+        description: 'The Euphonic release version to test e.g. 0.6.1'
+        required: true
+        type: string
+
+  workflow_dispatch:
+    inputs:
+      version:
+        description: 'The Euphonic release version to test e.g. 0.6.1'
+        required: true
```
oerc0122, line 14: Should this match?
```suggestion
        required: true
        type: string
```

.github/workflows/test_release_conda.yml
```diff
@@ -0,0 +1,68 @@
...
+
+jobs:
+  test:
+    strategy:
+      matrix:
+        os: [ubuntu-latest, windows-latest, macos-latest, macos-13]
+        python-version: ['3.10', '3.11', '3.12']
+      fail-fast: false
```
oerc0122, line 22: I know it's *exactly* the same thing, but I find it a bit more legible.
```suggestion
      fail-fast: false
      matrix:
        os: [ubuntu-latest, windows-latest, macos-latest, macos-13]
        python-version: ['3.10', '3.11', '3.12']
```

.github/workflows/test_release_pypi.yml
```diff
@@ -0,0 +1,94 @@
...
+      version:
+        description: 'The Euphonic release version to test e.g. 0.6.1'
+        required: true
+        type: string
+
+  workflow_dispatch:
+    inputs:
+      version:
+        description: 'The Euphonic release version to test e.g. 0.6.1'
+        required: true
```
oerc0122, line 14:
```suggestion
        required: true
        type: string
```

.github/workflows/test_release_pypi.yml
```diff
@@ -0,0 +1,94 @@
...
+
+jobs:
+  test:
+    strategy:
+      matrix:
+        os: [ubuntu-latest, windows-latest, macos-latest, macos-13]
+      fail-fast: false
```
oerc0122, line 21:
```suggestion
      fail-fast: false
      matrix:
        os: [ubuntu-latest, windows-latest, macos-latest, macos-13]
```

oerc0122 (approved):
Just some minor personal legibility things and symmetry. None of which are probably important.

.github/workflows/test_release_conda.yml
ajjackson: Weirdly, `type` is a required parameter for workflow call but not for workflow dispatch. I guess there is no harm in having it on both.

.github/workflows/test_release_conda.yml
ajjackson: I see what you mean!

PR #359, opened 2025-01-23
Tidy up euphonic.spectra: split into multiple files and lint

### LINKED ISSUE #358: Consider breaking up euphonic.spectra into submodules
The file is getting rather long and difficult to navigate; perhaps the classes should be moved to submodules and imported.

I imagine import loops will be an issue we leave the base Spectrum at top level and then try to import its child classes.

ajjackson:
Closes #358


ajjackson:
Hmm, at the moment git-blame from the github interface is succesfully tracking the history of lines in spectra.base, but not in spectra.collecions 😞

ajjackson:
See #360 for why a pylint warning is suppressed, related to a changing function signature.

euphonic/spectra/base.py
```diff
@@ -1,28 +1,21 @@
...
-from functools import partial, reduce
-from itertools import product, repeat, starmap
+from functools import partial
 
 import math
-import json
 from numbers import Integral, Real
-from operator import contains
-from typing import (Any, Callable, Dict, Generator, List, Literal, Optional,
+from typing import (Any, Callable, Dict, List, Literal, Optional,
```
oerc0122, line 10: Given Euphonic is now 3.10+, use modern typing. I.e. `list`, `dict`, `|`-unions, import `Callable` from `collections.abc`, etc.

euphonic/spectra/base.py
```diff
@@ -180,9 +174,9 @@ def _cut_x_ticks(x_tick_labels: XTickLabels | None,
         """Crop and shift x labels to new x range"""
         if x_tick_labels is None:
             return None
-        else:
-            return [(int(x - x0), label) for (x, label) in x_tick_labels
-                    if (x >= x0) and ((x1 is None) or (x < x1))]
+
+        return [(int(x - x0), label) for (x, label) in x_tick_labels
+                if (x >= x0) and ((x1 is None) or (x < x1))]
```
oerc0122, line 179: Comment describing logic of filtering? Also, is the `x` of `x_tick_labels` not `int`?

`# Check x in range(x0, x1)`

Could also say `x1 = np.inf if x1 is None else x1` and just have
`x0 <= x < x1`

euphonic/spectra/base.py
```diff
@@ -342,12 +335,11 @@ def _is_bin_edge(data_length: int, bin_length: int) -> bool:
...
-        elif bin_length == data_length:
+        if bin_length == data_length:
             return False
-        else:
-            raise ValueError((
-                f'Unexpected data axis length {data_length} '
-                f'for bin axis length {bin_length}'))
+        raise ValueError((
+            f'Unexpected data axis length {data_length} '
+            f'for bin axis length {bin_length}'))
```
oerc0122, line 342:
```suggestion
        raise ValueError(
            f'Unexpected data axis length {data_length} '
            f'for bin axis length {bin_length}')
```

euphonic/spectra/collections.py
oerc0122, line 1: Be careful, as this module shadows the `collections` stdlib module within this folder.

oerc0122: I am assuming this is mostly unmodified from the original. If anything's changed, can you highlight it for me?

meson.build
```diff
@@ -24,6 +24,7 @@ py_src = {
   'euphonic/data': ['__init__.py', 'bluebook.json', 'sears-1992.json',
                     'reciprocal_spectroscopy_definitions.txt'],
   'euphonic/readers': ['__init__.py', 'castep.py', 'phonopy.py'],
+  'euphonic/spectra': ['__init__.py', 'base.py', 'collections.py'],
```
oerc0122, line 27: Nice, glad this is moderately manageable even if you do have to manually track the modules.

oerc0122 (commented):
Only major thing is to watch out for shadowing. 

I suggest that updates porting the typing to 3.10 style is a separate PR or done through Boy Scout Rule.

euphonic/spectra/base.py
ajjackson: Yeah, I had this stuff in mind to do as well. Separate PR seems harmless, there's enough in this one for now.

euphonic/spectra/collections.py
ajjackson: The diff seems to be working correctly, which is nice! We want to _not_ squash this branch to make sure that keeps working. I'll do a bit of interactive rebasing (i.e. reordering and squashing) to keep the introduced history somewhat manageable.

oerc0122: For me it's saying that the entire file has changed (which is true) and I wasn't going to read through every line.

ajjackson: > For me it's saying that the entire file has changed (which is true) and I wasn't going to read through every line.

That's odd. I get the folded "large diff" buttons
Screenshot 2025-01-24 at 10 38 05

which unfold to a proper diff

Screenshot 2025-01-24 at 10 39 06

I appreciate there's a lot there though, I'm doing a scout of commit history to see if there are any significant changes worth bringing extra attention to.

ajjackson: Interestingly if in *spectra/\_\_init\_\_.py* I try

```python
from . import collections
from collections import Spectrum1DCollection
```

the first import works and the second one fails; somehow stdlib took priority!

euphonic/spectra/base.py
ajjackson: How about
```python
    @staticmethod
    def _cut_x_ticks(x_tick_labels: XTickLabels | None,
                     x0: int,
                     x1: int | None) -> XTickLabels | None:
        """Crop x labels to new x range, shifting indices accordingly"""
        if x_tick_labels is None:
            return None

        def in_range(x: int) -> bool:
            return x0 <= x < (float('inf') if x1 is None else x1)

        return [(int(x - x0), label)
                for (x, label) in x_tick_labels if in_range(x)]
```

ajjackson:
That's interesting, this windows test failure looks similar to the ones we see on Conda-forge https://github.com/pace-neutrons/Euphonic/actions/runs/12948116067/job/36116091358?pr=359#logs

euphonic/spectra/collections.py
oerc0122: > For me it's saying that the entire file has changed (which is true) and I wasn't going to read through every line.
>
> That's odd. I get the folded "large diff" buttons
> [screenshot]
> which unfold to a proper diff
>  [screenshot]
>I appreciate there's a lot there though, I'm doing a scout of commit history to see if there are any significant changes worth bringing extra attention to.

Note I'm referring to `collections.py` not `base.py`

oerc0122: > Interestingly if in _spectra/__init__.py_ I try
> 
> ```python
> from . import collections
> from collections import Spectrum1DCollection
> ```
> 
> the first import works and the second one fails; somehow stdlib took priority!

Yeah, the issue I had was from starting the Python interpreter/having `PYTHONPATH` in the folder.

ajjackson: > Note I'm referring to `collections.py` not `base.py`

Gotcha, that one has working blame but diff is less helpful

ajjackson: > Yeah, the issue I had was from starting the Python interpreter/having `PYTHONPATH` in the folder.

Ah yes, there could be a problem if running things from inside the *spectra* folder, I hadn't thought about that. I don't expect it will happen to me but it might happen to someone and be quite painful to figure out.

I suppose it could be named `spectrum_collection` then. Or just `collection` but that's asking for problems with typos...

oerc0122: I think that provided there's awareness, that's a good start. I'm happy for this to go in.

oerc0122 (approved):
Happy for this to go in once you're happy with the tests, given we're aware of the risk of shadowing, I don't think it'll be a major concern like you say.

euphonic/spectra/base.py
ajjackson: Ok will leave for now and have created an Issue to look at this separately

oerc0122: Sorry, I missed that, was in a meeting. 

It looks ok, but I think it would be fine to just have:
```python
    @staticmethod
    def _cut_x_ticks(x_tick_labels: XTickLabels | None,
                     x0: int,
                     x1: int | None) -> XTickLabels | None:
        """Crop x labels to new x range, shifting indices accordingly"""
        if x_tick_labels is None:
            return None

        if x1 is None:
            x1 = float('inf')

        return [(int(x - x0), label)
                for (x, label) in x_tick_labels if x0 <= x < x1]
```

ajjackson:
Many thanks for review, appreciate this sort of refactor is never much fun to look at!

PR #363, opened 2025-01-24
Implement alternate _bin_centres_to_bin_edges mode

### LINKED ISSUE #355: get_bin_edges limits to existing range, this causes trouble for broadening
Variable-width broadening does not suffer from #354 but can give incorrect results if the axis is provided in "bin centre" form (i.e. where the length of spectrum array matches length of axis array).

This is because
- the bin edges are calculated, but the outermost edges are not allowed to extended beyond the initial range. This leads to first and last bins that are narrower than the rest; presumably this was originally intended to give better visualisation results.
- The variable-width broadening assumes that all bin widths will equal the first one. This throws off the "bin width units" by a factor 2

I propose to fix this by
- adding an optional argument `spectrum.get_bin_edges(restrict_range=True)`, which is True by default
- set this to false when performing broadening
- change the default to False from Euphonic 2.0 and deprecate the option
- remove the option from Euphonic 3.0

This will be a _slightly_ breaking change in Euphonic 1.0 as the bins returned in broadening situations may change; it seems acceptable given that the incorrect results are fixed in the process.

I believe most applications of Euphonic broadening (i.e. in the CLI) are unaffected by this bug as they work with bin-edges directly. This remains the recommended workflow as it avoids ambiguity.

ajjackson:
Fixes #355

### DIFF
--- a/euphonic/spectra/base.py
+++ b/euphonic/spectra/base.py
@@ -336,8 +336,7 @@ class Spectrum(ABC):
                 [bin_centres[0]],
                 (bin_centres[1:] + bin_centres[:-1])/2,
                 [bin_centres[-1]],
-            )
-        )
+            ))
 
         half_diff = np.diff(bin_centres) * 0.5
         return np.concatenate(
@@ -358,7 +357,7 @@ class Spectrum(ABC):
             f'Unexpected data axis length {data_length} '
             f'for bin axis length {bin_length}')
 
-    def get_bin_edges(self, restrict_range=True) -> Quantity:
+    def get_bin_edges(self, restrict_range: bool = True) -> Quantity:
         """
         Get x-axis bin edges. If the size of x_data is one element larger
         than y_data, x_data is assumed to contain bin edges, but if x_data
@@ -979,8 +978,8 @@ class Spectrum2D(Spectrum):
         """
         assert axis in ('x', 'y')
 
-        bins = spectrum.get_bin_edges(bin_ax=axis)
-        bin_widths = np.diff(bins)
+        bins = spectrum.get_bin_edges(bin_ax=axis, restrict_range=False)
+        bin_widths = spectrum.get_bin_widths(bin_ax=axis, restrict_range=False)
 
         if not np.all(np.isclose(bin_widths, bin_widths[0])):
             bin_width = bin_widths.mean()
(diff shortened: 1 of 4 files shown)


ajjackson:
I'm not getting useful information from the codecov "patch" feedback, let's just ignore that failure...

ajjackson:
The last windows failure looks related to an incompatibility between Pint and numpy versions; Numpy 2.0 support was added in 0.24 but those tests installed Pint 0.23.

euphonic/spectra/base.py
```diff
@@ -675,14 +716,18 @@ def broaden(self: T, x_width,
...
+            self.assert_regular_bins(
+                message=(
+                    "Broadening with convolution requires a "
+                    "regular sampling grid."
+                ),
+                restrict_range=False,
+            )
             y_broadened = variable_width_broadening(
-                self.get_bin_edges(),
+                self.get_bin_edges(restrict_range=False),
```
oerc0122, line 727: Does the fact that this can change bins need to be documented in this docstring directly?

euphonic/spectra/base.py
```diff
@@ -976,29 +1021,38 @@ def copy(self: T) -> T:
...
-        is assumed to contain bin edges, but if bin_ax is the same size,
-        bin_ax is assumed to contain bin centres and a conversion is made.
-        In the conversion, the bin edges will not go outside the existing
-        data bounds so the first and last bins may be half-size. In addition,
-        each bin edge is assumed to be halfway between each bin centre,
-        which may not be an accurate assumption in the case of differently
-        sized bins.
+        Get bin edges for the axis specified by bin_ax. If the size of bin_ax
+        is one element larger than z_data along that axis, bin_ax is assumed to
+        contain bin edges. If bin_they are the same size, bin_ax is assumed to
```
oerc0122, line 1033: > if bin_they

? Looks like a train of thought got cut off.

ajjackson: Good spot, should be
```suggestion
        contain bin edges. If they are the same size, bin_ax is assumed to
```

euphonic/spectra/base.py
oerc0122: That makes sense!

ajjackson:
@oerc0122 did you want time to look at this some more or shall I auto-merge?

oerc0122 (approved):

PR #364, opened 2025-02-05
Get FC import from CASTEP 25.1 working

### LINKED ISSUE #294: Forthcoming  (release 25.1) changes to CASTEP checkpoint format require Euphonic update
I am planning a small change in the CASTEP binary checkpoint file format for the next release (25.1) in the handling of the force constant matrix, and I wanted to ensure that this will not generate problems with Euphonic's CASTEP binary reader.

The change will be do do with the dipole-dipole model Coulomb tail subtraction of the force constant matrix.  Currently the stored FCM may or may-not be Coulomb-subtracted depending on wither DFPT or FD/Supercell methods were used.  Euphonic has some logic to figure this out, but I haven't completely understood this.  I propose to add a new flag to the checkpoint format to explicitly indicate whether this is the case, so a minor change to Euphonic will be needed to read this (*).  This will also require a checkpoint version test (>=25.1?read new flag:retain old logic).

The reason for this is that I have implemented reading in of externally computed Born effective charges for calculating the dipole tail, LO/TO splitting correction etc, as requested, so the logic of Coulomb subtraction will be more complicated if used.

(*) As the flag is a Fortran logical some twiddling will be required to reliably convert to a 0/1 boolean - Fortran logicals are a compiler-dependent representation.

ajjackson:
Closes #294 

- CASTEP 25.1 adds extra fields to BORN and DIELECTRIC sections of castep_bin file; read (and discard) this information
- It is also now possible to _create_ a BORN section without corresponding DIELECTRIC, using a finite-displacement scheme with ultrasoft pseudopotentials. That throws off some Euphonic logic which assumes you would only ever have BORN and DIELECTRIC data at the same time. 
- As FC behaviour isn't really defined when you only have one, let's discard it in that case.

I need to test a wider range of import/present/not-present cases to be sure this will always do the right thing.

The good news is that most cases seem to work without this update; the main issue is with Berry-phase Born charge calculations.


ajjackson:
I deliberately pushed a failing commit with updated tests, to verify that the new tests cover a real problem. The actual fix is in the following commit.

euphonic/readers/castep.py
```diff
@@ -491,9 +491,20 @@ def read_interpolation_data(
             elif header == b'BORN_CHGS':
                 born = np.reshape(
                     _read_entry(f, float_type), (n_atoms, 3, 3))
+
+                if castep_version > Version("25.1"):
+                    # Extra field marks if born charges were read from BORN
+                    _ = _read_entry(f)
+
+
```
ajjackson, line 499:
```suggestion
```

ajjackson (commented):
Looks like I got slightly carried away with whitespace

euphonic/force_constants.py
```diff
@@ -108,16 +108,16 @@ def __init__(self, crystal: Crystal, force_constants: Quantity,
...
-        if born is not None:
+        if born is not None and dielectric is not None:
             self._born = born.to(ureg.e).magnitude
             self.born_unit = str(born.units)
             self._dielectric = dielectric.to(ureg(
                 'e**2/(bohr*hartree)')).magnitude
             self.dielectric_unit = str(dielectric.units)
         else:
-            self._born = born
+            self._born = None
```
oerc0122, line 118: Do you think this makes it API breaking? And should this only be done for >=25.1 files?

euphonic/readers/castep.py
```diff
@@ -528,6 +538,7 @@ def read_interpolation_data(
         data_dict['dielectric'] = dielectric*ureg(
             'e**2/(hartree*bohr)').to(dielectric_unit).magnitude
         data_dict['dielectric_unit'] = dielectric_unit
+
     except UnboundLocalError:
```
oerc0122, line 542:
```
from contextlib import suppress

...
with suppress(UnboundLocalError):
    ...
```
Is the more modern way of doing this.

oerc0122 (approved):
Looks sensible and simple. Concerns about changing the process generally when it should presumably only affect 25.1 files?

euphonic/force_constants.py
ajjackson: Interesting question.

- I don't know that it is possible to create a file with Born charges and do dielectric tensor before CASTEP 25.1. But I don't know that it was impossible either.
- Although it was possible to construct and serialise ForceConstants with only the Born charges, you couldn't go on to do Fourier interpolation without hitting errors.
- In a Euphonic ForceConstants, the presence of Born charges signals that the force constants matrix has had long-range terms subtracted. But without the dielectric tensor, the information is not sufficient to reproduce them. So the ForceConstants are in an ill-defined off-spec state and not really meaningful.

I think it does technically break API for people using Euphonic to construct and dump incomplete Force Constants information. But it seems unreasonable to support that use case?

euphonic/readers/castep.py
ajjackson: I think there are likely better ways of doing this bit without exceptions, anyway. But it seems out of scope for this PR. I could open an Issue to refactor?

oerc0122: Nah, just more of a highlight thing. It's not changed in this PR currently, and probably isn't worth doing so until next time you touch the file.

euphonic/force_constants.py
oerc0122: So given points 2 & 3, this is more of an earlier (safer) raise than a change of protocol as such. Happy with that.

ajjackson: I think so. It seems like a https://xkcd.com/1172/
situation

oerc0122: Yeah, the (dis)advantage of being more or less removed from the project is that I don't necessarily understand the context it lives in and will ask potentially weird questions based only on 10 lines of a 10,000 line project.

ajjackson: We are trying to be pretty responsible with breaking changes in Euphonic so it's always worth asking the question IMO. The package mostly exists to be used by downstream software, after all.

ajjackson:
Forgot the changelog 😅

oerc0122 (approved):
And here I thought `FC` stood for `Forget the Changelog`. 

Looks good!

PR #366, opened 2025-02-07
1-step release action

### LINKED ISSUE #326: Create a release Workflow that integrates version-numbering and runs tests before pushing

ajjackson:
Closes #326

### DIFF
--- a/.github/workflows/set_version.yml
+++ b/.github/workflows/set_version.yml
@@ -21,8 +21,9 @@ on:
     inputs:
       ref:
         description: "Target branch"
-        required: true
+        required: false
         type: string
+        default: ""
       version:
         description: "New version number"
         required: true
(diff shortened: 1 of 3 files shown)


ajjackson:
Move to #368 to avoid annoying duplicate test triggers due to "release" branch name.

PR #368, opened 2025-02-09
One-step release action

### LINKED ISSUE #326: Create a release Workflow that integrates version-numbering and runs tests before pushing

ajjackson:
Closes #326 
Replaces #364 
(Forgot that when branch name contains "release" tests get run twice. Should probably clean that up, but for now just use a different branch name...)


ajjackson:
https://github.com/pace-neutrons/Euphonic/actions/runs/13246775974
is the most realistic check I dare run for now: pypi push is commented out but everything else live. (Will need to pull the ensuing Github release...)

ajjackson:
All looks pretty convincing amd release page has appeared. GH is also showing some scary deployment rocket logos, but as far as I can tell nothing actually went to PyPi.

I guess the next step is to uncomment the PyPI lines, merge this, and try it for real 🫣

ajjackson:
@oerc0122 I think I've run enough sanity checks that it works, but maybe you could have a glance at the overall logic?

build_utils/bump_changelog.py
```diff
@@ -0,0 +1,115 @@
...
+    Each section should begin with two lines in format
+
+      `TAG <repository_address/compare/PREV_TAG...TAG>`_
+      --------------------------------------------------
+
+    except for 'Unreleased' which compares with HEAD
+    """
+
+    with open(changelog_file, "rt", encoding="utf8") as fd:
+        split_text = re.split(r"`(\S+) <\S+/compare/(\S+)\.\.\.\S+>`_\n-+", fd.read())
```
oerc0122, line 37: Huh, never realised that `re.split` split out the captured groups. That's good to know in future.

build_utils/bump_changelog.py
```diff
@@ -0,0 +1,115 @@
...
+
+    Each section should begin with two lines in format
+
+      `TAG <repository_address/compare/PREV_TAG...TAG>`_
+      --------------------------------------------------
+
+    except for 'Unreleased' which compares with HEAD
+    """
+
+    with open(changelog_file, "rt", encoding="utf8") as fd:
```
oerc0122, line 36:
```suggestion
    with changelog_file.open(encoding="utf8") as fd:
```

build_utils/bump_changelog.py
oerc0122: To avoid the next line, could also do:
```suggestion
        _, *split_text = re.split(r"`(\S+) <\S+/compare/(\S+)\.\.\.\S+>`_\n-+", fd.read())
```
However, this may be pretty arcane and less clear.

build_utils/bump_changelog.py
```diff
@@ -0,0 +1,115 @@
...
+
+    # Update "Unreleased" section
+    blocks[0].content = ""
+    blocks[0].previous_tag = tag
+
+
+def tag_to_header(tag: str, previous_tag: str) -> str:
+    """Produce header including github diff link"""
+    compare_tag = "HEAD" if tag == "Unreleased" else tag
+    header = f"`{tag} <{REPOSITORY_ADDRESS}/compare/{previous_tag}...{compare_tag}>`_"
```
oerc0122, line 76: Wonder if it's worth making this act like a template str.
```python
LINK_STR = "`{tag} <{REPOSITORY_ADDRESS}/compare/{previous_tag}...{compare_tag}>`_"
...
split_text = re.split(re.escape(LINK_STR).format(tag=r"(\S+)", 
                                                 previous_tag=r"(\S+)", 
                                                 compare_tag=r"\S+") +
                      r"\n-+")
...
header = LINK_STR.format(tag=tag, previous_tag=previous_tag, compare_tag=compare_tag)
```

.github/workflows/build_sdist.yml
```diff
@@ -0,0 +1,96 @@
...
+      - name: Upload source dist as build artifact
+        uses: actions/upload-artifact@v4
+        with:
+            name: python-source-distribution
+            path: dist/
+            if-no-files-found: error
+
+  test-sdist:
+    needs: build-sdist
+    name: Test build from sdist on Windows
```
oerc0122, line 43: Is there a test on mac/Linux for the sdist? Am I missing it?

oerc0122 (commented):
Looks good function-wise. Just some possible alternatives in the Python.

build_utils/bump_changelog.py
ajjackson: It's a nice trick but I think a bit misleading here. The first (empty) item _is_ probably something to do with the split text, whereas this implies it is some other return item.

build_utils/bump_changelog.py
```diff
@@ -0,0 +1,115 @@
...
+    """
+
+    with open(changelog_file, "rt", encoding="utf8") as fd:
+        split_text = re.split(r"`(\S+) <\S+/compare/(\S+)\.\.\.\S+>`_\n-+", fd.read())
+
+    # First item is always empty?
+    split_text = split_text[1:]
+
+    # From Python 3.12 can replace partition with itertools.batched
+    blocks = [Block(*block_data) for block_data in partition(3, split_text)]
```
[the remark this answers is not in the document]
ajjackson: Sure, why not. It made more sense when Block was a NamedTuple but I needed to switch to dataclass.

.github/workflows/build_sdist.yml
ajjackson: Nah, just testing it on one platform. Windows was the platform that had trouble in the past which is why it was used; I just ported it here from previous workflow.

Can probably switch back to ubuntu now if we like, and save a little CI time/budget.

build_utils/bump_changelog.py
ajjackson: What's the benefit here?

oerc0122: Stops `ruff` complaining.

build_utils/bump_changelog.py
oerc0122: Just realised the `re.escape` would escape the substitution brackets, which possibly makes this more complicated to handle than useful.

ajjackson: Ooh, I'll give it a try

build_utils/bump_changelog.py
oerc0122: https://docs.astral.sh/ruff/rules/builtin-open/
https://docs.astral.sh/ruff/rules/redundant-open-modes/

build_utils/bump_changelog.py
ajjackson: I just had to replace the `re.escape` with a string replacement `. -> \.`.

build_utils/bump_changelog.py
```diff
@@ -0,0 +1,122 @@
...
+"""
+from argparse import ArgumentParser
+from dataclasses import dataclass
+from pathlib import Path
+import re
+
+from toolz.itertoolz import partition
+
+REPOSITORY_ADDRESS = "https://github.com/pace-neutrons/Euphonic"
+LINK_STRING = "`{tag} <" + REPOSITORY_ADDRESS + "/compare/{previous_tag}...{compare_tag}>`_"
```
oerc0122, line 15: Again, not sure if cleaner or more obscure..?
```suggestion
LINK_STRING = f"`{{tag}} <{REPOSITORY_ADDRESS}/compare/{{previous_tag}}...{{compare_tag}}>`_"
```

build_utils/bump_changelog.py
ajjackson: I don't get it. PT123 is meant to be "use Pathlib!" but also enforces this entirely equivalent `open` syntax which "can improve readability"?

I get that `Path(path).open()` is more readable than `open(Path(path))`, but don't see that `path.open()` is more readable than `open(path)`.

oerc0122: So the one "advantage" I can see is that it enforces the use of `Path` consistently and will raise if it isn't. Whether this is "good" behaviour is certainly debatable.

ajjackson: Will commit it "for fun" then, in this case it doesn't matter much either way.

PR #369, opened 2025-02-12
Don't update CHANGELOG for pre-releases

ajjackson:
Changes should still appear as "Unreleased". This avoids the need to merge pre-release change notes into the final release.


ajjackson:
Dry run looks good, just need to revert the version bump and delete that tag...

https://github.com/pace-neutrons/Euphonic/actions/runs/13287622881

PR #379, opened 2025-03-20
Linting: whitespace, negation, type comparison, mutable defaults, staticmethods

ajjackson:
Pylint has been complaining for a while about some stuff that is easy to fix but would clutter other PRs. They should be done in manageable batches in separate PRs.

### DIFF
--- a/euphonic/io.py
+++ b/euphonic/io.py
@@ -26,7 +26,7 @@ def _to_json_dict(dictionary: Dict[str, Any]) -> Dict[str, Any]:
 
 
 def _from_json_dict(dictionary: Dict[str, Any],
-                    type_dict: Dict[str, Type[Any]]|None = None) -> Dict[str, Any]:
+                    type_dict: Dict[str, Type[Any]] | None = None) -> Dict[str, Any]:
     """
     For a dictionary read from a JSON file, convert all list key values
     to that specified in type_dict. If not specified in type_dict, just
--- a/euphonic/io.py
+++ b/euphonic/io.py
@@ -26,12 +26,14 @@ def _to_json_dict(dictionary: Dict[str, Any]) -> Dict[str, Any]:
 
 
 def _from_json_dict(dictionary: Dict[str, Any],
-                    type_dict: Dict[str, Type[Any]] = {}) -> Dict[str, Any]:
+                    type_dict: Dict[str, Type[Any]]|None = None) -> Dict[str, Any]:
     """
     For a dictionary read from a JSON file, convert all list key values
     to that specified in type_dict. If not specified in type_dict, just
     uses the default conversion provided by np.array(list)
     """
+    type_dict = {} if type_dict is None else type_dict
+
     for key, val in dictionary.items():
         if isinstance(val, list):
             if key in type_dict and type_dict[key] == tuple:
@@ -76,13 +78,17 @@ def _obj_to_dict(obj: T, attrs: Sequence[str]) -> Dict[str, Any]:
     return dout
 
 
-def _process_dict(dictionary: Dict[str, Any], quantities: Sequence[str] = [],
-                  optional: Sequence[str] = []) -> Dict[str, Any]:
+def _process_dict(dictionary: Dict[str, Any],
+                  quantities: Sequence[str]|None = None,
+                  optional: Sequence[str]|None = None) -> Dict[str, Any]:
     """
     Process an input dictionary for creating objects. Convert keys in
     'quantities' to Quantity objects, and if any 'optional' keys are
     missing, set them to None
     """
+    quantities = [] if quantities is None else quantities
+    optional = [] if optional is None else optional
+
     dictionary = copy.deepcopy(dictionary)
     for okey in optional:
         if not okey in dictionary.keys():
@@ -109,10 +115,12 @@ def _obj_to_json_file(obj: T, filename: str) -> None:
 
 
 def _obj_from_json_file(cls: Type[T], filename: str,
-                        type_dict: Dict[str, Type[Any]] = {}) -> T:
+                        type_dict: Dict[str, Type[Any]]|None = None) -> T:
     """
     Generic function for reading from a JSON file to a Euphonic object
     """
+    type_dict = {} if type_dict is None else type_dict
+
     with open(filename, 'r') as f:
         obj_dict = json.loads(f.read())
     obj_dict = _from_json_dict(obj_dict, type_dict)
--- a/euphonic/io.py
+++ b/euphonic/io.py
@@ -86,8 +86,6 @@ def _process_dict(dictionary: Dict[str, Any],
     'quantities' to Quantity objects, and if any 'optional' keys are
     missing, set them to None
     """
-    quantities = [] if quantities is None else quantities
-    optional = [] if optional is None else optional
 
     dictionary = copy.deepcopy(dictionary)
     for okey in optional:
--- a/euphonic/io.py
+++ b/euphonic/io.py
@@ -79,8 +79,8 @@ def _obj_to_dict(obj: T, attrs: Sequence[str]) -> Dict[str, Any]:
 
 
 def _process_dict(dictionary: Dict[str, Any],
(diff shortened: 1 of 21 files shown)


euphonic/io.py
```diff
@@ -26,12 +26,14 @@ def _to_json_dict(dictionary: Dict[str, Any]) -> Dict[str, Any]:
 
 
 def _from_json_dict(dictionary: Dict[str, Any],
-                    type_dict: Dict[str, Type[Any]] = {}) -> Dict[str, Any]:
+                    type_dict: Dict[str, Type[Any]]|None = None) -> Dict[str, Any]:
```
oerc0122, line 29: Most linters would suggest:
```suggestion
                    type_dict: Dict[str, Type[Any]] | None = None) -> Dict[str, Any]:
```
 Also, given you're using `|` unions, do the types need to be title-case?

euphonic/io.py
```diff
@@ -76,13 +78,17 @@ def _obj_to_dict(obj: T, attrs: Sequence[str]) -> Dict[str, Any]:
     return dout
 
 
-def _process_dict(dictionary: Dict[str, Any], quantities: Sequence[str] = [],
-                  optional: Sequence[str] = []) -> Dict[str, Any]:
+def _process_dict(dictionary: Dict[str, Any],
+                  quantities: Sequence[str]|None = None,
+                  optional: Sequence[str]|None = None) -> Dict[str, Any]:
```
oerc0122, line 83: Given they're only required to be sequence, why not:
```suggestion
                  quantities: Sequence[str] = (),
                  optional: Sequence[str] = ()) -> Dict[str, Any]:
```

euphonic/validate.py
```diff
@@ -103,7 +103,7 @@ def _replace_dim(expected_shape: Tuple[int, ...],
 
 
 def _ensure_contiguous_attrs(obj: object, required_attrs: List[str],
-                             opt_attrs: List[str] = []) -> None:
+                             opt_attrs: List[str]|None = None) -> None:
```
oerc0122, line 106: Does it need to be a list or is a `Sequence` ok?
```suggestion
                             opt_attrs: Sequence[str] = ()) -> None:
```

tests_and_analysis/test/euphonic_test/test_crystal.py
```diff
@@ -383,7 +383,7 @@ def test_cell_volume(self, crystal, expected_vol):
         (get_crystal('LZO'), lzo_spglib_cell)])
     def test_to_spglib_cell(self, crystal, expected_spglib_cell):
         spglib_cell = crystal.to_spglib_cell()
-        assert type(spglib_cell) is tuple
+        assert isinstance(spglib_cell, tuple)
```
oerc0122, line 386: Does it have to be a `tuple` or would any collection do? Or is it actually testing it's not `None`?

oerc0122 (commented):
Couple of stylistic things. Looks good. Are you looking to include formal linting in the CI at some point?

euphonic/io.py
ajjackson: Coming back to needless typing imports later 😄

euphonic/io.py
ajjackson: Hmm, that _should_ be ok. Let's try it.

euphonic/validate.py
ajjackson: I think any Collection will do, actually.

tests_and_analysis/test/euphonic_test/test_crystal.py
ajjackson: The spglib docs say tuple. Other things might work, but tuple is the official spec. https://spglib.readthedocs.io/en/stable/python-interface.html#crystal-structure-cell

euphonic/validate.py
```diff
@@ -122,6 +122,8 @@ def _ensure_contiguous_attrs(obj: object, required_attrs: List[str],
         will not throw an error. e.g. Depending on the material
         ForceConstants objects may or may not have 'born' defined
     """
+    opt_attrs = [] if opt_attrs is None else opt_attrs
```
oerc0122, line 125:
```suggestion
```

euphonic/io.py
```diff
@@ -76,13 +78,17 @@ def _obj_to_dict(obj: T, attrs: Sequence[str]) -> Dict[str, Any]:
...
+def _process_dict(dictionary: Dict[str, Any],
+                  quantities: Sequence[str] = (),
+                  optional: Sequence[str] = ()) -> Dict[str, Any]:
     """
     Process an input dictionary for creating objects. Convert keys in
     'quantities' to Quantity objects, and if any 'optional' keys are
     missing, set them to None
     """
+    quantities = [] if quantities is None else quantities
+    optional = [] if optional is None else optional
```
oerc0122, line 90:
```suggestion
```

tests_and_analysis/test/euphonic_test/test_crystal.py
oerc0122: It's more what's being tested here. Why are we doing this assert?

ajjackson:
CI linting is something to think about. I don't have a rigid view on how that should work, I guess the options are:

- Github action that gives a status based on lint but doesn't block anything. (Effectively this is what we have via Codacy.)
- Github action that blocks merge
- Github action that auto-fixes and commits. Can be a bit annoying if you are making multiple small commits.
- pre-commit hook that blocks and/or auto-fixes locally before you are allowed to commit. This can be a nice workflow but hard to enforce on new users; it works best combined with an equivalent CI job that blocks the tests from running.

Pre-commit hooks can be annoying if you want to locally commit work-in-progress to hop over to another branch, and there is a good reason the code is "invalid" at this moment. I guess there is a way to bypass it?

oerc0122:
> CI linting is something to think about. I don't have a rigid view on how that should work, I guess the options are:
> 
> * Github action that gives a status based on lint but doesn't block anything. (Effectively this is what we have via Codacy.)
> * Github action that blocks merge
> * Github action that auto-fixes and commits. Can be a bit annoying if you are making multiple small commits.
> * pre-commit hook that blocks and/or auto-fixes locally before you are allowed to commit. This can be a nice workflow but hard to enforce on new users; it works best combined with an equivalent CI job that blocks the tests from running.
> 
> Pre-commit hooks can be annoying if you want to locally commit work-in-progress to hop over to another branch, and there is a good reason the code is "invalid" at this moment. I guess there is a way to bypass it?

Honestly, blocking linting is probably the way to go, otherwise you end up with red on the tests with minimal gain. 

Start it with minimal invasiveness and as you feel more comfortable only enable the feature you think are good and stop there. Nothing says linting has to go whole hog. And no auto-formatting (personal taste)...

tests_and_analysis/test/euphonic_test/test_crystal.py
ajjackson: The job of this method is to produce correct input for spglib. This is a unit test which checks it produces what we understand to be correct input. Part of that spec is the type, part of that spec is the data.

oerc0122: Ah! I missed this was in testing. No worries, then.

ajjackson:
One option is to limit auto-formatting to isort and trailing whitespace

oerc0122:
> One option is to limit auto-formatting to isort and trailing whitespace

Eh, if they're raised in linting, I think that point is moot. Running `isort` or `ruff check --fix` is not a hard time avoiding `ruff format`

ajjackson:
Ok I have implemented a linter as gate to rest of CI: the requirements for this should be expanded slowly and carefully!

ajjackson:
> Looking good. Hopefully the linting doesn't bother thing too much. But if you only enable the things you agree with, it'll force everyone to obey your standards. No problem, right?

I just hope that is as easy as it sounds 😉 

Anyway there is nothing to stop people editing the list of checks as part of a PR, if some component turns out to be really oppressive. (Otherwise it can be disabled in-place with `# noqa: E666`)

PR #381, opened 2025-03-20
Lint imports

ajjackson:
Remove unused imports and sort the rest of them.

I don't really like how the automatic sorting separates `import mymodule` from `import thing from mymodule`. But if it's easy to keep consistent automatically, I can get used to it.


oerc0122:
> Remove unused imports and sort the rest of them using ruff.
> 
> Use `__all__` to make the public `euphonic.__init__.py` namespace more explicit. There is technically a tiny API break there by excluding some of imports which are used inside **init**... Maybe that merits a minor version bump?
> 
> I don't really like how the automatic sorting separates `import mymodule` from `import thing from mymodule`. But if it's easy to keep consistent automatically, I can get used to it.

C.f.
https://docs.astral.sh/ruff/settings/#lint_isort_no-sections

ajjackson:
> > I don't really like how the automatic sorting separates `import mymodule` from `import thing from mymodule`. But if it's easy to keep consistent automatically, I can get used to it.
> 
> C.f. https://docs.astral.sh/ruff/settings/#lint_isort_no-sections

Aha, I think it's this one: https://docs.astral.sh/ruff/settings/#lint_isort_force-sort-within-sections

euphonic/ureg.py
```diff
@@ -0,0 +1,17 @@
+from importlib.resources import files
```
oerc0122, line 1: Do probably want a docstring here, given we're introducing the module in this PR.

oerc0122: I've never used `importlib.resources`, what is the advantage over a `pathlib.Path`?

euphonic/__init__.py
```diff
@@ -23,16 +23,16 @@
...
-    DebyeWaller,
-    ForceConstants,
-    QpointFrequencies,
-    QpointPhononModes,
-    Quantity,
-    Spectrum1D,
-    Spectrum1DCollection,
-    Spectrum2D,
-    StructureFactor,
+    "__version__",
```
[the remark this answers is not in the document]
ajjackson: Good point, getting `__version__` from * would be bonkers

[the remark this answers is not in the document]
ajjackson: Ruff does pester for it, should suppress that somehow.

```
euphonic/__init__.py:9:22: F401 [*] `.version.__version__` imported but unused; consider removing, adding to `__all__`, or using a redundant alias
   |
 7 | from .structure_factor import StructureFactor
 8 | from .ureg import Quantity, ureg
 9 | from .version import __version__
   |                      ^^^^^^^^^^^ F401
10 |
11 | __all__ = [
   |
   = help: Add unused import `__version__` to __all__

Found 1 error.
[*] 1 fixable with the `--fix` option.
```

euphonic/ureg.py
ajjackson: It's based on the package structure rather than directly sniffing around the filesystem. I think that makes it a bit more robust for eggs, editable installs etc. Anyway it's the recommended way of handling data.

euphonic/__init__.py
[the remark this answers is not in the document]
oerc0122: The alternative is:
```python
from .version import __version__ as __version__
```
and ruff won't complain

euphonic/ureg.py
oerc0122: Does that mean it needs a `__init__.py` in the data folder?

euphonic/__init__.py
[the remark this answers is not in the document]
ajjackson: Done!

euphonic/ureg.py
ajjackson: Presumably: there is one. Haven't though about it for a while 😆

ajjackson: > Do probably want a docstring here, given we're introducing the module in this PR.

done!

oerc0122 (approved):

PR #382, opened 2025-04-07
Enable more linting rules and make notes on blocked rules

ajjackson:
Here is a manageable amount of progress on linting:

- enable lots of rules that were already passing
- make modest fixes for a few more
- make some notes in pyproject.toml on why certain rules remain commented-out or ignored


euphonic/brille.py
```diff
@@ -66,6 +66,10 @@ def __init__(self, crystal: Crystal,
         self._grid = grid
         self.crystal = crystal
 
+    def get_grid(self) -> br.BZTrellisQdc | br.BZMeshQdc |br.BZNestQdc:
```
oerc0122, line 69:
```suggestion
    def get_grid(self) -> br.BZTrellisQdc | br.BZMeshQdc | br.BZNestQdc:
```

euphonic/cli/intensity_map.py
oerc0122, line 1: Empty files? Or is that just a github thing?

euphonic/util.py
```diff
@@ -68,8 +68,7 @@ def is_gamma(qpt: np.ndarray) -> bool | np.ndarray:
         provided
     """
     tol = 1e-15
-    isgamma = np.sum(np.absolute(qpt - np.rint(qpt)), axis=-1) < tol
-    return isgamma
+    return np.sum(np.absolute(qpt - np.rint(qpt)), axis=-1) < tol
```
oerc0122, line 71: `np.allclose` ? 

Alternatively [`is_integer`](https://docs.python.org/3/library/stdtypes.html#float.is_integer):
```
return all(map(float.is_integer, qpt))

**Review by oerc0122 — COMMENTED** [review]
Looks good, just a couple of things. Also, not tempted to drive yourself mad with [`PTH`](https://docs.astral.sh/ruff/rules/#flake8-use-pathlib-pth) yet?

**Comment by ajjackson** [comment]
> not tempted to drive yourself mad with [`PTH`](https://docs.astral.sh/ruff/rules/#flake8-use-pathlib-pth) yet?

Didn't get that far down the lib yet 😉 

There is already an Issue related to this https://github.com/pace-neutrons/Euphonic/issues/316 . I want to keep things backward-compatible initially which might make it hard to comply with; will see when we get there!

File: euphonic/cli/intensity_map.py
Comment (ajjackson) [reply euphonic/cli/intensity_map.py:1]: This is the change in file mode to remove "executable" bits.

Comment (ajjackson) [reply euphonic/cli/intensity_map.py:1]: It's subtle in the Github UI! 
image

File: euphonic/util.py (hunk 2 above)
Comment (oerc0122) [reply euphonic/util.py:71]: Ah, are you looking for `np.isclose` then? with a reduce in axis 1?

Comment (ajjackson) [reply euphonic/util.py:71]: `np.all(np.isclose(qpt, np.rint(qpt)), axis=-1)` seems good, do you see any gotchas there?

Comment (oerc0122) [reply euphonic/util.py]: Tolerance (can be passed to `isclose`)? Could also chain ops and avoid an extra wrapping with:
```python
np.isclose(qpt, np.rint(qpt), atol=1e-15).all(axis=-1)
```
if you find that clearer.

I do find this form expresses what you're actually asking better at first glance!

oerc0122 (approved):
Happy with this. 

My suggestion on `isclose` is 100% optional.

euphonic/util.py
ajjackson: Yeah I had the `atol` in code but missed here for some reason. Nice dot-chaining, will put that in.

(I do find it hard to remember in numpy which methods can dot-chain and which work slightly differently in their function form.)

PR #383, opened 2025-04-07
(Mostly) fix flake8-bugbear and enable

ajjackson:
zip(*args, strict=True) is annoying because I like what it _does_ but not how it looks.

Maybe we need to increase the horizontal space a little 😭 

For B008 I choose to generally allow Quantity but not `ureg`, to discourage use of `ureg` where there could be Quantity. There is one place a couple of `ureg` _are_ appropriate (because we really are passing live units); it's a private method and I'm more comfortable to suppress there.


oerc0122:
> zip(*args, strict=True) is annoying because I like what it _does_ but not how it looks.
> 
> Maybe we need to increase the horizontal space a little 😭

Could be more obscure, but could add
```python
def zip_strict(*args):
    return zip(*args, strict=True)
```

euphonic/cli/powder_map.py
```diff
@@ -87,9 +87,9 @@ def get_parser() -> ArgumentParser:
 
 
 def _get_broaden_kwargs(q_broadening: Optional[Sequence[float]] = None,
-                        q_unit: Unit = ureg('1/angstrom').units,
+                        q_unit: Unit = ureg('1/angstrom').units,  # noqa: B008
```
oerc0122, line 90: I believe you might be able to do
```suggestion
                        q_unit: Unit = ureg('1/angstrom').units,  # noqa: function-call-in-default-argument
```
Which is a bit more self documenting (though in this case quite verbose)

euphonic/qpoint_phonon_modes.py
```diff
@@ -149,7 +150,7 @@ def reorder_frequencies(self,
                 dot_mat = np.reshape(dots, (n_branches, n_branches))
 
                 # Find greatest dot product
-                for j in range(n_branches):
+                for _ in repeat(None, n_branches):
                     max_i = (np.argmax(dot_mat))
```
oerc0122, line 154:
```suggestion
                    max_i = np.argmax(dot_mat)
```

euphonic/qpoint_phonon_modes.py
```diff
@@ -149,7 +150,7 @@ def reorder_frequencies(self,
                 dot_mat = np.reshape(dots, (n_branches, n_branches))
 
                 # Find greatest dot product
-                for j in range(n_branches):
+                for _ in repeat(None, n_branches):
```
oerc0122, line 153: Why do you consider `repeat(None, n_branches)` clearer than `range(n_branches)`?

ajjackson: It's a little bit more explicit (that the outer loop does nothing but repeat), and goes straight to an iterator rather than via the iterable lazy object thing.

But this is very marginal and debatable; I'm just playing around really 😅 . Can go back to `range` if you think it would confuse/slow future maintenance.

The important part here was making it clear that `j` is not used, and `_` achieves that.

ajjackson:
> > zip(*args, strict=True) is annoying because I like what it _does_ but not how it looks.
> > Maybe we need to increase the horizontal space a little 😭
> 
> Could be more obscure, but could add
> 
> ```python
> def zip_strict(*args):
>     return zip(*args, strict=True)
> ```

Adding a function to save a couple of characters makes thing worse, I think!

What I'd really like to do is to overload `zip` so that it defaults to strict=True, and make strict=False the special case. But then it becomes easy to forget the import and get subtly different behaviour in some modules... (I guess the linter would catch it, though?)

euphonic/cli/show_sampling.py
```diff
@@ -86,23 +92,37 @@ def main(params: Optional[List[str]] = None) -> None:
 
         ax.scatter(*zip(
             *euphonic.sampling.sphere_from_square_grid(n_theta * 2, n_theta,
-                                                       jitter=args.jitter)),
+                                                       jitter=args.jitter),
+            strict=True),
                    marker='x')
 
-    elif args.sampling == 'spherical-polar-improved':
+    elif args.sampling == "spherical-polar-improved":
```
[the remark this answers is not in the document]
ajjackson: It'll be a big diff that indents most of that file, so let's make a new issue

oerc0122:
> > > zip(*args, strict=True) is annoying because I like what it _does_ but not how it looks.
> > > Maybe we need to increase the horizontal space a little 😭
> > 
> > 
> > Could be more obscure, but could add
> > ```python
> > def zip_strict(*args):
> >     return zip(*args, strict=True)
> > ```
> 
> Adding a function to save a couple of characters makes thing worse, I think!
> 
> What I'd really like to do is to overload `zip` so that it defaults to strict=True, and make strict=False the special case. But then it becomes easy to forget the import and get subtly different behaviour in some modules... (I guess the linter would catch it, though?)

Hm... shadowing `zip` like that screams bad idea, particularly for people joining the project.

euphonic/cli/powder_map.py
ajjackson: Surely that goes over the line width 😞

ajjackson:
Yeah, it does. I don't think there's a realistic prospect that the default will change upstream before Python 4.

Maybe I should drop this particular lint then; the safety check is nice but the boilerplate is adding a lot of clutter and in many places the check isn't really needed.

oerc0122:
> Yeah, it does. I don't think there's a realistic prospect that the default will change upstream before Python 4.
> 
> Maybe I should drop this particular lint then; the safety check is nice but the boilerplate is adding a lot of clutter and in many places the check isn't really needed.

I still "like" it, and I do think it wouldn't be too bad to have:
```python
zips = partial(zip, strict=True)
```
in a utility module

euphonic/qpoint_phonon_modes.py
ajjackson: Fixed. In the process, realised one of these loops was broken anyway 😆

ajjackson:
Ok, `zips` is in! I didn't go too crazy with importing it where there is currently a single ok-looking `zip`, but it really helps clean up some sections.

euphonic/force_constants.py
```diff
@@ -738,14 +738,14 @@ def _calculate_phonons_at_qpts(
...
-                    f'Unexpected values for mode gradients at {n_idx}/{rmode_gradients.size} '
-                    f'indices {[x for x in zip(*idx)][:n_print]}..., '
-                    f'expected near-zero imaginary elements, got values of '
-                    f'{rmode_gradients.imag[idx][:n_print]}..., compared to a '
-                    f'max real value of {max_real}. Data may have been lost '
-                    f'when casting to real mode gradients',
-                    stacklevel=3)
+                warnings.warn(f"""
+Unexpected values for mode gradients at {n_idx}/{rmode_gradients.size} indices
+{[x for x in zip(*idx, strict=False)][:n_print]} ..., expected near-zero
```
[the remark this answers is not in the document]
ajjackson: Actually I think `{list(zip(*idx, strict=True))[:n_print]}` is fine, there should never be a length mismatch here.

As a bonus I saw a nice place for walrus above this!

oerc0122 (approved):
Looks good.

ajjackson:
Added a few tests to uncovered areas in the hope codacy will chill out.

There is currently some code to support deprecated CLI arguments, but it isn't used. I suppose best-practice is to delete the code; even if it will be useful later we can recover it from the git history.

ajjackson:
The removed code actually does a bit more than that: it ensures the warning will be seen by tweaking visibility level, and it redirects the parameter to another argument. (It's really meant for _renaming_ arguments rather than removing them.)

euphonic/cli/powder_map.py
ajjackson: Will leave it for now but if there was a better-flowing way to do this I'd like to.

oerc0122 (approved):

ajjackson:
Many thanks for looking at this @oerc0122

PR #389, opened 2025-04-14
388 Finished first pass through ruff rules list

ajjackson:
Some of the bigger / more-challenging / more-contentious ruff sections have been left until later, but we now have all the low-hanging fruit and some notes to guide future linting work.

### DIFF
--- a/euphonic/cli/utils.py
+++ b/euphonic/cli/utils.py
@@ -770,8 +770,8 @@ def _get_cli_parser(features: Collection[str] = {},
                 choices=('gauss', 'lorentz'),
                 help='The broadening shape')
         else:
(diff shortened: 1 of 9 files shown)


ajjackson:
PLW2901 is an interesting one: it identifies when a loop variable is reassigned inside the loop. This can be a dangerous event and worth flagging. But for many of the cases it appears in Euphonic the loop body is small and it seems an elegant way of applying some retyping or adjustment to a value before it is used.

I've tried to apply some common sense to where this is `# noqa`ed and where it is fixed. It seems a bad idea to skip this rule entirely when it can catch nasty things, so let's only suppress in safe-looking places.

PR #401, opened 2025-04-29
Linting: quote marks

ajjackson:
Use ruff to standardise on single-quote marks where possible.

The .git-blame-ignore-revs file marks revisions that can be usefully ignored from git-blame. To use this from CLI either set it as the default git-blame config using `git config blame.ignoreRevsFile .git-blame-ignore-revs` or explicitly call with `git blame --ignore-revs-file .git-blame-ignore-revs <args>`.

(For what it's worth, I personally prefer double-quotes! They are more consistent with the C/Julia/go notation that single-quote makes a character and double-quote makes a string. But single-quotes are more established in this codebase and the diff is much smaller this way around.)


ajjackson:
I don't entirely understand _why_ the quote manipulation in version.py broke `test-docs`, but reverting `version.py` seems to sort it out. I suspect it is related to source builds and the way the file is read into the build process?

oerc0122 (approved):
I trust this is the only thing which has been changed. If so, happy to just merge.

ajjackson:
The only "funny business" is the need to except `version.py`. It may be wise to do a pre-release so we can release-test (including lots of source installs etc.) before the main release.

PR #402, opened 2025-04-30
Linting: flake8-errmsg

ajjackson:
flake8-errmsg recommends to replace string literals in Exception calls with variable references. The reason is to give better-looking error messages that don't print the message twice: as output, _and_ in the stacktrace. https://pypi.org/project/flake8-errmsg/

I am taking the opportunity to also replace some awkward multiline string concatenations with triple-quoted multiline strings, to be dedented (and wrapped if possible) using `textwrap`. This should be a bit less error-prone and require less character-escapes (or switching between quote styles). A new function is provided (and tested) which will convert e.g.

```
"""\
    This text is indented
    and spans multiple lines.

    There are two paragraphs.
"""
```

to

```
"""This text is indented and spans multiple lines.

There are two paragraphs."""
```
(with line-wrapping to 70 char.)

This way we can have our large messages be attractive in both the source code and the user output!

Note that a lot of the triple-quotes open with a backslash to prevent the first line-break. While this line-break will be stripped out anyway, it would cause the indentation to be misinterpreted. An alternative (i.e. deranged) solution is to put space characters between the backticks and newline so that the indentation is correct.


ajjackson:
I don't generally worry too much about coverage complaints during linting. But as this lint makes changes to code execution (by assigning/using a variable) perhaps it would be worth reviewing these and making sure a few more error states are tested.

ajjackson:
Ok, this is as much extra test coverage as I can bear to sort out in this PR 😭 

Some of these error states are difficult to produce; I've dropped some but would prefer to leave the rest for now.

euphonic/broadening.py
```diff
@@ -78,12 +79,16 @@ def variable_width_broadening(
     if width_convention.lower() == 'fwhm' and shape == 'gauss':
         sigma_function = (lambda x: width_function(x) * FWHM_TO_SIGMA)
     elif width_convention.lower() == 'std' and shape == 'lorentz':
-        raise ValueError('Standard deviation unavailable for Lorentzian '
-                         'function: please use FWHM.')
+        msg = (
+            'Standard deviation unavailable for Lorentzian '
+            'function: please use FWHM.'
```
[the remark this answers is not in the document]
ajjackson: I agree, but I think that's a "minor version" kind of change rather than bugfix. I was hoping to include this linting in a bugfix release soon. Let's create an issue for error message formatting and milestone it to the next minor release.

euphonic/cli/intensity_map.py
```diff
@@ -34,15 +34,17 @@ def main(params: Optional[list[str]] = None) -> None:
...
+        msg = 'Eigenvectors are required to use "--weighting coherent" option'
+        raise TypeError(msg)
     if (args.weighting.lower() == 'coherent'
         and args.temperature is not None
         and not isinstance(data, ForceConstants)
     ):
-        raise TypeError('Force constants data is required to generate '
-                        'the Debye-Waller factor. Leave "--temperature" '
-                        'unset if plotting precalculated phonon modes.')
+        msg = dedent_and_fill("""\
```
oerc0122, line 43: By the way, is there a reason to not strip newlines in `dedent_and_fill` so you don't need the `"""\`?

euphonic/cli/show_sampling.py
```diff
@@ -95,20 +95,17 @@ def main(params: Optional[list[str]] = None) -> None:
...
     elif args.sampling == 'recurrence-cube':
         ax.scatter(
             *zips(
                 *euphonic.sampling.recurrence_sequence(args.npts, order=3)),
             marker='x',
         )
 
-    else:
-        raise ValueError('Sampling type f{args.sampling} is not implemented.')
+    else:  # random-sphere
```
oerc0122, line 105: Why remove the invalid check?

euphonic/cli/intensity_map.py
ajjackson: I guess the function could handle it, yes!

The main problem isn't that the newline exists, by the way, but that it breaks the indentation detection.

euphonic/cli/show_sampling.py
ajjackson: Because the argument goes through an argparse `choices` constraint this code is unreachable

oerc0122: I still personally prefer the explicit check rather than comment, but I'm happy.

euphonic/plot.py
```diff
@@ -171,11 +178,17 @@ def plot_1d(spectra: Spectrum1D | Spectrum1DCollection | Sequence[Spectrum1D] |
         # Check units are consistent
         for spectrum in spectra[1:]:
             if spectrum.x_data_unit != spectra[0].x_data_unit:
-                raise ValueError('Something went wrong: x data units are not '
-                                 'consistent between spectrum subplots.')
+                msg = (
+                    'Something went wrong: x data units are not '
```
oerc0122, line 182: Something went wrong is kind of implied by the `raise`, but don't want to change messages this PR

euphonic/readers/castep.py
```diff
@@ -72,10 +73,7 @@ def read_phonon_dos_data(
             weights = np.concatenate((weights, [frequency_block.weight]))
             freqs = np.concatenate((freqs, [frequency_block.frequencies]))
             mode_grads = np.concatenate((mode_grads, [qmode_grad.squeeze()]))
-        line = f.readline()
-        if 'BEGIN DOS' not in line:
-            raise RuntimeError(
-                f'Expected "BEGIN DOS" in {filename}, got {line}')
+        f.readline()  # Skip line
```
oerc0122, line 76: Slight change in functionality?

euphonic/spectra/base.py
```diff
@@ -1239,23 +1252,24 @@ def apply_kinematic_constraints(spectrum: Spectrum2D,
...
                                  .to('meV angstrom^2'))
 
     if (e_i is None) == (e_f is None):
-        raise ValueError('Exactly one of e_i and e_f should be set. '
+        msg = (
+            'Exactly one of e_i and e_f should be set. '
                          '(The other value will be derived from energy '
-                         'transfer).')
+                         'transfer).'
+        )
```
oerc0122, line 1271: Bad indent?

euphonic/spectra/base.py
```diff
@@ -1325,13 +1339,12 @@ def _distribution_1d(xbins: np.ndarray,
                      shape: Literal['lorentz'] = 'lorentz',
                      ) -> np.ndarray:
     x = _get_dist_bins(xbins)
-    if shape == 'lorentz':
-        dist = _lorentzian(x, xwidth)
-        dist = dist / np.sum(dist)  # Naively normalise
-    else:
-        raise ValueError("Expected shape: 'lorentz'")
+    if shape != 'lorentz':
+        msg = "Expected shape: 'lorentz'"
```
oerc0122, line 1343: Prefer adding `got` in future

euphonic/spectra/collections.py
```diff
@@ -234,13 +235,16 @@ def _validate_item(self, item: Integral | slice | Sequence[Integral] | np.ndarra
             return
         if isinstance(item, slice):
             if (item.stop is not None) and (item.stop >= len(self)):
```
oerc0122, line 237: There is a (little known) [`slice.indices`](https://docs.python.org/3/reference/datamodel.html#slice.indices) method for this.

euphonic/util.py
```diff
@@ -20,6 +21,30 @@
...
+    text
+        Multiline string with convenient linebreaks for source code display
+
+    Returns
+    -------
+    text
+        String for display to user via print() or Exception handling
+    """
+
+    text = textwrap.dedent(text)
```
oerc0122, line 42:
```suggestion
    text = textwrap.dedent(text.strip())
```
?

euphonic/util.py
```diff
@@ -228,33 +253,42 @@ def custom_decode(dct):
...
     else:
-        raise ValueError(
-            f'No data files known for collection "{collection}". '
-            f'Available collections: '
-            + ', '.join(list(_reference_data_files)))
+        msg = dedent_and_fill(f"""
+            No data files known for collection "{collection}".
+            Available collections:
+            """ + ', '.join(list(_reference_data_files)),
+        )
```
oerc0122, line 260:
```suggestion
        msg = dedent_and_fill(f"""
            No data files known for collection "{collection}".
            Available collections: {', '.join(_reference_data_files)}
            """
        )
```

euphonic/util.py
```diff
@@ -228,33 +253,42 @@ def custom_decode(dct):
...
     data = file_data['physical_property'].get(physical_property)
     if data is None:
-        raise ValueError(
-            f'No such collection "{collection}" with property '
-            f'"{physical_property}". Available properties for this collection'
-            ': ' + ', '.join(list(file_data['physical_property'].keys())))
+        msg =  dedent_and_fill(f"""\
+            No such collection "{collection}" with property
+            "{physical_property}". Available properties for this collection:
+            """ + ', '.join(list(file_data['physical_property'].keys())),
```
oerc0122, line 272:
```suggestion
            """ + ', '.join(file_data['physical_property']),
```
If this is a dict type

euphonic/util.py
```diff
@@ -723,7 +765,8 @@ def _get_supercell_relative_idx(cell_origins: np.ndarray,
             if np.all(dist_min <= 16*sys.float_info.epsilon):
                 break
         if np.any(dist_min > 16*sys.float_info.epsilon):
```
oerc0122, line 767: Why 16?

euphonic/validate.py
```diff
@@ -45,15 +45,20 @@ def _check_constructor_inputs(
         if not isinstance(typ, list):
             typ = [typ]  # noqa: PLW2901 redefined-loop-name
         if not any(isinstance(obj, t) for t in typ):
-            raise TypeError(f"The type of {name} {type(obj)} doesn't "
-                             f'match the expected type(s) {typ}')
+            msg = (
+                f"The type of {name} {type(obj)} doesn't "
```
oerc0122, line 49:
```suggestion
                f"The type of {name} {type(obj).__name__} doesn't "
```
In future?

pyproject.toml
```diff
@@ -89,7 +89,7 @@ select = [
        "COM",     # flake8-commas
        "C4",      # flake8-comprehensions: Good stuff, worth its own PR
        "T10",     # flake8-debugger
-       # "EM",      # flake8-errmsg: Big diff, worth its own PR
+       "EM",      # flake8-errmsg: Big diff, worth its own PR
```
oerc0122, line 92:
```suggestion
       "EM",      # flake8-errmsg
```

oerc0122 (commented):
All looks reasonable, just a couple of things to consider (mostly future PRs if worth thinking about, I suspect)

euphonic/cli/show_sampling.py
ajjackson: I do too, really. Maybe it can be tested by monkey-patching argparse... Or maybe we just accept that these things are untestable. Either way I don't like ending this with an `elif` and no `else`.

euphonic/plot.py
```diff
@@ -10,10 +11,14 @@
     import matplotlib.pyplot as plt
 
 except ModuleNotFoundError as err:
-    raise ModuleNotFoundError(
-        'Cannot import Matplotlib for plotting (maybe Matplotlib is '
-        "not installed?). To install Euphonic's optional Matplotlib "
-        'dependency, try:\n\npip install euphonic[matplotlib]\n') from err
+    err_msg = dedent("""\
```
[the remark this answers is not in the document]
ajjackson: Yeah, line wrapping doesn't respect the indented code block :-(

euphonic/plot.py
ajjackson: Hopefully will be spotted while reformatting error messages for #405

euphonic/readers/castep.py
ajjackson: Again, it's a specific warning for a failure state that seems very hard to reach. A file like this will still lead to an error, just a slightly less helpful one. But I don't know how you end up with such a file; non-DOS CASTEP outputs will blow up while reading numbers from the header section.

euphonic/spectra/base.py
ajjackson: Might as well dedent_and_fill this, I suspect the message will otherwise wrap poorly.

euphonic/spectra/collections.py
ajjackson: Created issue

euphonic/util.py
ajjackson: Don't know, don't dare change it without deeper investigation.

Magic numbers should be CONSTANTS but they also need good names.

euphonic/validate.py
ajjackson: Well, we are making small formatting changes to a few error messages here so it shouldn't hurt.

oerc0122 (approved):
Once this is passing test and lint, happy for it to go in!

PR #404, opened 2025-05-07
CI troubleshooting

ajjackson:
The min-requirements tests have started failing. There's no obvious connection to code changes so it seems likely to be dependency-related.

Just before the error, the log mentions an install of setuptools 80.3 which seems a bit odd.

But I can't reproduce the problem locally, regardless of setuptools version.


oerc0122 (approved):
Looks like it solves the problem and, hey, this is what documented min requirements are.

ajjackson:
Thanks for looking!

PR #408, opened 2025-05-16
Register pytest marks in pyproject.toml

ajjackson:
Currently the test suite produces a lot of warnings about the "marks" used to control which subset of tests runs.

I see these when running locally, but not in the CI logs. Not sure if they are suppressed somehow or it is a matter of different pytest versions, but it doesn't hurt to clear them up either way.

(Tox tends to hide the pytest warnings but you can see them in the GCC14 CI logs, which call the run_tests.py script directly. e.g. https://github.com/pace-neutrons/Euphonic/actions/runs/15066709878/job/42353084713)

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,6 +1,13 @@
 `Unreleased <https://github.com/pace-neutrons/Euphonic/compare/v1.4.4...HEAD>`_
 -------------------------------------------------------------------------------
 
+- Maintenance
+
+  - Pytest configuration has been moved from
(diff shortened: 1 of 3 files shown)


oerc0122 (approved):
Simple. Do these need to be removed from a `pytest.ini` somewhere or are they happy being duplicated.

ajjackson:
Ah that would explain it; on the CI the tests pick up this information from `tests_and_analysis/test/pytest.ini`. Let's find out what happens if we now delete it.

PR #415, opened 2025-07-17
Support Janus-style naming for phonopy files from CLI

### LINKED ISSUE #372: Support Janus-style phonopy outputs
https://github.com/stfc/janus-core produces Phonopy force constants with file prefixes e.g.

- seedname.phonopy.yml
- seedname.force_constants.hdf5

These are not recognised by Euphonic CLI tools which expect _force_constants.yaml_.

ajjackson:
Closes #372

Janus violates a couple of assumptions Euphonic makes about Phonopy
outputs:

- .yml file extension is used instead of .yaml;
- phonopy.yaml file also gains a prefix so has form "prefix-phonopy.yml".
- Corresponding force_constants.hdf5 also gains this prefix

Changes are only required to the filename-guessing logic in cli.utils;
the core phonopy import functions require explicit filenames.


ajjackson:
@mducle This isn't the direct Python access you were after, but still a workflow improvement for Janus users. Perhaps you can give it a test run?

ajjackson:
Tests are green. (Coverage is also fine, the red is for not _improving_ the coverage as much as it would like.)

In the end I got something workable by running the tests on my ARM Mac with `tox` (i.e. with pip installs in clean environments) rather than directly from my Conda env. I'm not thrilled that these little numerical discrepancies between installs are turning up occasionally. Sampling fewer q-points seems to help as the differences accumulate less?

mducle (approved):
Code looks good and CLI tools now work with Janus-Core generated phonopy-style files.

PR #417, opened 2025-08-06
Tweak calculation of atom positions during dipole-dipole subtraction

### LINKED ISSUE #377: NAC corrections applied twice if already present in phonopy.yaml
I found that the NAC are applied twice if they are already present in the `phonopy.yaml` file. In particular, the 4 lines of code which cause the issue are in the `from_phonopy` class method the `ForceConstants` class: 

```python
if fc.born is not None:
    fc = cls.from_total_fc_with_dipole(
                fc.crystal, fc.force_constants, fc.sc_matrix, fc.cell_origins,
                born=fc.born, dielectric=fc.dielectric)
```

The system is GaAs.
I can send the files to reproduce the error, but here is the wrong plot (where I also did not impose any ASR, but does not really matter):

Image

And here the corrected one (so, obtained by commenting the above lines):

Image

I am not sure how the developers want to fix this, so I will not open any PR for now

ajjackson:
Fixes #377 

Issue #377 reported some inconsistencies between Phonopy and Euphonic band structures when using long-range corrections. The system in that case has a rotation between unit cell and primitive cell, which leads to a non-symmetric derived supercell matrix. That makes it rather easy for a missing `transpose()` operation to cause subtle problems!

There seems to be an error in the calculation of long-range contributions to the force constants. When importing force constants from Phonopy these are subtracted to give short-range force constants in the form used by Euphonic.

### DIFF
--- a/euphonic/force_constants.py
+++ b/euphonic/force_constants.py
@@ -1786,7 +1786,9 @@ casting to real mode gradients.
         sc_atom_r = (np.repeat(cell_origins, n_atoms, axis=0)
                      + np.tile(crystal.atom_r, (n_cells, 1)))
         sc_to_u_matrix = np.linalg.inv(sc_matrix).transpose()
-        sc_atom_r_scell = np.einsum('ij,jk->ik', sc_atom_r, sc_to_u_matrix)
+        sc_atom_r_scell = np.einsum('ij,jk->ik',
+                                    sc_atom_r,
+                                    sc_to_u_matrix.transpose())
         sc_vecs = np.einsum('ji,ik->jk', sc_matrix, crystal._cell_vectors)
         sc_mass = np.tile(crystal._atom_mass, n_cells)
         sc_crystal = Crystal(sc_vecs*ureg('bohr'), sc_atom_r_scell,


ajjackson:
I have posted plots to #377 showing that this looks good for that system. It would be good to check a wider range of systems and cells if possible. We need to add something to the test suite which would catch a regression.

ajjackson:
Here is some data independently reproducing the bug. Force constants of MgO were calculated in a [[0, 2, 0], [0, 0, 2], [2, 0, 0]] supercell of the cubic unit cell, using the NEP89 potential and a BORN file calculated with CASTEP25/DFPT/PBEsol.

[phonopy.yaml.gz](https://github.com/user-attachments/files/21679737/phonopy.yaml.gz)

**Master branch**
master-bands

**With fix**
fix-bands

This could be used for a unit test that directly checks the corrected force constants match the fixed version, or indirectly checks that the resulting band structure contains no imaginary modes.

ajjackson:
A unit test is added using the MgO data. I have verified locally that this unit test fails if run without the previous commit.

ajjackson:
Ok, I have stared at the linear algebra enough to be convinced that there is no great significance in where the transpose() operations are placed; in fact they can cancel one another and may be removed.

I would like to give a more clear and consistent explanation of how each matrix is defined and what it means, but it's still a bit fiddly and I'm not confident I'd avoid making a mistake and worsening things.

The ideal reviewer for this would be @krefson if he's about 😄 Otherwise maybe @oerc0122 can take a quick look? I'd like to merge this in and release soon if possible. (aiidalab-qe-vibroscopy is currently shipping an incorrect workaround for this bug!)

ajjackson:
(I will replace some unrelated `einsum` with `@` in a separate cleanup PR once this is in.)

ajjackson:
I think we are confident enough at this point to move forward.

PR #418, opened 2025-08-08
Tidy up some numpy code

ajjackson:
~~Based on top of #417 , rebase on master when that is done.~~

- Don't use einsum where a matmul (@) will do the job. There can be a performance benefit to matmul as it will be dispatched to the appropriate BLAS routine more reliably. But more importantly it makes some operations easier to read and easier to learn the conventions at work. The more complex einsums are left to stand out as needing closer scrutiny.

- There are a few places `np.newxis` is aliased to `ax` to save space. That's quite confusing because `ax` usually means a Matplotlib axis. And it is unnecesssary as `np.newxis` is just an alias to `None` and this is compact enough already.

### DIFF
--- a/euphonic/force_constants.py
+++ b/euphonic/force_constants.py
@@ -556,8 +556,7 @@ class ForceConstants:
         # expensive phase calculations later
         sc_image_r = get_all_origins(
             np.repeat(n_sc_shells, 3) + 1, min_xyz=-np.repeat(n_sc_shells, 3))
-        sc_origins = np.einsum('ij,jk->ik', sc_image_r,
-                               self.sc_matrix).astype(np.int32)
+        sc_origins = (sc_image_r @ self.sc_matrix).astype(np.int32)
         unique_sc_origins = [[] for i in range(3)]
         unique_sc_i = np.zeros((len(sc_origins), 3), dtype=np.int32)
         unique_cell_origins = [[] for i in range(3)]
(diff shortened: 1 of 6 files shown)


euphonic/force_constants.py
```diff
@@ -556,8 +556,7 @@ def _calculate_phonons_at_qpts(
         # expensive phase calculations later
         sc_image_r = get_all_origins(
             np.repeat(n_sc_shells, 3) + 1, min_xyz=-np.repeat(n_sc_shells, 3))
-        sc_origins = np.einsum('ij,jk->ik', sc_image_r,
-                               self.sc_matrix).astype(np.int32)
+        sc_origins = (sc_image_r @ self.sc_matrix).astype(np.int32)
```
oerc0122, line 559: It should be noted (though this was in the original) that this will truncate rather than round.

euphonic/force_constants.py
```diff
@@ -556,8 +556,7 @@ def _calculate_phonons_at_qpts(
         # expensive phase calculations later
         sc_image_r = get_all_origins(
             np.repeat(n_sc_shells, 3) + 1, min_xyz=-np.repeat(n_sc_shells, 3))
-        sc_origins = np.einsum('ij,jk->ik', sc_image_r,
-                               self.sc_matrix).astype(np.int32)
+        sc_origins = (sc_image_r @ self.sc_matrix).astype(np.int32)
         unique_sc_origins = [[] for i in range(3)]
```
oerc0122, line 560: Any reason why this is the empty list when it's overwritten? Would

`[None]*3` not be clearer as to intent?

oerc0122 (approved):
All looks sensible.

euphonic/force_constants.py
ajjackson: I think None would be less clear because it makes it look more like a flat 3-vector! It is a `list[list[int]]` and we don't know in advance what the list lengths will be except that the first dimension is 3.

PR #419, opened 2025-08-11
Python support: Test py310 to 314, build 310 to 313

ajjackson:
- Python 3.14 is in pre-release
- Brille doesn't have wheels for 3.13 or 3.14 yet so don't run Brille tests on those

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,9 +4,14 @@
 - Requirements
 
   - Python 3.13 is supported
-  - (Testing only) migrated from *pytest-lazy-fixture* (which appears
-    to be abandoned and incompatible) to *pytest-lazy-fixtures*, an
-    active project with similar functionality.
+
+  - Test requirements
+
+    - Migrated from pytest 7 to pytest 8
+
+    - Migrated from *pytest-lazy-fixture* (which appears to be
+      unmaintained and incompatible) to *pytest-lazy-fixtures*, an
+      active project with similar functionality.
 
 - Maintenance
   - Started testing with Python 3.14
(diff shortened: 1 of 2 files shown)


ajjackson:
Building h5py on Ubuntu gives a test pass, which is nice. But it also adds a few minutes of runtime, both installing hdf5 on the runner and building the package.

It would be nice if we could separately generate a binary wheel (including a copy of hdf5 lib) and download this for test runs. Seems a bit complicated to set this up, though!

The strategy could be something like:

- Add a "build h5py if required" workflow
  - This attempts to download the h5py wheel from artifacts
  - If unsuccessful
    - install dependencies and build h5py wheel
    - Upload h5py wheel artifact
  - else
    - do nothing
- This workflow becomes an early step of run_tests. (Maybe in parallel with linting?)
  - Ubuntu test runs should download this artifact to a predictable location
  - The predictable location should be passed to pip "--extra-index-url" (and needs to follow a standard folder structure)

It seems like quite a bit of work and fiddling. Alternatively, we tolerate slow builds until there is an official h5py wheel for Python 3.14 and then drop the "install hdf5" step.

It depends if we want to establish a pattern of testing pre-releases or waiting until new versions have settled in a bit.

ajjackson:
Here is the manual run including pre-release tests. https://github.com/pace-neutrons/Euphonic/actions/runs/17290867774

This is basically working now but a bit suboptimal:

- The workflow files are getting a bit complicated
- The 3.14 test run also tests 3.12 because it happens to be installed and tox runs all it can see.

I'd like to address those two issues by adding more actions, but won't easily be able to test/develop them until something dispatchable is in master. This is the purpose of test_prerelease.yml.

PR #421, opened 2025-08-29
Separate workflows for prerelease-python and general tests

ajjackson:
The main test action is getting very cluttered with logic. As the actual test run is just a call to `tox`, we can simplify this a bit by having separate workflows for prerelease and supported Python versions.


ajjackson:
Manual dispatch here: https://github.com/pace-neutrons/Euphonic/actions/runs/17319825370

PR #422, opened 2025-09-04
Allow spectrum collection slices to exceed length

### LINKED ISSUE #407: More elegant use of slice argument
There is a (little known) [`slice.indices`](https://docs.python.org/3/reference/datamodel.html#slice.indices) method for this.

_Originally posted by @oerc0122 in https://github.com/pace-neutrons/Euphonic/pull/402#discussion_r2083537636_

ajjackson:
Closes #407

Following established conventions for python list and numpy array, if a slice exceeds the array length the result can quietly return up to the last available item. This allows things like [:10] to mean "up to first ten items" which is nice for printing etc.

As discussed in #407 this might be a dubious API break. But it's difficult to see the real-world harm.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,8 +1,20 @@
 `Unreleased <https://github.com/pace-neutrons/Euphonic/compare/v1.6.0...HEAD>`_
 -------------------------------------------------------------------------------
 
+- Features
+
+  - Spectrum1DCollection and Spectrum2DCollection can be indexed with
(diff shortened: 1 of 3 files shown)


ajjackson:
As this is quite low priority and technically breaking, let's defer it until Euphonic 2.0.

Let's also not wait _too_ long to make Euphonic 2.0. API version numbers don't have to be major "political" releases, that is a trap larger projects often fall into.

PR #426, opened 2025-10-07
Migrate CI macos-13 to macos-15-intel, drop brille tests on Mac

ajjackson:
Github is closing down macos-13 runners but providing macos-15-intel, let's use that instead.

---------------
*Update: we ran into some other problems while enabling this, particularly with Brille-Mac tests for brille>=0.8.1. The issue does not only manifest on the new Mac runners, but as we're overhauling the test suite it made sense to combine the work into one PR.*

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -13,8 +13,6 @@
       unmaintained and incompatible) to *pytest-lazy-fixtures*, an
       active project with similar functionality.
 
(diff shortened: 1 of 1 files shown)


ajjackson:
Well this is odd, the tests hang with macos-15. (Not even intel yet...) I tried both `tox run-parallel` and `tox run`.

ajjackson:
Yeah, it's hanging on `master` as well now 😞

PR #427, opened 2025-10-08
Pin macos-latest -> macos-14

ajjackson:
I'm having issues migrating to `macos-15`. Tests are also failing on `master` now and it's not clear if that is the same problem; see if we can get it green again by pinning back to `macos-14`.


ajjackson:
Particularly odd/frustrating as I test with tox on macos-14 locally with no issue

PR #435, opened 2025-11-25
Bump github actions

ajjackson:
As suggested by @oerc0122


ajjackson:
Testing wheel build action here: https://github.com/pace-neutrons/Euphonic/actions/runs/19684473456

PR #437, opened 2025-12-12
Overhaul handling of RNG: pass generators explicitly

### LINKED ISSUE #394: Pass seed parameter to functions that currently require "spooky" numpy seed manipulation.
In #393 , a `#noqa` is added to several unit tests which use `np.random.seed()`. This would be better avoided by passing seed values directly to functions.

ajjackson:
Closes #394

This approach requires less monkey-patching at test time and gives the user more opportunity to use fixed seeds or quasirandom generators

A monkey-patch is still used to handle the show_sampling CLI tool tests: this tool doesn't provide user input for seeds so a well-defined hook for patching seems a reasonably clean solution.

Surprisingly, it turns out that test_powder_map doesn't actually need a fixed random seed because none of the tests use jitter. That's possibly a bad thing but for now let's roll with it and use that to simplify further. If we introduce some RNG-sensitive tests we can use the same "clear monkeypatch hook" strategy as test_sampling.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -38,6 +38,18 @@
   - Improved error message when C extension import fails due to
     missing dynamic libraries.
 
+  - Random sampling functions (including deterministic sampling
+    functions with a random jitter option) accept a new ``rng``
+    parameter for a numpy.random Generator (generally preferred) or
+    RandomState (used for testing). This allows the user to manipulate
+    the random seed and bit generator to ensure reproducible or
+    independent calculations.
+
+  - The default argument to ``rng`` is a quasi-global Generator
+    instance in ``euphonic.util``. This avoids the computational
+    overhead of creating a new generator every time a random function
+    is called.
+
 - Maintenance
 
   - Bumped cibuildwheel to v3.3.0


euphonic/cli/powder_map.py
```diff
@@ -207,28 +207,29 @@ def main(params: Optional[list[str]] = None) -> None:
...
             spectrum_1d = _arrange_pdos_groups(spectrum_1d_col, args.pdos)
         elif args.weighting == 'coherent':
             spectrum_1d = sample_sphere_structure_factor(
                 fc, q,
                 dw=dw,
                 temperature=temperature,
-                sampling=args.sampling, jitter=args.jitter,
-                npts=npts,
-                energy_bins=energy_bins,
+                **general_kwargs,
```
oerc0122: Maybe time for a `match` in future? (Not this PR)

```python
        match args.weighting:
            case 'dos' if args.pdos is None:
                spectrum_1d = sample_sphere_dos(
                    fc, q, **general_kwargs, **calc_modes_kwargs
                )
            case (*_, 'dos', *_):
                spectrum_1d_col = sample_sphere_pdos(
                    fc, q,
                    weighting=_get_pdos_weighting(args.weighting),
                    **general_kwargs,
                    **calc_modes_kwargs)
                spectrum_1d = _arrange_pdos_groups(spectrum_1d_col, args.pdos)
            case 'coherent':
                spectrum_1d = sample_sphere_structure_factor(
                    fc, q,
                    dw=dw,
                    temperature=temperature,
                    **general_kwargs,
```
Not sure if this is what you mean by `"dos" in args.weighting`?

euphonic/cli/show_sampling.py
```diff
@@ -41,7 +46,8 @@ def main(params: Optional[list[str]] = None) -> None:
 
     if args.sampling == 'golden-square':
```
oerc0122: This is definitely a space for a `match` at some point! (Not this PR, since it's not being changed.)

tests_and_analysis/test/euphonic_test/test_sampling.py
```diff
@@ -1,16 +1,19 @@
...
 import pytest
 
 import euphonic.sampling
 
+# @pytest.fixture
+# def fix_np_rng(monkeypatch):
+#         monkeypatch.setattr(np.random, 'default_rng',
+#                             partial(np.random.RandomState, seed=0))
+
+def rng_seed_0():
```
oerc0122: Does this return a new generator or does it just affect the global state? I think the former from what I read, but best to just make sure.

Also, this approach seems to be considered [legacy](https://numpy.org/doc/stable/reference/random/legacy.html#numpy.random.RandomState)

oerc0122 (approved):
Just some things to consider, if you're happy with them, this is good to go.

euphonic/powder.py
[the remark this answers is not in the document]
ajjackson: Agreed. It was a bit annoying putting `rng` at the end of these parameter lists to avoid API break when it would naturally sit a bit earlier (next to `jitter` where that is available!)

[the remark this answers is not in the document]
ajjackson: Created new Issue #440

PR #443, opened 2025-12-17
Fix intermittent Linux test failures: rework parallelism over unit tests

ajjackson:
We have been getting some segfaults on Linux Euphonic-Brille tests. That's odd because there haven't been any relevant changes to Euphonic or Brille releases since the tests all passed as part of 1.5.0 release.

### Update:
The problem seems to be using `tox run-parallel` with Brille wheels; each python version gets its own wheel with its own copy of OpenMP. Even though they are the same version we activate different instances and get familiar-looking segfaults.

The solution here is a bit over-engineered especially given that Brille issue should go away by itself soon (as they are moving away from OpenMP). But it has other benefits, too. Moving from `tox run-parallel` to `pytest-xdist` means a single Python version is tested at a time but the tests are run in parallel. This is a more useful kind of parallelism for local development; typically on my laptop I only test one or two python versions and the remaining 10 cores are sitting unused!

`coverage` as currently used in Euphonic doesn't quite work right with `pytest-xdist`, this is addressed in the `pytest-cov` package but that doesn't seem to support the way Euphonic was calling coverage. In the end passing command-line args seems simpler anyway.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,6 +1,21 @@
 `Unreleased <https://github.com/pace-neutrons/Euphonic/compare/v1.5.0...HEAD>`_
 -------------------------------------------------------------------------------
 
+- Requirements
+
+  - Test requirements
+
+    - Added *pytest-xdist* and *pytest-cov* requirements for parallel
+      testing at pytest level.
+
+- Maintenance
+
+  - Reworked CI strategy from parallelism over tox environments to
+    parallelism over unit tests. This addresses OpenMP conflicts
+    between simultaneous test environments, and gives a large speedup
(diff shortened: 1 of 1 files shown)


ajjackson:
Hmm, with separate `-m brille` and `-m not brille` things all pass on CI.

ajjackson:
Trying parallelism at pytest level instead of tox, so we only have one version loaded at a time.

Linux py3.14 gave an error related to lack of coverage information. Could be spurious, could be a compatibility issue with pytest-xdist?

ajjackson:
Still some issues with pytest-cov, trying .coveragerc tweaks suggested here: https://github.com/pytest-dev/pytest-cov/issues/129

PR #450, opened 2026-01-07
Replace Optional with 3.10+ alternation

oerc0122:
Since 3.10, the preferred way to do optional with with `| None` alternation. 

Replace all instances of `Optional` and remove from imports.


oerc0122:
Codacy failures seem to just be existing issues re-highlighted due to changes.

PR #452, opened 2026-01-22
Refactor copy semantics to enable `copy.copy` behaviour

oerc0122:
Reword `Spectrum*` to Support Python stdlib [`copy`](https://docs.python.org/3/library/copy.html) semantics while remaining non-api breaking.


ajjackson:
I see ubuntu test is passing, red x is from codacy upload which fails from fork.

PR #457, opened 2026-01-27
Bump seekpath

### LINKED ISSUE #319: Check if Seekpath reference cells are consistent
We could get incorrect q-point paths if seekpath defines a path relative to some cell that is rotated compared to the input cell. If not practical to rotate the data, some warning/error should be raised.

ajjackson:
Closes #319 

Seekpath 2.1 adds functions can use to fix #319 

Version 2.2.0 improved compatibility with recent spglib versions. As seekpath doesn't have a lot of other dependencies, it doesn't seem unreasonable to demand this and make good use of the spglib compatibility to clean things up.

However, there was to be a bug in the SymmetryDetectionError logic when used with older versions of spglib. A workaround was been offered upstream and is now merged: https://github.com/giovannipizzi/seekpath/pull/118 . So we can depend on 2.2.1+ to get this fix.

As this PR fixes incorrect behaviour, it rides a slightly fuzzy line between "breaking" and bugfix. Some users might be surprised that their band structure is now reordered or has different labels: this should only happen if there was a problem with the previous behaviour, but it is possible some users will not recognise/agree that it was a problem.

As there are dependency updates, this cannot be a "bugfix release" anyway; I propose that this goes into the next minor version release.


ajjackson:
Compared with the example in #319 , switching to the new function seems to fix the problem of inconsistent band structures as the cell is mutated

ge-bandsge_rot-bands

But it also adds a test failure. That _likely_ means the original test involved a non-spglib-friendly cell orientation and the new results are correct. But it needs a closer look.

euphonic/cli/utils.py
```diff
@@ -331,24 +331,44 @@ def _insert_gamma(bandpath: dict) -> None:
 SplitArgs = dict[str, Any]
 
 
+def _convert_labels_to_fractions(bandpath: dict, limit: int = 32) -> None:
```
oerc0122:
```suggestion
def _convert_labels_to_fractions(bandpath: dict, *, limit: int = 32) -> None:
```
Do you know the types in the dict? You know the minimum requirement is:

```python
class BandpathDict(TypedDict, total=False):
    explicit_kpoints_labels: Sequence[str]
    explicit_kpoints_rel: Iterable[float]
```

ajjackson: Done!

oerc0122 (approved):

PR #459, opened 2026-02-13
Use keyword-only arguments in euphonic.powder

### LINKED ISSUE #440: Use mandatory kwargs in euphonic.powder functions
API breaking (not this PR), but I think at some point you want a `*` kwargs only here and elsewhere in this module.

_Originally posted by @oerc0122 in https://github.com/pace-neutrons/Euphonic/pull/437#discussion_r2615449490_

ajjackson:
Closes #440 

This makes the API more resilient against changes (e.g. because a new algorithm takes extra arguments). It is a breaking change so should be part of Euphonic 2.0 release.

I also take the opportunity to move the `rng` argument next to `jitter` as they are related.

Note that this PR targets `release-2.0.0`, not `master`. As there are a few things to do for 2.0 we may want to make another bugfix release from `master` before v2 is ready.

### DIFF
--- a/euphonic/powder.py
+++ b/euphonic/powder.py
@@ -25,11 +25,12 @@ SphericalSamplingOptions = Literal['golden',
 
 def sample_sphere_dos(fc: ForceConstants,
                       mod_q: Quantity,
+                      *,
                       sampling: SphericalSamplingOptions = 'golden',
                       npts: int = 1000,
                       jitter: bool = False,
-                      energy_bins: Quantity = None,
                       rng: RNG = rng,
+                      energy_bins: Quantity = None,
                       **calc_modes_args,
                       ) -> Spectrum1D:
     """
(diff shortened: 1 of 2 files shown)


euphonic/powder.py
```diff
@@ -117,13 +118,14 @@ def sample_sphere_dos(fc: ForceConstants,
 def sample_sphere_pdos(
         fc: ForceConstants,
         mod_q: Quantity,
+        *,
         sampling: SphericalSamplingOptions = 'golden',
         npts: int = 1000,
         jitter: bool = False,
+        rng: RNG = rng,
         energy_bins: Quantity = None,
```
oerc0122:
```suggestion
        energy_bins: Quantity | None = None,
```

euphonic/powder.py
oerc0122:
```suggestion
phonons = fc.calculate_qpoint_phonon_modes(qpts_frac, **calc_modes_args)
```

oerc0122 (approved):
Looks good. Just a bit of typing cleanup since you're here

euphonic/powder.py
```diff
@@ -25,11 +25,12 @@
 
 def sample_sphere_dos(fc: ForceConstants,
                       mod_q: Quantity,
+                      *,
                       sampling: SphericalSamplingOptions = 'golden',
                       npts: int = 1000,
                       jitter: bool = False,
-                      energy_bins: Quantity = None,
                       rng: RNG = rng,
+                      energy_bins: Quantity = None,
```
ajjackson:
```suggestion
                      energy_bins: Quantity | None = None,
```

euphonic/powder.py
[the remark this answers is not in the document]
ajjackson: I think it's mostly there for humans, and this notation is even better.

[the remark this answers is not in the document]
ajjackson:
```suggestion
        scattering_lengths: dict = get_reference_data(
            physical_property='coherent_scattering_length',
            collection=scattering_lengths)
```

oerc0122 (approved):
Looks good!

PR #471, opened 2026-03-18
Fix Changelog check

oerc0122:
Attempts to fix the issue with the changelog check. Should not be merged without reverting the changelog changes. 

Issues were due to misplaced `--` and naming of branch due to not checked out.

### DIFF
--- a/.github/workflows/check_changelog.yml
+++ b/.github/workflows/check_changelog.yml
@@ -17,7 +17,7 @@ jobs:
           set +e  # Don't immediately fail when bash sees exit 1
           git diff --exit-code "origin/${GITHUB_BASE_REF}" -- CHANGELOG.rst
           if [ $? -eq 1 ]; then
-              return 0
+              exit 0
           fi
           # git-diff did not return 1; either no changes or another error encountered
-          return 1
+          exit 1
--- a/.github/workflows/check_changelog.yml
+++ b/.github/workflows/check_changelog.yml
@@ -14,4 +14,4 @@ jobs:
           fetch-depth: 0
       - name: Get changes
         run: |
-          ! git diff --exit-code "${GITHUB_BASE_REF}" -- CHANGELOG.rst
+          ! git diff --exit-code "origin/${GITHUB_BASE_REF}" -- CHANGELOG.rst
--- a/.github/workflows/check_changelog.yml
+++ b/.github/workflows/check_changelog.yml
@@ -14,4 +14,10 @@ jobs:
           fetch-depth: 0
       - name: Get changes
         run: |
-          ! git diff --exit-code "origin/${GITHUB_BASE_REF}" -- CHANGELOG.rst
+          set +e  # Don't immediately fail when bash sees exit 1
+          git diff --exit-code "origin/${GITHUB_BASE_REF}" -- CHANGELOG.rst
+          if [ $? -eq 1 ]; then
+              return 0
+          fi
+          # git-diff did not return 1; either no changes or another error encountered
+          return 1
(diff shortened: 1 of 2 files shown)


CHANGELOG.rst
```diff
@@ -1,3 +1,5 @@
+This is definitely a change the changelog needs.
+
```
ajjackson:
```suggestion
```
The whitespace change below should be enough for it to pass here

ajjackson (approved):

oerc0122:
I am slightly concerned that this passes if the `git diff` fails. I'm trying to work out how to stop that.

ajjackson:
> I am slightly concerned that this passes if the `git diff` fails. I'm trying to work out how to stop that.

I think for most failures the exit code will be something other than 1; 1 should be reserved for "yes there is a diff". So we need to somehow check the exit code is exactly 1.

oerc0122:
> > I am slightly concerned that this passes if the `git diff` fails. I'm trying to work out how to stop that.
> 
> I think for most failures the exit code will be something other than 1; 1 should be reserved for "yes there is a diff". So we need to somehow check the exit code is exactly 1.

The thing I considered was

```bash
test -n $(git diff --name-only "origin/${GITHUB_BASE_REF}" -- CHANGELOG.rst)
```
Which should work.

.github/workflows/check_changelog.yml
```diff
@@ -14,4 +14,4 @@ jobs:
           fetch-depth: 0
       - name: Get changes
         run: |
-          ! git diff --exit-code -- "${GITHUB_BASE_REF}" "${GITHUB_HEAD_REF}" CHANGELOG.rst
+          ! git diff --exit-code "origin/${GITHUB_BASE_REF}" -- CHANGELOG.rst
```
ajjackson: How about something like
```
set +e  # Don't immediately fail when bash sees exit 1
git diff --exit-code "origin/${GITHUB_BASE_REF}" -- CHANGELOG.rst
if [ $? -eq 1 ]; then
    return 0
fi
# git-diff did not return 1; either no changes or another error encountered
return 1
```

oerc0122:
```suggestion
          set +e  # Don't immediately fail when bash sees exit 1
          git diff --exit-code "origin/${GITHUB_BASE_REF}" -- CHANGELOG.rst
          if [ $? -eq 1 ]; then
              return 0
          fi
          # git-diff did not return 1; either no changes or another error encountered
          return 1
```

.github/workflows/check_changelog.yml
```diff
@@ -14,4 +14,10 @@ jobs:
...
       - name: Get changes
         run: |
-          ! git diff --exit-code -- "${GITHUB_BASE_REF}" "${GITHUB_HEAD_REF}" CHANGELOG.rst
+          set +e  # Don't immediately fail when bash sees exit 1
+          git diff --exit-code "origin/${GITHUB_BASE_REF}" -- CHANGELOG.rst
+          if [ $? -eq 1 ]; then
+              return 0
+          fi
+          # git-diff did not return 1; either no changes or another error encountered
+          return 1
```
oerc0122:
```suggestion
              exit 0
          fi
          # git-diff did not return 1; either no changes or another error encountered
          exit 1
```

ajjackson: 🤦 Of course, `return` is for functions

ajjackson:
Looking good, testing the "no change to CHANGELOG" case: https://github.com/pace-neutrons/Euphonic/actions/runs/23269137658/job/67657430745

PR #474, opened 2026-04-07
Drop redundant get_args function from cli.utils

### LINKED ISSUE #399: Remove cli.utils.get_args()
This function is unnecessary as argparse `get_args()` already has an appropriate default value.

It's a breaking change, as someone could plausibly have copied code from one of the euphonic CLI tools and expect this API to work.

ajjackson:
Closes #399 

This function was previously simplified to a one-liner, but kept in API to avoid compatibility break. For v2 release we can drop it.


oerc0122 (approved):
Nice cleanup. Always good to improve code locality.

PR #475, opened 2026-04-09
Refactor _check_constructor_inputs to transpose parameters

### LINKED ISSUE #315: Refactor internal Spectrum construction syntax
Wondering if this would be better passed as a sequence of tuples tying the data together (i.e transposed to what it is now:
```python
_check_constructor_inputs(
    (z_data, Quantity, (-1, -1, -1), "z_data"),
    ...
)
```

_Originally posted by @oerc0122 in https://github.com/pace-neutrons/Euphonic/pull/305#discussion_r1765858043_

oerc0122:
Fixes #315

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -42,6 +42,9 @@
 
 `v1.6.0 <https://github.com/pace-neutrons/Euphonic/compare/v1.5.1...v1.6.0>`_
 
+  - ``_check_constructor_inputs`` refactored to list arguments as sequence of
+    tuples for each input.
+
 - Requirements
 
   - Increased minimum version of SeeK-path to 2.2.1
--- a/euphonic/brille.py
+++ b/euphonic/brille.py
@@ -56,18 +56,18 @@ class BrilleInterpolator:
             Brille documentation for details.
         """
         _check_constructor_inputs(
-            [crystal, grid],
-            [Crystal, [br.BZTrellisQdc, br.BZMeshQdc, br.BZNestQdc]],
-            [(), ()],
-            ['crystal', 'grid'])
+            (crystal, Crystal, (), 'crystal'),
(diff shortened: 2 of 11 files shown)


ajjackson (approved):
Great! This hurts my head a lot less than the existing arrangement. And because it's an implementation detail, involves no changes to test files 😄 

I think we can go ahead and merge unless there's something in particular you wanted me to look at?

oerc0122:
> Great! This hurts my head a lot less than the existing arrangement. And because it's an implementation detail, involves no changes to test files 😄
> 
> I think we can go ahead and merge unless there's something in particular you wanted me to look at?

If you're happy I'm happy, only potential changes would be:
- NamedTuple support class rather than bare tuples (permit bare tuples though)
- Change instead to one call per test

ajjackson:
NamedTuple would be nice! As this is internal API I don't see the usual downside in convenience.

> Change instead to one call per test

By this do you mean that the shape information should be examined at the beginning, so we don't get those second calls with derived information? I suppose the current arrangement is to avoid needing extra validation/error-handling steps if the input doesn't have `shape()` etc.

ajjackson:
Let's merge as-is and play with namedtuple later.

oerc0122:
No, I meant instead of passing `*details` calling `_check_constructor_inputs` multiple times

ajjackson:
Ah right, that's the opposite 😆. I suspect in that case we would be tempted to `map` over a list anyway?

oerc0122:
> Ah right, that's the opposite 😆. I suspect in that case we would be tempted to `map` over a list anyway?

I did consider if a `NamedTuple` was done, a top-level constant template and `TEMPLATE._replace(value=...)` would be a neat way of preserving all the constant information.

E.g.
```python
X_DATA_PROPS = NamedyTupley(..., np.ndarray, (...,), 'x_data')

[...]

    _check_constructor_inputs(X_DATA_PROPS._replace(value=x_data, shape=[(nx,), (nx + 1,)]))
```

PR #476, opened 2026-04-10
Make crystal.{reciprocal_cell,cell_volume} cached props

### LINKED ISSUE #289: Consider making Crystal.cell_volume() a @property
Suggested by @oerc0122  in #288 review

Things would look a little cleaner if this was accessed as a property, and it would be nice if the result was cached.

It's a breaking change though, so should be deferred until v2 if it happens at all.

oerc0122:
Possible risks: setting `cell_vectors` directly as `_cell_vectors` will not clear the caches.

Fixes #289

### DIFF
--- a/euphonic/crystal.py
+++ b/euphonic/crystal.py
@@ -1,4 +1,5 @@
 from collections import OrderedDict
+from functools import cached_property
 from math import ceil
 from pathlib import Path
 from typing import Any, TypeVar
@@ -97,6 +98,10 @@ class Crystal:
     def cell_vectors(self, value: Quantity) -> None:
         self.cell_vectors_unit = str(value.units)
         self._cell_vectors = value.to('bohr').magnitude
+        if hasattr(self, 'cell_volume'):
+            del self.cell_volume
+        if hasattr(self, 'reciprocal_cell'):
+            del self.reciprocal_cell
 
     @property
     def atom_mass(self) -> Quantity:
@@ -113,6 +118,7 @@ class Crystal:
                                ['cell_vectors_unit', 'atom_mass_unit'])
         super().__setattr__(name, value)
 
+    @cached_property
     def reciprocal_cell(self) -> Quantity:
         """
         Calculates the reciprocal lattice vectors
@@ -128,11 +134,12 @@ class Crystal:
         bxc = np.cross(cv[1], cv[2])
         cxa = np.cross(cv[2], cv[0])
         axb = np.cross(cv[0], cv[1])
-        vol = self.cell_volume()
+        vol = self.cell_volume
         norm = 2*np.pi/vol
 
         return np.vstack((norm * bxc, norm * cxa, norm * axb))
 
+    @cached_property
     def cell_volume(self) -> Quantity:
         """
         Calculates the cell volume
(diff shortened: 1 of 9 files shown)


ajjackson:
We could take some extra steps to make `crystal` quasi-immutable.
- frozen dataclass base
- disable write on underlying numpy arrays with `setflags(write=False)`

I'm trying out that combination here: https://github.com/pace-neutrons/abins-lib/blob/4e6286df73be9c44d7095c76dd15be09ba773864/src/abinslib/displacements.py#L36

The `@dataclass` part might not work well with the way this stores arrays/units separately. (I think that is for benefit of C code?) But `@property` can still be used to keep things pretty locked down.

ajjackson:
I think this one needs another rebase, sorry! I rebased release-2.0.0 on master to get the pytest update and that seems to have really confused the diff here.

euphonic/crystal.py
```diff
@@ -97,6 +98,10 @@ def cell_vectors(self) -> Quantity:
     def cell_vectors(self, value: Quantity) -> None:
         self.cell_vectors_unit = str(value.units)
         self._cell_vectors = value.to('bohr').magnitude
+        if hasattr(self, 'cell_volume'):
+            del self.cell_volume
+        if hasattr(self, 'reciprocal_cell'):
+            del self.reciprocal_cell
```
ajjackson, line 104: I find this `del` syntax to clear cache pretty weird, but that's what the Python docs say!

tests_and_analysis/test/euphonic_test/test_powder.py
```diff
@@ -127,12 +127,12 @@ def mock_qpf(self, mocker, mock_s, mock_dw):
     @pytest.fixture
     def mock_crystal(self, mocker):
         crystal = mocker.MagicMock()
+        crystal.reciprocal_cell = (np.array([[1, 0, 0],
+                                            [0, 1, 0],
+                                            [0, 0, 1]])
+                                   * ureg('1 / angstrom'))
```
ajjackson, line 133: This could be
```python
Quantity(np.eye(3), '1 / angstrom')
```

Any thoughts on `eye` vs explicit array?

ajjackson (commented):
This is a really nice little API improvement for v2 😄 

Any thoughts on the idea to make the lattice vectors immutable while we are at it?

tests_and_analysis/test/euphonic_test/test_powder.py
oerc0122: I like `eye`, but I was keeping the original (minimal changes). More than happy to adjust.

ajjackson: Let's leave it for now.

At some point I should do a sweep of all instances of `* ureg` in Euphonic as building the `Quantity` seems to have better performance (by dodging some unnecessary checks).

euphonic/powder.py
```diff
@@ -37,7 +37,7 @@ def sample_sphere_dos(fc: ForceConstants,
                       npts: int = 1000,
                       jitter: bool = False,
                       rng: RNG = rng,
-                      energy_bins: Quantity | None = None,
+                      energy_bins: Quantity = None,
```
ajjackson: Looks like a regression sneaking in here

euphonic/powder.py
```diff
@@ -130,7 +130,7 @@ def sample_sphere_pdos(
         npts: int = 1000,
         jitter: bool = False,
         rng: RNG = rng,
-        energy_bins: Quantity | None = None,
+        energy_bins: Quantity = None,
```
ajjackson: Another one here
```suggestion
        energy_bins: Quantity | None = None,
```

ajjackson (approved):
Great! We'll play with immutability in another branch

PR #477, opened 2026-04-10
Add NamedTuple for checking constructor inputs

oerc0122:
Makes things consistent, still debatable if it makes things clearer. 

C.f. euphonic/qpoint_phonon_modes.py for alternative style (will need to be made consistent before merging, hence draft)

### DIFF
--- a/euphonic/qpoint_phonon_modes.py
+++ b/euphonic/qpoint_phonon_modes.py
@@ -84,10 +84,10 @@ class QpointPhononModes(QpointFrequencies):
         # Check freqs axis 1 shape here - QpointFrequencies doesn't
         # enforce that the number of modes = 3*(number of atoms)
         _check_constructor_inputs(
-            [frequencies, eigenvectors],
-            [Quantity, np.ndarray],
-            [(n_qpts, 3*n_at), (n_qpts, 3*n_at, n_at, 3)],
-            ['frequencies', 'eigenvectors'])
+            (frequencies, Quantity, (n_qpts, 3*n_at), 'frequencies'),
+            (eigenvectors, np.ndarray,
+             (n_qpts, 3*n_at, n_at, 3), 'eigenvectors'),
+        )
         self.eigenvectors = eigenvectors
 
     def reorder_frequencies(self,
--- a/euphonic/qpoint_phonon_modes.py
+++ b/euphonic/qpoint_phonon_modes.py
@@ -16,7 +16,13 @@ from euphonic.readers import castep, phonopy
 from euphonic.spectra import Spectrum1DCollection
 from euphonic.structure_factor import StructureFactor
 from euphonic.ureg import Quantity, ureg
-from euphonic.util import direction_changed, get_reference_data, is_gamma
+from euphonic.util import (
+    comma_join,
+    direction_changed,
+    format_error,
+    get_reference_data,
+    is_gamma,
+)
 from euphonic.validate import _check_constructor_inputs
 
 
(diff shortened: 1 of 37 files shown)


euphonic/spectra/collections.py
```diff
@@ -44,6 +45,11 @@
 LineData = Sequence[OneLineData]
 Metadata = dict[str, str | int | LineData]
 
+X_TICK_CHECK = InputCheck(..., (list, NoneType),
+                          {}, 'x_tick_labels')
+METADATA_CHECK = InputCheck(..., (dict, NoneType),
+                            {}, 'metadata')
+
```
ajjackson: It's ok for `spectra.collections` to import things from `.base` for consistency; this is already necessary to get classes of component items.

euphonic/qpoint_phonon_modes.py
```diff
@@ -23,7 +23,10 @@
     get_reference_data,
     is_gamma,
 )
-from euphonic.validate import _check_constructor_inputs
+from euphonic.validate import InputCheck, _check_constructor_inputs
+
+FREQ_CHECK = InputCheck(..., (Quantity,), {(-1,)}, 'frequencies')
```
ajjackson: This one could be defined in `qpoint_frequencies` and imported/re-used here: QpointPhononModes inherits QpointFrequencies and has the same expectations for frequency inputs.

euphonic/validate.py
```diff
@@ -8,35 +8,39 @@
...
 
+class InputCheck(NamedTuple):
+    """Data structure for checking inputs."""
+
+    #: The object to check.
+    value: object
+
+    #: The expected class of the input. If multiple types are
+    #: accepted, the expected class can be a list of types. e.g.
+    #: types=(list, np.ndarray).
```
ajjackson:
```suggestion
    #: accepted, the expected class can be a tuple of types. e.g.
    #: types=(list, np.ndarray).
```

euphonic/validate.py
```diff
@@ -8,35 +8,39 @@
...
+    typ: tuple[type, ...]
+
+    #: The expected shape of the object (if the object has a shape
+    #: attribute). If the shape of some dimensions don't matter,
+    #: provide -1 for those dimensions, or if none of the dimensions
+    #: matter, provide an empty tuple (). If multiple shapes are
+    #: accepted, the expected shapes can be a list of tuples. e.g.
+    #: shape=[(n, 3), (n + 1, 3)].
+    shape: Iterable[tuple[int, ...]]
```
ajjackson: In practice you've been going with `set` but indeed any Iterable works with current implementation. I'd be tempted to recommend `set` in the comments rather than list, while sticking with the Iterable signature.

euphonic/validate.py
```diff
@@ -8,35 +8,39 @@
 from euphonic.util import comma_join, format_error
 
 
+class InputCheck(NamedTuple):
+    """Data structure for checking inputs."""
```
ajjackson: Do we need to get the comments into this docstring to make them accessible through documentation outside the source code?

ajjackson (commented):
This is looking really good! A bit more strictness is yielding simpler logic, and the system is generally more self-documenting without gaining much verbosity. The mixture of "templated" instances and fully-defined instances seems fine; I wouldn't like to impose the boilerplate of always defining them away from point of use, but common imports create a clear link between things that _should_ be the same.

I think the natural place for most explanation to go is in the InputCheck docstring; other places the system is encountered will draw the curious developer to that class.

euphonic/validate.py
oerc0122: `#:` **should** pull them out. Not 100% though, will check.

ajjackson: `help()` is giving me the type annotations but nothing else:

```
Help on class InputCheck in module euphonic.validate:

class InputCheck(builtins.tuple)
 |  InputCheck(
 |      value: object,
 |      typ: tuple[type, ...],
 |      shape: Iterable[tuple[int, ...]],
 |      name: str
 |  )
 |
 |  Data structure for checking inputs.
 |
 |  Method resolution order:
 |      InputCheck
 |      builtins.tuple
 |      builtins.object
 |
 |  Methods defined here:
 |
 |  __getnewargs__(self) from collections.InputCheck
 |      Return self as a plain tuple.  Used by copy and pickle.
```

ajjackson (approved):

PR #478, opened 2026-04-13
Add mccabe complexity check

ajjackson:
Setting an initial threshold of 20, which is higher than default but not terrible.

For now I've given the single biggest/ugliest function (force constants Fourier interpolation) a noqa as I don't want it to limit the general target level but also have little appetite to overhaul it right now.

The CLI utils logic to create Argparse parsers was the next big offender. This has been "fixed" by extracting some code to methods. I don't love this approach as it feels like gaming the metric rather than fixing the underlying design. But it does make sense that this function is a big pile of logic, and the extraction does at least make it clear that only a few pieces of data are really interacting.

It also makes it obvious that this function is dominating `euphonic.cli.utils` so the natural response is to factor out to another file.

### DIFF
--- a/euphonic/cli/powder_map.py
+++ b/euphonic/cli/powder_map.py
@@ -140,7 +140,7 @@ def _validate_args(args: Namespace) -> None:
 
 
 def _get_e_max(args: Namespace) -> float:
-    if args.e_i is not None and args.e_max is None:
+    if args.e_max is None:
         return args.e_i
     return args.e_max
 
--- a/euphonic/cli/powder_map.py
+++ b/euphonic/cli/powder_map.py
@@ -154,36 +154,38 @@ def _get_spectrum_calculator(
 ) -> Callable[[Quantity, int], Spectrum1D]:
     """Get the appropriate Spectrum1D calculator for loop over Q"""
 
-    if cli_args.weighting == 'dos' and cli_args.pdos is None:
-        def calc_spectrum(q: Quantity, npts: int) -> Spectrum1D:
-            return sample_sphere_dos(
-                fc, q, npts=npts, **kwargs)
-
-    elif 'dos' in cli_args.weighting:
-        def calc_spectrum(q: Quantity, npts: int) -> Spectrum1D:
-            spectrum_1d_col = sample_sphere_pdos(
+    match cli_args.weighting, cli_args.pdos:
+        case 'coherent', _:
+            def calc_spectrum(q: Quantity, npts: int) -> Spectrum1D:
+                return sample_sphere_structure_factor(
                     fc,
                     q,
                     npts=npts,
-                    weighting=_get_pdos_weighting(cli_args.weighting),
+                    dw=dw,
+                    temperature=temperature,
                     **kwargs)
-            return _arrange_pdos_groups(spectrum_1d_col, cli_args.pdos)
+            return calc_spectrum
 
-    elif cli_args.weighting == 'coherent':
-        def calc_spectrum(q: Quantity, npts: int) -> Spectrum1D:
-            return sample_sphere_structure_factor(
-                fc,
-                q,
-                npts=npts,
-                dw=dw,
-                temperature=temperature,
-                **kwargs)
+        case 'dos', None:
+            def calc_spectrum(q: Quantity, npts: int) -> Spectrum1D:
+                return sample_sphere_dos(
+                    fc, q, npts=npts, **kwargs)
+            return calc_spectrum
 
-    else:
-        msg = f'Cannot interpret weighting "{cli_args.weighting}".'
-        raise ValueError(msg)
+        case weighting, pdos if 'dos' in weighting:
+            def calc_spectrum(q: Quantity, npts: int) -> Spectrum1D:
+                spectrum_1d_col = sample_sphere_pdos(
+                        fc,
+                        q,
+                        npts=npts,
+                        weighting=_get_pdos_weighting(weighting),
+                        **kwargs)
+                return _arrange_pdos_groups(spectrum_1d_col, pdos)
+            return calc_spectrum
 
-    return calc_spectrum
+        case _, __:
+            msg = f'Cannot interpret weighting "{cli_args.weighting}".'
+            raise ValueError(msg)
 
 
 def main(params: list[str] | None = None) -> None:
--- a/euphonic/cli/powder_map.py
+++ b/euphonic/cli/powder_map.py
@@ -1,4 +1,4 @@
-from argparse import ArgumentParser
+from argparse import ArgumentParser, Namespace
 from collections.abc import Callable, Sequence
 from math import ceil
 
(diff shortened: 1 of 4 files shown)


ajjackson:
@oerc0122 What do you think?

Unfortunately the kind of method-extraction mccabe promotes tends to lead to large/awkward diffs :( Even worse in cases like this, where I realise that the reason I don't want to extract methods is that the file is too large - and so they should be extracted to new files.

euphonic/cli/utils/__init__.py
oerc0122, line 1: Never a huge fan of lots of definitions in the `__init__`. Would prefer these broken up a bit and re-exported in the `__init__` if necessary.

euphonic/cli/powder_map.py
```diff
@@ -148,6 +138,71 @@ def main(params: list[str] | None = None) -> None:
...
+                spectrum_1d_col = sample_sphere_pdos(
+                        fc,
+                        q,
+                        npts=npts,
+                        weighting=_get_pdos_weighting(weighting),
+                        **kwargs)
+                return _arrange_pdos_groups(spectrum_1d_col, pdos)
+            return calc_spectrum
+
+        case _, __:
```
oerc0122:
```suggestion
        case _:
```
Will do for "match anything"

euphonic/cli/powder_map.py
```diff
@@ -148,6 +138,71 @@ def main(params: list[str] | None = None) -> None:
         )
         raise ValueError(msg)
 
+
+def _get_e_max(args: Namespace) -> float:
+    if args.e_i is not None and args.e_max is None:
+        return args.e_i
```
oerc0122:
```suggestion
    if args.e_max is None:
        return args.e_i
```
Can simplify the logic since either path would result in `None`

euphonic/cli/utils/__init__.py
oerc0122: I'm also assuming all of these have been pulled out of `_cli_parser.py` verbatim.

euphonic/cli/utils/_cli_parser.py
```diff
@@ -553,6 +61,143 @@ def _get_cli_parser(features: Collection[str] = {},
...
+    _setup_filename_arg(features, sections)
+    _setup_fc_args(features, sections)
+    _setup_pdos_ins_args(features, sections)
+    _setup_powder_args(features, sections)
+    _setup_plotting_args(features, sections)
+    _setup_ebins_adaptive_broadening_args(features, sections)
+
+    # 'ins-weighting' implies 'mp-grid' as well because mesh is needed for DW
+    # factor
+    if {'ins-weighting', 'mp-grid'}.intersection(features):
```
oerc0122: If you cast features to a `set` explicitly, you can possibly make these nicer. At the moment it's a bit of a Yoda condition.


```python
    if features & {'ins-weighting', 'mp-grid'}:
```

euphonic/cli/utils/_cli_parser.py
```diff
@@ -553,6 +61,143 @@ def _get_cli_parser(features: Collection[str] = {},
...
+            help='The difference between each dipole_parameter to test',
+        )
+
+    return parser, sections
+
+
+def _setup_filename_arg(
+    features: Collection[str], sections: dict[str, _ArgumentGroup],
+) -> None:
     if {'read-fc', 'read-modes'}.intersection(features):
```
oerc0122, line 556: I know it "increases the complexity", but I *much* prefer the conditional in the main block rather than here. It makes it more obvious when debugging which parts are relevant.

euphonic/cli/utils/__init__.py
ajjackson: `euphonic/cli/utils/__init__.py` directly replaces `euphonic/cli/utils.py` (and appears identical to import statements). From this, `_get_cli_parser` was factored out to a new file `euphonic/cli/utils/_cli_parser.py`, then additional methods were broken out of that function into the same file.

ajjackson: Unfortunately the git diff process has got this backwards and thinks that `euphonic/cli/_cli_parser.py` is the new `euphnonic/cli/utils.py` 😞

ajjackson:
How about I revert the _get_cli_args refactor and `#noqa` it here, then it is much easier to read the other stuff.

Then in a separate PR do that refactor and remove the `#noqa`.

euphonic/cli/utils/_cli_parser.py
ajjackson: Yeah it would make them less magical. In some cases it might require a slightly different conditional to be cooked up (because the method spans several conditions), but that's ok.

oerc0122:
> How about I revert the _get_cli_args refactor and `#noqa` it here, then it is much easier to read the other stuff.
> 
> Then in a separate PR do that refactor and remove the `#noqa`.

I don't know, maybe I'm being too whingy about it, but I don't see an increase in clarity there.

ajjackson:
At the moment it's pretty hard to see the parts where I _did_ meaningfully change code because there are stuck between a bunch of "moving stuff" diff.

ajjackson:
Ok, apologies that this negated some of the review suggestions (I can check against them in a later step) but now you have a _much_ nicer diff for the smaller (but significant) function fixes and the nasty one can be done afterwards.

euphonic/cli/powder_map.py
```diff
@@ -173,32 +228,35 @@ def main(params: list[str] | None = None) -> None:
...
+    general_kwargs = {
+        'sampling': args.sampling,
+        'jitter': args.jitter,
+        'energy_bins': energy_bins,
+    }
+
+    calc_spectrum = _get_spectrum_calculator(
+        fc, args, temperature, dw, **general_kwargs, **calc_modes_kwargs)
+
     for q_index in tqdm(range(n_q_bins)):
```
oerc0122, line 260: (Not this PR, but https://tqdm.github.io/docs/tqdm/#trange ?)

oerc0122 (approved):
Looks ok, the main thing this indicates to me is that most of the code is not too complex except where it needs to be, which is reassuring. 

Still not sure `more short functions === less complexity` as a big advocate of code locality there are definitely cases where the core logic should be local for clarity even if that's "More complex", but also I like flat code (hence a lot of overly magical functional programming at times which may or may not be McCabe complex)

euphonic/cli/powder_map.py
ajjackson: I don't think if I saw `trange` in some code I would successfully guess it was a `tqdm` progress bar. Maybe it's a good idea in a codebase that does this more often?

ajjackson: One could import as `tqdm_range` but then there is no space saved and it's literally the same code. https://github.com/tqdm/tqdm/blob/75bdb6c379bcfc6c592b6342dc791a092b5d6ae0/tqdm/std.py#L1522

euphonic/cli/utils/_cli_parser.py
ajjackson: Nice point, would be good to revisit this. Creating an issue. #485

PR #481, opened 2026-04-14
Refactor CLI utils module

ajjackson:
In support of work on mccabe complexity.

It is very difficult to review code that moves _and_ changes simultaneously, so here we separately implement the code _movement_ of this rather large utility module into private submodules.


euphonic/cli/utils/__init__.py
```diff
@@ -0,0 +1,41 @@
...
+    '_get_cli_parser',
+    '_get_debye_waller',
+    '_get_energy_bins',
+    '_get_pdos_weighting',
+    '_get_q_distance',
+    '_get_tick_labels',
+    '_get_title',
+    '_grid_spec_from_args',
+    '_insert_gamma',
+    '_plot_label_kwargs',
```
oerc0122, line 37: Should I be concerned you're exporting a bunch of "private" functions?

euphonic/cli/utils/_band_structure.py
```diff
@@ -0,0 +1,166 @@
...
+
+        [(0, 'L'), (3, 'X'), (5, '$\\Gamma$')]
+    """
+
+    label_indices = np.where(bandpath['explicit_kpoints_labels'])[0]
+    labels = [bandpath['explicit_kpoints_labels'][i] for i in label_indices]
+
+    for i, label in enumerate(labels):
+        if label == 'GAMMA':
+            labels[i] = r'$\Gamma$'
```
oerc0122: Could be done in one step as:
```suggestion
    labels = [val if (val := bandpath['explicit_kpoints_labels'][i]) != 'GAMMA' else r'$\Gamma$' for i in label_indices]
```

Probably better would be:
```suggestion
    from toolz.itertoolz import get
    labels = [
        val if val != 'GAMMA' else r'$\Gamma$' 
        for val in get(label_indices, bandpath['explicit_kpoints_labels'])
    ]
```
Could even do the whole function as:

```python
    from toolz.itertoolz import get
    labels = [
        (ind, val if val != 'GAMMA' else r'$\Gamma$')
        for ind, val in zip(
            label_indices,
            get(label_indices, bandpath['explicit_kpoints_labels']),
            strict=True,
        )
    ]
```
But we're starting to get a little overboard there.

Not sure if any of these is necessarily more legible.

euphonic/cli/utils/_band_structure.py
```diff
@@ -0,0 +1,166 @@
...
+    # Find break points between continuous spectra: wherever there are two
+    # adjacent labels
+    labels = np.array(bandpath['explicit_kpoints_labels'])
+
+    special_point_bools = np.fromiter(
+        map(bool, labels), dtype=bool)
+
+    # [T F F T T F T] -> [F F T T F T] AND [T F F T T F] = [F F F T F F] -> 3,
+    adjacent_non_empty_labels = np.logical_and(special_point_bools[:-1],
+                                               special_point_bools[1:])
```
oerc0122, line 63: Is this clearer?
```suggestion
    adjacent_non_empty_labels = special_point_bools[:-1] &
special_point_bools[1:]
```

euphonic/cli/utils/_band_structure.py
```diff
@@ -0,0 +1,166 @@
...
+
+    adjacent_different_labels = (labels[:-1] != labels[1:])
+
+    break_points = np.where(np.logical_and(adjacent_non_empty_labels,
+                                           adjacent_different_labels))[0]
```
oerc0122, line 68:
```suggestion
    break_points = np.where(adjacent_non_empty_labels &
                            adjacent_different_labels)[0]
```

euphonic/cli/utils/_band_structure.py
```diff
@@ -0,0 +1,166 @@
...
+    if bandpath.get('is_supercell'):
+        _convert_labels_to_fractions(bandpath, limit=32)
+
+    x_tick_labels = _get_tick_labels(bandpath)
+    split_args = {'indices': _get_break_points(bandpath)}
+
+    print(
+        'Computing phonon modes: {n_modes} modes across {n_qpts} q-points'
+        .format(n_modes=(data.crystal.n_atoms * 3),
+                n_qpts=len(bandpath['explicit_kpoints_rel'])))
```
oerc0122, line 155:
```suggestion
        f'Computing phonon modes: {data.crystal.n_atoms * 3} modes '
        f'across {len(bandpath['explicit_kpoints_rel'])} q-points'
```

euphonic/cli/utils/_plotting.py
```diff
@@ -0,0 +1,77 @@
...
+    if user_args.style:
+        style += user_args.style
+
+    # Explicit args take priority over any other
+    explicit_args = {}
+    for user_arg, mpl_property in {'cmap': 'image.cmap',
+                                   'fontsize': 'font.size',
+                                   'font': 'font.sans-serif',
+                                   'linewidth': 'lines.linewidth',
+                                   'figsize': 'figure.figsize'}.items():
```
oerc0122, line 58: Would this mapping be better expressed in a constant?

euphonic/cli/utils/_plotting.py
```diff
@@ -0,0 +1,77 @@
...
+        if getattr(user_args, user_arg, None):
+            explicit_args.update({mpl_property: getattr(user_args, user_arg)})
```
oerc0122, line 60:
```suggestion
        if (arg := getattr(user_args, user_arg, None)) is not None:
            explicit_args.update({mpl_property: arg})
```
Alternatively you could use `vars(user_args)` to get the dict.

euphonic/cli/utils/_plotting.py
```diff
@@ -0,0 +1,77 @@
...
+
+    if 'font.sans-serif' in explicit_args:
+        explicit_args.update({'font.family': 'sans-serif'})
```
oerc0122, line 63:
```suggestion
        explicit_args['font.family'] = 'sans-serif'
```

euphonic/cli/utils/_plotting.py
```diff
@@ -0,0 +1,77 @@
...
+        explicit_args['figure.figsize'] = [dim.to('inches').magnitude
+                                           for dim in dimensioned_figsize]
+
+    style.append(explicit_args)
+    return style
+
+
+def _get_title(filename: str, title: str | None = None) -> str:
+    """Get a plot title: either user-provided string, or from filename"""
+    return title if title is not None else Path(filename).stem
```
oerc0122, line 77: Could be:

```suggestion
    return title or Path(filename).stem
```
However, this will mean `title = ""` becomes filename.

euphonic/cli/utils/_plotting.py
```diff
@@ -0,0 +1,77 @@
...
+    """
+    Save or show the current matplotlib plot.
+    Show if save_filename is not None which by default it is.
+
+    Parameters
+    ----------
+    save_filename
+        The file to save the plot in
+    """
+    import matplotlib.pyplot as plt
```
oerc0122, line 23: Just waiting on that 3.15 `lazy import` huh?

euphonic/cli/utils/_pdos.py
```diff
@@ -0,0 +1,50 @@
...
+    else:
+        idx = cl_arg_weighting.rfind('-')
+        if idx == -1:
+            msg = format_error(
+                f'Unexpected weighting "{cl_arg_weighting}"',
+                fix='Check weighting argument. Should be e.g. "coherent-dos".',
+            )
+            raise ValueError(msg)
+        pdos_weighting = cl_arg_weighting[:idx]
+    return pdos_weighting
```
oerc0122, line 50:
```suggestion
    if cl_arg_weighting == 'dos':
        return None
    
    idx = cl_arg_weighting.rfind('-')
    if idx == -1:
        msg = format_error(
            f'Unexpected weighting "{cl_arg_weighting}"',
            fix='Check weighting argument. Should be e.g. "coherent-dos".',
       )
        raise ValueError(msg)
        
    return cl_arg_weighting[:idx]
```

euphonic/cli/utils/_loaders.py
```diff
@@ -0,0 +1,161 @@
...
+            reason=f"""
+            CASTEP force constants data for
+            import should have extension from {castep_fc_suffixes}, CASTEP
+            phonon mode data for import should have extension
+            '{castep_qpm_suffixes}', data from Phonopy should have extension
+            from {phonopy_suffixes}, data from Euphonic should have extension
+            '.json'.""",
+            fix='Ensure file format in known formats.',
+        )
+        raise ValueError(msg)
```
oerc0122, line 158:
```suggestion
    castep_qpm_suffixes = ('.phonon',)
    castep_fc_suffixes = ('.castep_bin', '.check')
    phonopy_suffixes = ('.hdf5', '.yaml', '.yml')

    path = Path(filename)
    match path.suffix:
        case '.phonon' if frequencies_only:
            data = QpointFrequencies.from_castep(path)
        case '.phonon':
            data = QpointPhononModes.from_castep(path)
        case '.castep_bin' | '.check':
            data = ForceConstants.from_castep(path)
        case '.json':
            data = _load_euphonic_json(path, frequencies_only)
        case '.hdf5' | '.yaml' | '.yml':
            data = _load_phonopy_file(path, frequencies_only)
        case _:
            msg = format_error(
                 f'File format ({path.suffix}) not recognised.',
                 reason=f"""
                 CASTEP force constants data for
                 import should have extension from {castep_fc_suffixes}, CASTEP
                 phonon mode data for import should have extension
                 '{castep_qpm_suffixes}', data from Phonopy should have extension
                 from {phonopy_suffixes}, data from Euphonic should have extension
                 '.json'.""",
                 fix='Ensure file format in known formats.',
            )
            raise ValueError(msg)
```

euphonic/cli/utils/_loaders.py
```diff
@@ -0,0 +1,161 @@
...
+
+    If the filename follows the pattern "seedname-phonopy.yml" this will be
+    "seedname-force_constants.hdf5" in the same directory.
+
+    Otherwise, return Path.cwd(), which will fail an .is_file() check.
+    """
+
+    re_match = re.match(r'(?P<seedname>.+)-phonopy\.(?P<ext>ya?ml)',
+                        phonopy_file.name)
+    if re_match:
```
oerc0122, line 104:
```suggestion
    if re_match := re.match(r'(?P<seedname>.+)-phonopy\.(?P<ext>ya?ml)', phonopy_file.name):
```

euphonic/cli/utils/_loaders.py
```diff
@@ -0,0 +1,161 @@
...
+        seedname = re_match.group('seedname')
+        return Path(phonopy_file.parent / f'{seedname}-force_constants.hdf5')
```
oerc0122, line 106:
```suggestion
        return Path(phonopy_file.parent, f'{seedname}-force_constants.hdf5')
```

euphonic/cli/utils/_loaders.py
```diff
@@ -0,0 +1,161 @@
...
+                    ),
+                    fix='Ensure phonopy.yaml provided.',
+                )
+                raise ValueError(msg)
+        elif path.suffix in ('.yaml', '.yml'):
+            phonopy_kwargs['summary_name'] = path.name
+            # Assume this is a (renamed?) phonopy.yaml file
+            if (janus_fc := _janus_fc_filename(path)).is_file():
+                phonopy_kwargs['fc_name'] = janus_fc.name
+            elif (path.parent / 'force_constants.hdf5').is_file():
```
oerc0122, line 84:
```suggestion
            elif path.with_name('force_constants.hdf5').is_file():
```

euphonic/cli/utils/_loaders.py
```diff
@@ -0,0 +1,161 @@
...
+            loaded_data = QpointFrequencies.from_phonopy(
+                path=path.parent, phonon_name=path.name)
+
+    if loaded_data is None:
+        phonopy_kwargs: dict[str, str | os.PathLike] = {}
+        phonopy_kwargs['path'] = path.parent
+        if (path.parent / 'BORN').is_file():
+            phonopy_kwargs['born_name'] = 'BORN'
+        # Set summary_name and fc_name depending on input file
+        if path.suffix == '.hdf5':
```
oerc0122, line 64: `match`?

euphonic/cli/utils/_loaders.py
```diff
@@ -0,0 +1,161 @@
...
+    if 'frequencies' in data:
+        if 'eigenvectors' in data and not frequencies_only:
+            return QpointPhononModes.from_json_file(filename)
+        return QpointFrequencies.from_json_file(filename)
+
+    msg = format_error(
+        f'Could not identify Euphonic data in JSON file ({filename}).',
+        fix='Ensure JSON file contains "force_constants" or "frequencies".',
+    )
+    raise ValueError(msg)
```
oerc0122, line 34:
```suggestion
    match data:
        case {'force_constants': fc}:
            return ForceConstants.from_json_file(filename)
        case {'frequencies': freq, 'eigenvectors': eig} if not frequencies_only:
            return QpointPhononModes.from_json_file(filename)
        case {'frequencies': freq}:
            return QpointFrequencies.from_json_file(filename)
        case _:
             msg = format_error(
                 f'Could not identify Euphonic data in JSON file ({filename}).',
                 fix='Ensure JSON file contains "force_constants" or "frequencies".',
             )
             raise ValueError(msg)
```

euphonic/cli/utils/_kwargs.py
```diff
@@ -0,0 +1,39 @@
...
+                       default_ylabel: str = '') -> dict[str, str]:
+    """Collect title/label arguments that can be passed to plot_nd
+    """
+    plot_kwargs = {'title': args.title,
+                   'xlabel': default_xlabel,
+                   'ylabel': default_ylabel}
+    if args.ylabel is not None:
+        plot_kwargs['ylabel'] = args.ylabel
+    if args.xlabel is not None:
+        plot_kwargs['xlabel'] = args.xlabel
```
oerc0122, line 17:
```suggestion
    plot_kwargs = {'title': args.title,
                   'xlabel': getattr(args, 'xlabel', None) or default_xlabel,
                   'ylabel': getattr(args, 'ylabel', None) or default_ylabel}
```

euphonic/cli/utils/_grids.py
```diff
@@ -0,0 +1,75 @@
...
+        msg = format_error(
+            'Maximum energy should be greater than minimum.',
+            fix='Check --e-min and --e-max arguments.',
+        )
+        raise ValueError(msg)
+    return np.linspace(emin, emax, n_ebins) * modes.frequencies.units
+
+
+def _grid_spec_from_args(crystal: Crystal,
+                           grid: Sequence[int] | None = None,
```
oerc0122, line 67: `grid` strictly only needs to be `Iterable` based on its usage.

euphonic/cli/utils/_dw.py
```diff
@@ -0,0 +1,30 @@
...
+                      grid: Sequence[int] | None = None,
+                      grid_spacing: Quantity = 0.1 * ureg('1/angstrom'),
+                      **calc_modes_kwargs,
+                      ) -> DebyeWaller:
+    """Generate Debye-Waller data from force constants and grid specification
+    """
+    mp_grid_spec = _grid_spec_from_args(fc.crystal, grid=grid,
+                                        grid_spacing=grid_spacing)
+    print('Calculating Debye-Waller factor on {} q-point grid'
+          .format(' x '.join(map(str, mp_grid_spec))))
```
oerc0122, line 27:
```suggestion
    print(f'Calculating Debye-Waller factor on {" x ".join(map(str, mp_grid_spec))} q-point grid'
```

oerc0122 (commented):
Code looks good overall but failing. Many of these may not want to go in this PR, since I think we want this to mostly be a transcription, but can get away with some more modern Python features for potentially better expression.

euphonic/cli/utils/__init__.py
ajjackson: Maybe. It keeps the linter happy because otherwise it sees these things being imported and unused – but we know that other modules need to import them from here. The underscores are supposed to indicate that they are not public API.

Any other conventions that would make more sense?

oerc0122: Not that I'm aware of. 

The problem is they *are* part of the public API (at least as I understand it in python terms: they're used by other modules), just not the user API, but Python has real concept of public/private/anything else. 

I'm ok with it, it's just a mild code-smell I thought I should ask.

euphonic/cli/utils/_band_structure.py
ajjackson: I think we can use `np.take` instead of itertoolz.get; it looks more consistent with the prior call to `np.where`.

That gives something like:
```python
def _get_tick_labels(bandpath: dict) -> list[tuple[int, str]]:
    """Convert x-axis labels from seekpath format to euphonic format

    i.e.::

        ['L', '', '', 'X', '', 'GAMMA']   -->

        [(0, 'L'), (3, 'X'), (5, '$\\Gamma$')]
    """
    label_indices = np.where(bandpath['explicit_kpoints_labels'])[0]
    labels = ("$\Gamma$" if label == "GAMMA" else label
              for label in
              np.take(bandpath['explicit_kpoints_labels'], label_indices))
    return list(zip(label_indices, labels, strict=True))

```
which seems decent.

euphonic/cli/utils/_band_structure.py
ajjackson: Not with that linebreak! But
```python
    adjacent_non_empty_labels = (
        special_point_bools[:-1] & special_point_bools[1:]
    )
```
would be nice.

ajjackson: Similarly, below that we could have
```python
    break_points = np.where(
        adjacent_non_empty_labels & adjacent_different_labels
    )[0]
```

euphonic/cli/utils/_band_structure.py
oerc0122: Always forget numpy has this.

euphonic/cli/utils/_band_structure.py
ajjackson: I think this is a case where the .format() string is much nicer to read as you get an immediate sense of the final message without being lost in implementation.

euphonic/cli/utils/_band_structure.py
ajjackson: I think I'll commit this as it's tested :-)

euphonic/cli/utils/_plotting.py
ajjackson:
```python
    from operator import is_none
    from typing import Any
    from toolz.dicttoolz import valfilter

    mapping = {'cmap': 'image.cmap',
               'fontsize': 'font.size',
               'font': 'font.sans-serif',
               'linewidth': 'lines.linewidth',
               'figsize': 'figure.figsize'}

    def is_not_none(item: Any) -> bool:   # Python 3.14 adds this to operator
        return not is_none(item)

    # Explicit args take priority over any other
    args_dict = valfilter(is_not_none, vars(user_args))

    explicit_args = {
        mapping[key]: args_dict[key]
        for key in (args_dict.keys() & mapping.keys())
    }
```
Let's note that for another PR #486

oerc0122: Either that or `ChainMap` for priorities.

euphonic/cli/utils/_plotting.py
ajjackson: I don't think it makes much difference in this implementation but works as part of something like #486

euphonic/cli/utils/_plotting.py
ajjackson: Agree, let's tidy with rest of this function #486

euphonic/cli/utils/_plotting.py
ajjackson: YUP

euphonic/cli/utils/_plotting.py
ajjackson: So many options!

euphonic/cli/utils/_plotting.py
ajjackson: Although matplotlib is an optional dependency so there might still be need for something more defensive there.

euphonic/cli/utils/_plotting.py
ajjackson: Something funny is going on there because the default behaviour of `euphonic-dispersion` seems to be a blank title, but according to the parser the default value is None. More investigation needed.

Follow-up issue #487

euphonic/cli/utils/_loaders.py
ajjackson: Broken out to #488

euphonic/cli/utils/_dw.py
ajjackson: Again the `format` version seems much easier to read?

euphonic/cli/utils/_grids.py
ajjackson: If this were directly user-facing there would be a `len(grid) == 3` check which requires Sequence. In fact the argparse parser ensures it is nargs=3 so this is always a... tuple, I think?

Not sure how to proceed here

oerc0122: Maybe be explicit about `tuple[int, int, int] | None`?

euphonic/cli/utils/_loaders.py
ajjackson: Much nicer!

euphonic/cli/utils/_loaders.py
ajjackson: Follow-up: #488

euphonic/cli/utils/_loaders.py
ajjackson: I think this should just be
```python
return phonopy_file.parent / f'{seedname}-force_constants.hdf5'
```

euphonic/cli/utils/_grids.py
ajjackson: Yeah, it's internal API so that should be fine.

euphonic/cli/utils/_loaders.py
oerc0122: Nice to be able to use matches to some of their intent.

euphonic/cli/utils/_grids.py
ajjackson: It turns out to be a list in practice. Will sort out early-return logic as well:
```python
def _grid_spec_from_args(crystal: Crystal,
                           grid: list[int] | None = None,
                           grid_spacing: Quantity = 0.1 * ureg('1/angstrom'),
                           ) -> tuple[int, int, int]:
    """Get Monkorst-Pack mesh divisions from user arguments"""
    if grid:
        return tuple(grid)

    return crystal.get_mp_grid_spec(spacing=grid_spacing)
```

euphonic/cli/utils/_loaders.py
oerc0122: Or `phonopy_file.with_name(f'{seedname}-force_constants.hdf5')`

ajjackson: Done! (And a couple of similar places.)

euphonic/cli/utils/_band_structure.py
```diff
@@ -0,0 +1,164 @@
...
+
+
+def _get_tick_labels(bandpath: dict) -> list[tuple[int, str]]:
+    """Convert x-axis labels from seekpath format to euphonic format
+
+    i.e.::
+
+        ['L', '', '', 'X', '', 'GAMMA']   -->
+
+        [(0, 'L'), (3, 'X'), (5, '$\\Gamma$')]
```
oerc0122, line 28: Would this be better as doctest example? (not this PR)

euphonic/cli/utils/_band_structure.py
```diff
@@ -0,0 +1,164 @@
...
+XTickLabels = list[tuple[int, str]]
+SplitArgs = dict[str, Any]
+
+
+# Dictionary returned by seekpath.get_explicit_k_path_orig_cell
+# Not a complete specification, but these are the parts we care about.
+class BandpathDict(TypedDict, total=False):
+    explicit_kpoints_labels: Sequence[str]
+    explicit_kpoints_rel: Iterable[float]
+    is_supercell: bool
```
oerc0122: Strictly as written this is part of the public interface.
```suggestion

class BandpathDict(TypedDict, total=False):
   """Dictionary returned by seekpath.get_explicit_k_path_orig_cell.

    Not a complete specification, but these are the parts we care about."""
    
    explicit_kpoints_labels: Sequence[str]
    explicit_kpoints_rel: Iterable[float]
    is_supercell: bool
```

euphonic/cli/utils/_pdos.py
```diff
@@ -0,0 +1,50 @@
...
+
+    idx = cl_arg_weighting.rfind('-')
+    if idx == -1:
+        msg = format_error(
+            f'Unexpected weighting "{cl_arg_weighting}"',
+            fix='Check weighting argument. Should be e.g. "coherent-dos".',
+       )
+        raise ValueError(msg)
+
+    return cl_arg_weighting[:idx]
```
oerc0122, line 50: I feel this would be best expressed as a `.split` as that's what you want in the end, but not that important.

oerc0122 (approved):
Looks good, most of the things I've noted are probably not for this PR if at all! The only exception is the TypedDict docstring.

euphonic/cli/utils/_band_structure.py
ajjackson: Would be a slightly funny one as it's not a public method and so shouldn't be in the docs. But it would make sense to format consistently with other doctests...

ajjackson: Created issue #490

euphonic/cli/utils/_band_structure.py
ajjackson: Good point!

euphonic/cli/utils/_pdos.py
ajjackson: I think I'd be happier with this as a much stricter mapping, anyway. I guess it works in practice because `argparse` already constrains the choices. Once we drop 3.10 might be a good time to do this stuff with StrEnum instead but it would be compatibility-breaking unless string options are also available. One for v3 perhaps.

ajjackson: Created #491

PR #484, opened 2026-04-15
Make Crystal Immutable

### LINKED ISSUE #483: Immutable Crystal class
Following #476  (cached property access to derived properties of Crystal), we can consider making Crystal immutable (as far as Python allows...)

This removes the need for cache-invalidation logic and streamlines the use of Crystal in other cached calculations.

Users don't need to modify an existing crystal structure very often; this would generally happen earlier in the workflow while energy/force calculations are being performed and the structure is represented by e.g. `ase.Atoms`.

To facilitate modification where necessary we could implement a `__replace__` method and recommend the use of `copy.replace()` in documentation. https://docs.python.org/3/library/copy.html

ajjackson:
Closes #483 

Currently the doctests are failing on the page that describes the unit system. This currently
- declares that _all_ euphonic classes with Quantity attributes allow the unit to be changed by poking the `thing_unit` attribute.
- walks though dealing with the odd behaviour where you can replace a whole array but not always mutate elements of it

Will it confuse matters more to have only _some_ classes following this system? Maybe overhauling all the classes to have simpler property handling (and a `dataclass` base?) is a project for Euphonic v3. I don't see a lot of benefit to the complex mutability of the current system, especially if we provide a `__replace__` hook. The main benefits of current system are:
- direct access to private attributes in atomic units for the C code.
- array size validation

But these could also be handled with dataclass `__post_init__` hooks. And I wonder if the always-convert-to-au-on-init overhead is really worthwhile to simplify the C setup anyway; those things could be wrapped to call a cached conversion method.

### DIFF
--- a/euphonic/crystal.py
+++ b/euphonic/crystal.py
@@ -1,5 +1,5 @@
 from collections import OrderedDict
-from functools import cached_property
+from functools import cached_property, wraps
 from math import ceil
 from pathlib import Path
 from typing import Any, TypeVar
@@ -21,6 +21,18 @@ from euphonic.util import (
 )
 from euphonic.validate import _check_constructor_inputs
 
+
+def read_only_array(f):
+    """Call .setflags(write=False) on return value of decorated function
+
+    This ensures a numpy array isn't accidentally modified by accessing slices
+    """
+    @wraps(f)
+    def wrapper(*args, **kwds):
+        result = f(*args, **kwds)
+        result.setflags(write=False)
+        return result
+    return wrapper
 
 class read_only_cached_property(cached_property):
     def __set__(self, instance, value):
@@ -106,17 +118,13 @@ class Crystal:
         self._cell_vectors_unit = str(cell_vectors.units)
         self._atom_mass_unit = str(atom_mass.units)
 
-        # Set underlying arrays to read-only to defend during slice access
-        for array_attribute in (
-            self.atom_r, self.atom_type, self._cell_vectors, self._atom_mass,
-        ):
-            array_attribute.setflags(write=False)
-
     @property
+    @read_only_array
     def atom_r(self) -> np.ndarray:
         return self._atom_r
 
     @property
+    @read_only_array
     def atom_type(self) -> np.ndarray:
         return self._atom_type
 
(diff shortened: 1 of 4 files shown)


ajjackson:
There is a `noqa` to allow the decorator (implemented as a class) to use a snake_case name. It seems more consistent; the end-user doesn't know/care how the decorator was implemented in this case.

euphonic/crystal.py
```diff
@@ -19,7 +19,38 @@
...
+        else:
+            fix = (f'Make a new {instance.__class__.__name__} with '
+                   'the required values.')
+
+        msg = format_error(
+            f'{instance.__class__.__name__} property '
+            f'"{self.attrname}" is read-only.',
+            fix=fix,
+        )
+        raise AttributeError(msg)
```
oerc0122, line 53: Is this magic more generally usable? Probably doesn't belong in `crystal`, but `utilities`.

euphonic/crystal.py
```diff
@@ -19,7 +19,38 @@
...
+        result = f(*args, **kwds)
+        result.setflags(write=False)
+        return result
+    return wrapper
+
+
+class read_only_cached_property(cached_property):  # noqa: N801
+    """Modified functools.cached_property to disallow __set__"""
+    def __set__(self, instance, value):
+        if hasattr(self, '__replace__'):
```
oerc0122, line 41: Now here's a question I fear the answer to.

`cached_property` creates an attribute of the named property in the `object.__dict__` when first called (hence the need to `del x.prop`), `copy.replace` probably doesn't cull the property in `__dict__`. Is there a mechanism in `copy.replace` which can ensure recomputation?

euphonic/crystal.py
ajjackson: Yup, figured it would get moved as soon as it is used elsewhere. But could be moved from the start. (Assuming we go ahead with this at all!)

euphonic/crystal.py
ajjackson: I think `copy.replace` calls our `__replace__` method, and that method can do whatever it likes! I don't think it would be particularly simple or desirable to copy the cache as well?

A naive-but-ok implementation could be
```python
def __replace__(self, **changes):
    kwargs = {'attr1': self.attr1, 'attr2': self.attr2, ...}
    kwargs.update(changes)
    return type(self)(**kwargs)
```

oerc0122: That does look to be the case, I hadn't noticed the dunder in my looking at copy. Good to know, thanks!

ajjackson: I'll have a crack at implementing it here when we are happy with other issues.

euphonic/crystal.py
oerc0122: It is quite magical. But in utilities, magic can be nicely stowed. The outward decorators are quite clear and descriptive.

ajjackson:
Moving this to v3 milestone and closing for now. I like the idea but would like to apply it to more features at the same time. This puts it beyond reasonable v2 scope, which is supposed to be a collection of small breaking changes.

PR #493, opened 2026-04-29
Minor changes to fix uv sync

oerc0122:
This seems to fix the uv sync issues and should be fully portable. 

Issues:
- Test on different setups
- Whether to remove duplicate `test` group/dep.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -45,6 +45,13 @@ dependencies = [
 ]
 dynamic = ["version"]
 
+[dependency-groups]
+test = ["mock", "pytest>=9.0.3,<10.0", "pytest-cov", "pytest-mock", "pytest-lazy-fixtures", "pytest-xdist", "pytest-xvfb", "python-slugify"]
+dev = [
+    {include-group = "test"},
+    "ruff>0.15",
+]
+
 [project.urls]
 Homepage = "https://github.com/pace-neutrons/Euphonic"
 Repository = "https://github.com/pace-neutrons/Euphonic.git"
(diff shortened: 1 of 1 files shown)


ajjackson:
I'm still running into trouble here. Do you have suggested incantation to use this in a clean environment?

```
FROM python:3.14

RUN pip install uv

RUN git clone --branch fix-uv-sync --single-branch https://github.com/pace-neutrons/euphonic.git /opt/

WORKDIR /opt/euphonic

# This would be nice
# RUN uv sync
# RUN uv sync --upgrade

# But even this is failing
RUN bash -c 'uv venv && source .venv/bin/activate && uv sync --active --no-editable'
RUN bash -c  'source .venv/bin/activate && uv sync --active --no-editable --upgrade'
```

ajjackson:
Making a record from side chat: the problem seems to be the dynamic versioning which causes an editable install. If that is removed, things work:

```
FROM python:3.14

RUN pip install uv

RUN git clone --branch fix-uv-sync --single-branch https://github.com/pace-neutrons/euphonic.git /opt/euphonic

WORKDIR /opt/euphonic

RUN ls /opt && pwd && ls .
RUN sed -i 's/dynamic = \["version"\]//' pyproject.toml
RUN sed -i 's/name = "Euphonic"/name = "Euphonic"\nversion = "1.8.0"/' pyproject.toml
RUN cat pyproject.toml

RUN uv sync
RUN uv sync --upgrade
```

But we like the dynamic versioning because it means we can tell if a JSON dump came from a development build between full version tags.

ajjackson:
This `uv` issue was the hint we needed, to understand "why" editable builds were running even when under `--no-editable` https://github.com/astral-sh/uv/issues/10996

PR #494, opened 2026-05-05
New interface for isotope data

### LINKED ISSUE #439: Improve handling of reference data: define interface
The handling of atomic reference data in Euphonic has some limitations:

- `QpointPhononModes.calculate_pdos()` takes arguments `weighting: Optional[str] = None` and  `cross_sections: str | dict[str, Quantity] = 'BlueBook'`
- `QpointPhononModes.calculate_structure_factor()` takes argument `scattering_lengths: str | dict[str, Quantity] = 'Sears1992'`

If a `dict[str, Quantity]` is provided, the keys are compared with the atom symbols in Crystal and used to assign corresponding weight values. If a `str` is provided, this identifies a .json data file from which the appropriate property group is extracted, then indexed with symbol in the same way.

This setup cannot really support some more advanced use cases. Isotopic substitution is an important technique in neutron scattering that modifies the scattering properties of the substituted nuclei. In Abins, mass data is used to identify whether a given site is a particular isotope or the standard isotopic mixture. This is more difficult than it sounds to get right: https://github.com/mantidproject/mantid/pull/38448

So it would be good to:
- provide a capability to use atomic masses for assignment of scattering lengths / cross sections / other weights.
- maintain the existing capability to provide an ad-hoc dict of symbols and values as this enables experimentation and customisation
- avoid breaking the existing API until a convenient time
- avoid too much complicated type dispatch

I propose to address this by introducing a common interface with different implementations. Some top-level code might type-dispatch to construct the right implementation, but from there everything is consistent. (And options can be added/deprecated smoothly.) This would take responsibility for matching the data to a set of atom sites:

```python
class SpeciesDataset(Protocol):
    def get_property(crystal: Crystal, property: str) -> Quantity: ...

class UserDictDataset(SpeciesDataset):
    def __init__(self, user_dict: dict[str, Quantity]):
        self._data = user_dict
    def get_property(crystal: Crystal, property: str) -> Quantity:
        # In this case property str is ignored
        return Quantity.from_list([self._data[x] for x in self.crystal.atom_type])

class LegacyJsonDataset(SpeciesDataset):
    def __init__(self, json_file: Path):
        self._filename = str(json_file.resolve())
        with open(json_file) as fd:
            self._data = json.load(fd)

    def get_property(crystal: Crystal, property: str) -> Quantity:
        if property not in self._data:
            raise KeyError(f'Property "{property}" not found in "{self._filename}"')
        return Quantity.from_list([self._data[property][x] for x in self.crystal.atom_type])

class PropertyListDataset(SpeciesDataset):
    """This dataset does not attempt to map property to Crystal sites.

    You must provide the values in the correct order.
    """
    def __init__(self, values: Quantity, property: str):
        self._values = values
        self._property = property

    def get_property(crystal: Crystal, property: str):
        if property != self._property:
            raise KeyError()
        if len(crystal.atom_type) != len(self._values):
            raise IndexError()
        return self._values
```

The same interface could be used for a more sophisticated dataset that, e.g., loads CSV to an in-memory sqlite3 database and queries on symbol _or_ z-number to find candidates to match against the provided atomic mass.

ajjackson:
Closes #439

I suppose the big question is whether this is Euphonic v2, allowing breaking changes, or if we stick with having some compatibility layers for smoother transition and get it in before v2.

(I do actually need it quite soon.)

### DIFF
--- a/euphonic/data/isotopes.py
+++ b/euphonic/data/isotopes.py
@@ -176,7 +176,7 @@ class LegacyJsonData(AtomTypeDictData):
 
         if key not in self._data:
             msg = format_error(
-                f'Property {key!r} not found in "{self._collection}"',
+                f'Property {key!r} not found in {self._collection!r}',
                 fix=(f'Available keys: {comma_join(self._data.keys())}.'),
             )
             raise KeyError(msg)
--- a/euphonic/data/isotopes.py
+++ b/euphonic/data/isotopes.py
@@ -63,8 +63,8 @@ class ArrayFromValuesMixin:
         # instead create bare numpy array of appropriate type then add units.
 
         dtype = (
-            complex 
-            if any(isinstance(item.magnitude, complex) for item in items) 
+            complex
+            if any(isinstance(item.magnitude, complex) for item in items)
             else float
         )
         units = items[0].units
--- a/euphonic/data/isotopes.py
+++ b/euphonic/data/isotopes.py
@@ -62,12 +62,16 @@ class ArrayFromValuesMixin:
         # Quantity.from_list chokes on mixed float/complex items;
         # instead create bare numpy array of appropriate type then add units.
 
-        is_complex = any(isinstance(item.magnitude, complex) for item in items)
+        dtype = (
+            complex 
+            if any(isinstance(item.magnitude, complex) for item in items) 
+            else float
+        )
         units = items[0].units
 
         magnitude = np.fromiter(
             (item.to(units).magnitude for item in items),
-            dtype=(complex if is_complex else float),
+            dtype=dtype,
         )
         return Quantity(magnitude, units)
 
@@ -122,7 +126,7 @@ class AtomTypeShallowDictData(ArrayFromValuesMixin, IsotopeData):
 
           { atom_type_1: value_1, atom_type_2: value_2, ...}
 
-        As this data only contains a single property, the usual *key* will be
+        As these data only contains a single property, the usual *key* will be
         ignored. This is intended for convenient user input of weights at the
         time they are used, and not for development of a longer-term reusable
         dataset.
--- a/euphonic/data/isotopes.py
+++ b/euphonic/data/isotopes.py
@@ -268,19 +268,19 @@ def _get_all_dicts_from_json(
         return dct
 
     if filename := _reference_data_files.get(collection):
-
-        with open(files(euphonic.data) / filename) as fd:
-            file_data = json.load(fd, object_hook=custom_decode)
-
-    elif (filename := Path(collection)).is_file():
-        with open(filename) as fd:
-            file_data = json.load(fd, object_hook=custom_decode)
+        file_path = files(euphonic.data) / filename
     else:
+        file_path = Path(collection)
+
+    if not file_path.is_file():
         msg = format_error(
             f'No data files known for collection "{collection}".',
             fix=f'Available collections: {comma_join(_reference_data_files)}.',
         )
         raise ValueError(msg)
+
+    with file_path.open() as fd:
+        file_data = json.load(fd, object_hook=custom_decode)
 
     if 'physical_property' not in file_data:
         msg = format_error(
--- a/euphonic/data/isotopes.py
+++ b/euphonic/data/isotopes.py
@@ -102,9 +102,15 @@ class AtomTypeDictData(ArrayFromValuesMixin, IsotopeData):
         mass: float,  # noqa: ARG002
         key: str,
     ) -> Quantity:
+        from euphonic.util import comma_join, format_error
+
         if key not in self._data:
-            msg = f'Property "{key}" not found in dict'
+            msg = format_error(
+                f'Property {key!r} not found in dict.',
+                fix=(f'Available keys: {comma_join(self._data.keys())}.'),
+            )
             raise KeyError(msg)
+
         return self._data[key][symbol]
 
 
@@ -159,7 +165,6 @@ class LegacyJsonData(AtomTypeDictData):
     Only the atom_type attribute is used to select data from the structure:
     this is not very robust in case of isotopic substitutions.
     """
-
     def __init__(self, collection: str):
         self._collection = collection
 
@@ -167,8 +172,13 @@ class LegacyJsonData(AtomTypeDictData):
         super().__init__(data)
 
     def get_array(self, structure: Structure, key: str) -> Quantity:
+        from euphonic.util import comma_join, format_error
+
         if key not in self._data:
-            msg = f'Property "{key}" not found in "{self._collection}"'
+            msg = format_error(
+                f'Property {key!r} not found in "{self._collection}"',
+                fix=(f'Available keys: {comma_join(self._data.keys())}.'),
+            )
             raise KeyError(msg)
         return super().get_array(structure, key)
 
--- a/euphonic/data/isotopes.py
+++ b/euphonic/data/isotopes.py
@@ -1,0 +1,304 @@
+from abc import abstractmethod
+from importlib.resources import files
+import json
+from pathlib import Path
+from typing import Protocol
+
+import numpy as np
+from pint import UndefinedUnitError
+
+import euphonic.data
+from euphonic.ureg import Quantity, ureg
+
+
+class Structure(Protocol):
+    """Structure data like euphonic.crystal.Crystal
+
+    Specified as a more general protocol here, so that additional structure
+    types may be added for cluster, slab etc. without breaking the IsotopeData
+    interface (which could plausibly be implemented in external code.)
+    """
+
+    atom_type: np.ndarray
+    atom_mass: Quantity
+
+
+class IsotopeData(Protocol):
+    """Interface for data such as neutron scattering length, cross-section
+
+        Implementations may ignore mass or use it to distinguish between
+        isotopes.
+    """
+
+    def get_value(self, symbol: str, mass: float, key: str) -> Quantity:
+        """Get a property value for specified species"""
+        return self.get_item(symbol, mass)[key]
+
+    @abstractmethod
+    def get_item(self, symbol: str, mass: float) -> dict[str, Quantity]:
+        """Get available data for specified species"""
+
+    @abstractmethod
+    def get_array(self, structure: Structure, key: str) -> Quantity:
+        """Get a Quantity array of property corresponding to structure"""
+
+
+class ArrayFromValuesMixin:
+    """Basic implementation of IsotopeData.get_array using get_value method"""
+
+    def get_array(self, structure: Structure, key: str) -> Quantity:
+        """Get a Quantity array of property corresponding to structure"""
+
+        items = [
+            self.get_value(symbol, mass, key)
+            for symbol, mass in zip(
+                structure.atom_type, structure.atom_mass, strict=True
+            )
+        ]
+        return self._items_to_quantity(items)
+
+    @staticmethod
+    def _items_to_quantity(items: list[Quantity]) -> Quantity:
+        # Quantity.from_list chokes on mixed float/complex items;
+        # instead create bare numpy array of appropriate type then add units.
+
+        is_complex = any(isinstance(item.magnitude, complex) for item in items)
+        units = items[0].units
+
+        magnitude = np.fromiter(
+            (item.to(units).magnitude for item in items),
+            dtype=(complex if is_complex else float),
+        )
+        return Quantity(magnitude, units)
+
+
+class AtomTypeDictData(ArrayFromValuesMixin, IsotopeData):
+    def __init__(self, data: dict[str, dict[str, Quantity]]) -> None:
+        """Property data from simple dict
+
+        Dict must have structure::
+
+          { key: {atom_type_1: value_1, atom_type_2: value_2, ...}, ...}
+
+        Where *key* is a string key passed to corresponding argument of
+        .get_array() method (e.g. "coherent_cross_section"),
+        and atom types correspond to Structure.atom_type array (e.g. "Si").
+
+        Only the atom_type attribute is used to select data from the structure:
+        this is not very robust in case of isotopic substitutions.
+        """
+        self._data = data
+
+    def get_item(
+        self,
+        symbol: str,
+        mass: float,  # noqa: ARG002
+    ) -> dict[str, Quantity]:
+        return {key: value[symbol] for key, value in self._data.items()}
+
+    def get_value(
+        self,
+        symbol: str,
+        mass: float,  # noqa: ARG002
+        key: str,
+    ) -> Quantity:
+        if key not in self._data:
+            msg = f'Property "{key}" not found in dict'
+            raise KeyError(msg)
+        return self._data[key][symbol]
+
+
+class AtomTypeShallowDictData(ArrayFromValuesMixin, IsotopeData):
+    def __init__(self, data: dict[str, Quantity]) -> None:
+        """Property data from very simple dict
+
+        Dict must have structure::
+
+          { atom_type_1: value_1, atom_type_2: value_2, ...}
+
+        As this data only contains a single property, the usual *key* will be
+        ignored. This is intended for convenient user input of weights at the
+        time they are used, and not for development of a longer-term reusable
+        dataset.
+
+        Only the atom_type attribute is used to select data from the structure:
+        this is not very robust in case of isotopic substitutions.
+        """
+        self._data = data
+
+    def get_item(
+        self,
+        symbol: str,
+        mass: float,  # noqa: ARG002
+    ) -> dict[str, Quantity]:
+        """Get all properties for specified species
+
+        In this case, mass is ignored and sole property has empty-string key ''
+        """
+        return {'': self._data[symbol]}
+
+    def get_value(
+        self,
+        symbol: str,
+        mass: float,  # noqa: ARG002
+        key: str,  # noqa: ARG002
+    ) -> Quantity:
+        """Get property value for specified species
+
+        In this case, mass and key are ignored: only one value is available
+        for each symbol.
+        """
+        return self._data[symbol]
+
+
+class LegacyJsonData(AtomTypeDictData):
+    """Property data from JSON file
+
+    This provides an interface for the legacy reference data.
+
+    Only the atom_type attribute is used to select data from the structure:
+    this is not very robust in case of isotopic substitutions.
+    """
+
+    def __init__(self, collection: str):
+        self._collection = collection
+
+        data = _get_all_dicts_from_json(collection)
+        super().__init__(data)
+
+    def get_array(self, structure: Structure, key: str) -> Quantity:
+        if key not in self._data:
+            msg = f'Property "{key}" not found in "{self._collection}"'
+            raise KeyError(msg)
+        return super().get_array(structure, key)
+
+
+def get_reference_data(
+    collection: str = 'Sears1992',
+    physical_property: str = 'coherent_scattering_length',
+) -> dict[str, Quantity]:
+    """
+    Get physical data as a dict of (possibly-complex) floats from reference
+    data.
+
+    Each "collection" refers to a JSON file which may contain any set of
+    properties, indexed by physical_property.
+
+    Properties are stored in JSON files, encoding a single dictionary with the
+    structure::
+
+      {"metadata1": "metadata1 text", "metadata2": ...,
+       "physical_properties": {"property1": {"__units__": "unit_str",
+                                             "H": H_property1_value,
+                                             "He": He_property1_value,
+                                             "Li": {"__complex__": true,
+                                                    "real": Li_property1_real,
+                                                    "imag": Li_property1_imag},
+                                             "Nh": None,
+                                             ...},
+                               "property2": ...}}
+
+    Parameters
+    ----------
+    collection
+        Identifier of data file; this may be an inbuilt data set ("Sears1992"
+        or "BlueBook") or a path to a JSON file (e.g. "./my_custom_data.json").
+
+    physical_property
+        The name of the property for which data should be extracted. This must
+        match an entry of "physical_properties" in the data file.
+
+    Returns
+    -------
+    dict[str, Quantity]
+        Requested data as a dict with string keys and (possibly-complex)
+        float Quantity values. String or None items of the original data file
+        will be omitted.
+
+    """
+
+    # Avoid import loop; the problem will go away when this function is removed
+    from euphonic.util import _deprecation_warn, comma_join, format_error
+
+    _deprecation_warn(
+        'get_reference_data', 'IsotopeData.get_array', stacklevel=3
+    )
+
+    data = _get_all_dicts_from_json(collection)
+
+    if physical_property not in data:
+        msg = format_error(
+            (f'No such collection {collection!r} '
+             f'with property {physical_property!r}.'),
+            fix=('Available properties for this collection:'
+                 f'{comma_join(data)}.'),
+        )
+        raise ValueError(msg)
+
+    return data[physical_property]
+
+
+def _get_all_dicts_from_json(
+    collection: str = 'Sears1992',
+) -> dict[str, dict[str, Quantity]]:
+    from euphonic.util import comma_join, format_error  # Avoid import loop
+
+    _reference_data_files = {
+        'Sears1992': 'sears-1992.json',
+        'BlueBook': 'bluebook.json',
+    }
+
+    def custom_decode(dct):
+        if '__complex__' in dct:
+            return complex(dct['real'], dct['imag'])
+        return dct
+
+    if filename := _reference_data_files.get(collection):
+
+        with open(files(euphonic.data) / filename) as fd:
+            file_data = json.load(fd, object_hook=custom_decode)
+
+    elif (filename := Path(collection)).is_file():
+        with open(filename) as fd:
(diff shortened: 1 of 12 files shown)


ajjackson:
@oerc0122 I think this is in a good state to have a look at the design! Still not quite sure about all the names/locations of things. And need to decide the release/deprecation strategy.

Does it even make sense to merge this before a new/improved dataset is added? Maybe if it's for v2.

I really do need to get this into a release quickly though, so wouldn't like it to be blocked by other v2 things. But the remaining v2 things don't look too bad.

euphonic/data/isotopes.py
```diff
@@ -0,0 +1,312 @@
...
+    types may be added for cluster, slab etc. without breaking the IsotopeData
+    interface (which could plausibly be implemented in external code.)
+    """
+
+    atom_type: np.ndarray
+    atom_mass: Quantity
+
+
+@runtime_checkable
+class IsotopeData(Protocol):
```
oerc0122, line 26: For the record, according to the PEP, I think `Protocol` IS an `ABC` (c.f. https://peps.python.org/pep-0544/#defining-a-protocol ; https://peps.python.org/pep-0544/#explicitly-declaring-implementation) so these two may not be necessary.

euphonic/data/isotopes.py
```diff
@@ -0,0 +1,312 @@
...
+    """Interface for data such as neutron scattering length, cross-section"""
+
+    def get_value(
+        self, symbol: str, mass: float, key: str
+    ) -> Quantity: ...
+
+    def get_array(self, structure: Structure, key: str) -> Quantity: ...
+
+    def get_item(self, symbol: str, mass: float) -> dict[str, Quantity]:
```
oerc0122, line 38: I was potentially thinking you have a method
```python
    def __getitem__(self, key: tuple[str, float]) -> dict[str, Quantity]: ...
```
This could be implemented as a wrapper function? That way you might be able to support the Mapping Protocol.

euphonic/data/isotopes.py
ajjackson: Is it ok to put implementations on a Protocol, then?

oerc0122: Yes, though any protocol member **WILL** be required. So you may still need the `ABC` (which can simply inherit `IsotopeData` instead of `ABC`) with the implementation of the private method.

euphonic/data/isotopes.py
ajjackson: Hmm, that is a bit of an issue with this method name! The trouble is that it isn't always very mapping-like:

- there are two parameters
- a variety of values might lead to the same data for the smarter implementations to be implemented later.

So if we did a numpy-like tuple-as-index we would get cases like

```
isotope_data['H:2', 2.001] is isotope_data['D', 2.0]
```

which is quite surprising!

It is quite neat though, maybe not entirely out of the question 🤔 But I think I'd prefer a name/syntax that makes it clear some searching/analysis might be going on.

oerc0122: I'm not sure that's surprising as such. 

I don't think it's unreasonable to have tuples as keys e.g. in castep_ouputs `md_geom_frame["ions"]["H", 3]` which is a valid structure and those are pure `dict`s.

Likewise there's no rule saying that 2 names can't refer to the same object (in fact this is fairly native for how `dict`s work.

ajjackson: Yeah, tuples as keys can be quite nice actually. (And totally valid to do with regular `dict`.) But floats in keys looks quite alien.

Perhaps a bigger issue with making this a `Mapping` is the associated set of methods: Mapping should implement `__getitem__`, `__iter__` and `__len__`. Then `__contains__`, `keys`, `items`, `values`, `get`, `__eq__` and `__ne__` will be provided. I don't think all of these make sense for this kind of dataset.
- If we _do_ have `__iter__` I'd expect it to iterate over data rows, not keys
- It would be weird to access data using keys that _aren't_ part of `keys`
- `__len__` doesn't seem a good fit for data that might include "aliases" like D for deuterium.

Perhaps it should be an `Iterable[dict]` of data rows with `select()` and `select_one()` methods as well as `get_array()`.

euphonic/data/isotopes.py
```diff
@@ -0,0 +1,312 @@
...
+        'get_reference_data', 'IsotopeData.get_array', stacklevel=3
+    )
+
+    data = _get_all_dicts_from_json(collection)
+
+    if physical_property not in data:
+        msg = dedent_and_fill(
+            f"""
+            No such collection "{collection}" with property
+            "{physical_property}". Available properties for this collection:
```
oerc0122: Possibly?
```suggestion
            No such collection {collection!r} with property
            {physical_property!r}. Available properties for this collection:
```

euphonic/data/isotopes.py
```diff
@@ -0,0 +1,312 @@
...
+        'BlueBook': 'bluebook.json',
+    }
+
+    def custom_decode(dct):
+        if '__complex__' in dct:
+            return complex(dct['real'], dct['imag'])
+        return dct
+
+    if collection in _reference_data_files:
+        filename = _reference_data_files[collection]
```
oerc0122:
```suggestion
    if filename := _reference_data_files.get(collection):
        
```

euphonic/data/isotopes.py
oerc0122, line 1: None of the errors in this file currently follow the "standard" format, or is that 2.0.0?

euphonic/qpoint_phonon_modes.py
```diff
@@ -543,52 +544,19 @@ def calculate_pdos(
...
-        if cross_sections_data is not None:
-            cs = [cross_sections_data[0][x] for x in self.crystal.atom_type]
-            if len(cross_sections_data) == 2:
-                cs2 = [cross_sections_data[1][x]
-                        for x in self.crystal.atom_type]
-                cs = [sum(x) for x in zip(cs, cs2, strict=True)]
-            # Account for cross sections in different, or invalid, units
+        cs = _get_cross_sections(self.crystal, weighting, cross_sections)
+
+        if cs is not None:
```
oerc0122, line 557:
```suggestion
        if (cs := _get_cross_sections(self.crystal, weighting, cross_sections))
 is not None:
```

euphonic/qpoint_phonon_modes.py
```diff
@@ -732,3 +700,74 @@ def from_phonopy(cls: type[T], path: Path | str = '.',
...
+            msg = (
+                f'Unexpected type for scattering_lengths, should be str '
+                f'or dict, got {type(scattering_lengths)}'
+            )
+            raise TypeError(msg)
+
+def _validate_weighting(weighting: str | None):
+    weighting_opts = [
+        None, 'coherent', 'incoherent', 'coherent-plus-incoherent',
+    ]
```
oerc0122:
```suggestion
    weighting_opts = {
        None, 'coherent', 'incoherent', 'coherent-plus-incoherent',
    }
```

euphonic/data/isotopes.py
ajjackson: Same number of characters and maybe less readable. But possibly a good habit? It handles daft cases with quotes or newlines inside the string more robustly.

euphonic/data/isotopes.py
ajjackson: It's 2.0

euphonic/data/isotopes.py
oerc0122: I agree that maybe the `Mapping` ABC doesn't make so much sense to this (although it is really a key-value mapping just with key normalisation). I don't think `float`s in keys looks bad to me knowing that this isn't really a hash-map.

euphonic/data/isotopes.py
oerc0122: Dang, I think we want to create an issue for this to be merged into 1.x.

euphonic/qpoint_phonon_modes.py
ajjackson: Not sure about this line-breaking situation.

Perhaps the `is not None` can be removed anyway? I don't suppose it matters what units are used if the cross-section array has size zero.

euphonic/qpoint_phonon_modes.py
ajjackson: This is a list so that the error message uses a consistent order. (A tuple would also be ok.)

oerc0122: I would personally prefer `tuple` then (since it is not meant to be changed)

euphonic/data/isotopes.py
ajjackson: Either that or this MR targets 2.0 and we rush through remaining 2.0 requirements 😅

euphonic/qpoint_phonon_modes.py
ajjackson:
```suggestion
    weighting_opts = (
        None, 'coherent', 'incoherent', 'coherent-plus-incoherent',
    )
```

euphonic/data/isotopes.py
ajjackson: I think the [] access is likely to lead to confusion given that it is _sometimes_ a hash-map.

Can you think of any mainstream examples of libraries that use this syntax to access something that isn't basically an int index or a hash-map key?

ajjackson: Any thoughts on `select(**kwargs)`/`select_one(**kwargs)` methods? This would be a more general treatment allowing to query on other keys, that might allow nice things like selecting on mass number instead of mass. But... it also makes it much harder to indicate that certain keys are _required_.

oerc0122: But in the case of a hash-map key the mass will be ignored, truncated or built into the string to make it valid and that will be embedded in the function regardless of whether it's `.get` or `[]`. 

While not strictly float index, I have seen uses of float slices and datetime slices to extract/filter data. 

Python itself doesn't care, it's just a function and personally I see no difference between:
```python
x.get("Hi", 3.1)
x["Hi", 3.1]
```
Besides the suggestion that `[]` is the expected mechanism for returning me that value and raising a keyerror otherwise.

oerc0122: > Any thoughts on `select(**kwargs)`/`select_one(**kwargs)` methods? This would be a more general treatment allowing to query on other keys, that might allow nice things like selecting on mass number instead of mass. But... it also makes it much harder to indicate that certain keys are _required_.

I think regardless of whether the `ABC`/`protocol` implements `**kwargs` (remember with `@runtime_checkable`, parameters aren't checked), then children can implement explicit `*, x=...` without violating the LSP **provided** they have defaults.

ajjackson: > Python itself doesn't care

Python is a *very* flexible language, we need conventions and idioms to manage this!

euphonic/qpoint_phonon_modes.py
ajjackson: I think I'd rather keep this without walrus as `cs` also gets referenced further down in the code. This way it is easier to scan up and see where it was defined.

euphonic/data/isotopes.py
ajjackson: Have refactored: now the Protocol is used as ABC and the `get_array`-implementing class is a mix-in.

euphonic/data/isotopes.py
ajjackson: On further discussion:
- general systems like `select(**kwargs)` are neat, but YAGNI and they go on some parent class if we do need them later.
- the method names are _ok_ and usage pattern is not quite clear enough to commit to interesting magic access methods. If we change our mind things can be aliased at runtime with a wrapper function/class and the specification updated for Euphonic v3.

So for now we stick with get_value/get_item/get_array

euphonic/data/isotopes.py
ajjackson: Looks like that is the plan! I'd better rebase and sort out the error messages.

ajjackson:
I didn't fancy rebasing all those commits _again_ after tidying up release-2.0.0, so did a big squash and cherry-pick. It'll all be squashed in the end anyway...

euphonic/data/isotopes.py
ajjackson: Ok we are rebased on release-2 and the error messages should all be modernised

ajjackson:
From the HTML artifact I can see that test coverage for the new module is 100%. But for some reason this is entirely missed by the orgoro summary action?

It even notes the new file in log, then disregards its coverage.

image

oerc0122:
> From the HTML artifact I can see that test coverage for the new module is 100%. But for some reason this is entirely missed by the orgoro summary action?

Hmmm...

> Modified Files
> No covered modified files...

ajjackson:
And it's in the `coverage-combined.xml`:

```
        <class line-rate="1" branch-rate="1" complexity="0" name="isotopes.py" filename="/home/runner/work/Euphonic/Euphonic/.tox/py314/lib/python3.14/site-packages/euphonic/data/isotopes.py">
          <methods />
          <lines>
            <line number="1" hits="1" branch="False" />
            <line number="2" hits="1" branch="False" />
            <line number="3" hits="1" branch="False" />
            <line number="4" hits="1" branch="False" />
            <line number="5" hits="1" branch="False" />
...
```

oerc0122:
> /home/runner/work/Euphonic/Euphonic/.tox/py314/lib/python3.14/site-packages/euphonic/data/isotopes.py

Could the fact that it's:
`/home/runner/work/Euphonic/Euphonic/.tox/py314/lib/python3.14/site-packages/euphonic/data/isotopes.py`
be the problem? I can't imagine how it could be if it's detecting other coverage.

ajjackson:
Could be, this issue suggests it can be fixed with a tweak to coverage path-handling https://github.com/orgoro/coverage/issues/248

ajjackson:
Well that achieved nothing :(

Anyway, this would be a good time for re-review :)

euphonic/data/isotopes.py
```diff
@@ -0,0 +1,314 @@
...
+        """Get a property value for specified species"""
+        return self.get_item(symbol, mass)[key]
+
+    @abstractmethod
+    def get_item(self, symbol: str, mass: float) -> dict[str, Quantity]:
+        """Get available data for specified species"""
+
+    @abstractmethod
+    def get_array(self, structure: Structure, key: str) -> Quantity:
+        """Get a Quantity array of property corresponding to structure"""
```
oerc0122, line 43: I think you could probably implement the rough version here based on `get_item`, it can always be overridden, but this means implementers who inherit this only need to implement a `get_item`.

euphonic/data/isotopes.py
```diff
@@ -0,0 +1,314 @@
...
+    def _items_to_quantity(items: list[Quantity]) -> Quantity:
+        # Quantity.from_list chokes on mixed float/complex items;
+        # instead create bare numpy array of appropriate type then add units.
+
+        is_complex = any(isinstance(item.magnitude, complex) for item in items)
+        units = items[0].units
+
+        magnitude = np.fromiter(
+            (item.to(units).magnitude for item in items),
+            dtype=(complex if is_complex else float),
```
oerc0122: Would this be better phrased as:

```suggestion
        dtype = (
            complex 
            if any(isinstance(item.magnitude, complex) for item in items) 
            else float
        )
        units = items[0].units

        magnitude = np.fromiter(
            (item.to(units).magnitude for item in items),
            dtype=dtype,
```

euphonic/data/isotopes.py
```diff
@@ -0,0 +1,314 @@
...
+
+class AtomTypeShallowDictData(ArrayFromValuesMixin, IsotopeData):
+    def __init__(self, data: dict[str, Quantity]) -> None:
+        """Property data from very simple dict
+
+        Dict must have structure::
+
+          { atom_type_1: value_1, atom_type_2: value_2, ...}
+
+        As this data only contains a single property, the usual *key* will be
```
oerc0122:
```suggestion
        As these data only contains a single property, the usual *key* will be
```

euphonic/data/isotopes.py
```diff
@@ -0,0 +1,314 @@
...
+                fix=(f'Available keys: {comma_join(self._data.keys())}.'),
+            )
+            raise KeyError(msg)
+
+        return self._data[key][symbol]
+
+
+class AtomTypeShallowDictData(ArrayFromValuesMixin, IsotopeData):
+    def __init__(self, data: dict[str, Quantity]) -> None:
+        """Property data from very simple dict
```
oerc0122, line 123:
```suggestion
        """Property data from very simple dict.
```

euphonic/data/isotopes.py
```diff
@@ -0,0 +1,314 @@
...
+        magnitude = np.fromiter(
+            (item.to(units).magnitude for item in items),
+            dtype=(complex if is_complex else float),
+        )
+        return Quantity(magnitude, units)
+
+
+class AtomTypeDictData(ArrayFromValuesMixin, IsotopeData):
+    def __init__(self, data: dict[str, dict[str, Quantity]]) -> None:
+        """Property data from simple dict
```
oerc0122, line 81:
```suggestion
        """Property data from simple dict.
```

euphonic/data/isotopes.py
```diff
@@ -0,0 +1,314 @@
...
+
+    _reference_data_files = {
+        'Sears1992': 'sears-1992.json',
+        'BlueBook': 'bluebook.json',
+    }
+
+    def custom_decode(dct):
+        if '__complex__' in dct:
+            return complex(dct['real'], dct['imag'])
+        return dct
```
oerc0122, line 268: Does this want to be implemented as a `JSONDecoder` in `util` or even this in `util`?

euphonic/data/isotopes.py
```diff
@@ -0,0 +1,314 @@
...
+
+    if filename := _reference_data_files.get(collection):
+
+        with open(files(euphonic.data) / filename) as fd:
+            file_data = json.load(fd, object_hook=custom_decode)
+
+    elif (filename := Path(collection)).is_file():
+        with open(filename) as fd:
+            file_data = json.load(fd, object_hook=custom_decode)
```
oerc0122: I feel this setting `to_open = files(euphonic.data) / filename)` / `to_open = filename` in each branch would be slightly DRYer

euphonic/data/isotopes.py
```diff
@@ -0,0 +1,314 @@
...
+    result: dict[str[dict[str, Quantity]]] = {}
+    for physical_property, data in file_data['physical_property'].items():
+        unit_str = data.get('__units__')
+
+        no_units_msg = format_error(
+            f'No units in file ({filename}).',
+            fix='Ensure file specifies dimensions with "__units__" metadata.',
+        )
+        if unit_str is None:
+            raise ValueError(no_units_msg)
```
oerc0122, line 301:
```suggestion
        if (unit_str := data.get('__units__')) is None:
            no_units_msg = format_error(
                f'No units in file ({filename}).',
                fix='Ensure file specifies dimensions with "__units__" metadata.',
            )
            raise ValueError(no_units_msg)
```
Alternatively you may get away with `not`

euphonic/qpoint_phonon_modes.py
```diff
@@ -184,7 +185,8 @@ def reorder_frequencies(self,
 
     def calculate_structure_factor(
         self,
-        scattering_lengths: str | dict[str, Quantity] = 'Sears1992',
+        scattering_lengths: (
+            IsotopeData | str | dict[str, Quantity]) = 'Sears1992',
```
oerc0122: Does this want to become a `TypeAlias`
```python
IsotopeDataset: TypeAlias = IsotopeData | str | dict[str, Quantity]
```
(3.12+)
```python
type IsotopeDataset = IsotopeData | str | dict[str, Quantity]
```

euphonic/qpoint_phonon_modes.py
```diff
@@ -743,3 +705,79 @@ def from_phonopy(cls, path: Path | str = '.',
...
+                fix='scattering_lengths should be str or dict.',
+            )
+            raise TypeError(msg)
+
+def _validate_weighting(weighting: str | None):
+    weighting_opts = (
+        None, 'coherent', 'incoherent', 'coherent-plus-incoherent',
+    )
+    if weighting not in weighting_opts:
+        msg = (
```
oerc0122: Needs modern format

tests_and_analysis/test/euphonic_test/test_isotope_data.py
```diff
@@ -0,0 +1,144 @@
...
+    IsotopeData,
+    LegacyJsonData,
+    Structure,
+)
+from euphonic.ureg import Quantity
+
+
+def _compare_quantity(a: Quantity, b: Quantity) -> None:
+    assert a.units == b.units
+    assert_allclose(a.magnitude, b.magnitude)
```
oerc0122, line 19: I wonder if this is worth monkeypatching the `__eq__` method of Quantity

oerc0122 (commented):
Looking good so far!

euphonic/data/isotopes.py
ajjackson: The problem with putting that implementation on the Protocol is that it drags in a private method as well. The Mixin seems the appropriate solution to that problem while keeping things nicely scoped and structured.

euphonic/data/isotopes.py
ajjackson: Sure!

euphonic/data/isotopes.py
ajjackson: I know some linters/style-guides do this, but so far it is not generally the style in Euphonic. Personally, I dislike incomplete sentences ending with a period. This makes it doubly enraging when they would take one over the linewidth-limit and some linter is demanding it.

If we're going to end fragments with periods it should happen everywhere and carry the intent that fragments are gradually reworked into sentences.

I would like to increase the linewidth limit to 88 at some point, which might help with this!

euphonic/data/isotopes.py
ajjackson: This exceeds the linewidth limit

euphonic/data/isotopes.py
ajjackson: In general I find functions are easy to summarise as a proper sentence but with classes it gets rather contrived.

euphonic/data/isotopes.py
ajjackson: At the moment this format is only used for these files so there is no duplication. It's likely to be deleted before it's re-used; given that there are only a few lines of code I think it makes sense to keep it local.

euphonic/data/isotopes.py
ajjackson: Makes sense. How about
```python
    if filename := _reference_data_files.get(collection):
        file_path = files(euphonic.data) / filename
    else:
        file_path = Path(collection)

    if not file_path.is_file():
        msg = format_error(
            f'No data files known for collection "{collection}".',
            fix=f'Available collections: {comma_join(_reference_data_files)}.',
        )
        raise ValueError(msg)

    with file_path.open() as fd:
        file_data = json.load(fd, object_hook=custom_decode)

```

euphonic/qpoint_phonon_modes.py
ajjackson: Would make sense. This group is also used in powder.py but I suppose that can import it from qpoint_phonon_modes.py

euphonic/data/isotopes.py
oerc0122: Looks nice and clean with no repetition.

tests_and_analysis/test/euphonic_test/test_isotope_data.py
ajjackson: We generally use `assert_allclose` rather than monkey-patch `np.ndarray.__eq__`.

Another approach is [`pytest.approx`](https://docs.pytest.org/en/stable/reference/reference.html#pytest-approx): a wrapper that exposes a modified comparison operation.

euphonic/data/isotopes.py
oerc0122: It's more that in the docs it looks weird, vs. in the code it looks weird. I'm not sure which I prefer, but I prefer the full stop to indicate "There are no other lines here". Very much a personal thing.

ajjackson: Rather than make things inconsistent now, let's open an issue to revisit this at the same time line-length is expanded. (Between v2 and v3?)

ajjackson: #506

ajjackson:
I would like to also included a superior isotope-dataset for euphonic-2.0.0; much as I try to explain to people that SemVer means major version numbers don't imply new features, it would be nice to have _some_ useful feature to announce with it.

It seems better to do that in a separate PR, after this one establishes a reasonable API. But then if API problems are encountered then there is still a chance to rework it before 2.0.

euphonic/data/isotopes.py
oerc0122: It doesn't have to drag in a private method, it could be an top-level function or an internal function. Still not sure that's  good design though

ajjackson: Right, any of those three options would work.

I think having a lot of implementation on the Protocol is distracting and makes it more of a base class. Hopefully the current version with the methods abstract or one-liner signals that you're _supposed_ to think about which implementation to use and not assume the default is good.

That could also be achieved with module-level private functions and a comment in the Protocol saying something like `# fallback implementation`. I slightly prefer the Mixin though, there's something elegant about composing things out of building blocks.

oerc0122 (approved):
Looks good, let's get it in!

PR #495, opened 2026-05-07
Fix some "easy wins" in typing

oerc0122:
- Replace `Self` `TypeVars` with actual `typing_extensions.Self`
- Add `None` as union for `Optional`s
- Abort when None would come from `spglib`
- Declare types in cases where function could return `tuple`, but won't/can't for argument reasons (until overloads are well defined) 
- Add `Dictable` protocol for those types which support `to/from_dict`


euphonic/spectra/base.py
```diff
@@ -11,14 +11,14 @@
...
     Literal,
-    TypeVar,
     overload,
 )
 import warnings
 
 import numpy as np
 from pint import DimensionalityError, Quantity
 from scipy.ndimage import correlate1d, gaussian_filter
+from typing_extensions import Self
```
ajjackson, line 21: `typing_extensions` is not currently in the Euphonic requirements. It's slightly surprising this passes tests, presumably another dependency is dragging it in? Still we can't assume a dependency will keep relying on it, and this should be added to pyproject.toml. That would also typically be associated with a full minor version bump, rather than bugfix release.

euphonic/io.py
```diff
@@ -11,6 +11,12 @@
 
 T = TypeVar('T')
 
+class Dictable(Protocol):
+    def to_dict(self) -> dict: ...
+
+    @classmethod
+    def from_dict(cls: type[T], dictionary: dict) -> T: ...
```
ajjackson: Is it a problem that the parameter is usually called `d` in current implementations?

Maybe that would be a good v2 improvement.

euphonic/spectra/base.py
oerc0122: I agree **cough**:
```sh
jacob@nyx [22:36:01] ~/Euphonic/euphonic %  [master]
> grep typing_extensions **/*.py
spectra/collections.py:from typing_extensions import Self
```

ajjackson: Haha, good spot! I guess that makes it a bugfix then

oerc0122: I'll stick the dep formally in

pyproject.toml
```diff
@@ -42,6 +42,7 @@ dependencies = [
     "pint>=0.22",
     "threadpoolctl>=3.0.0",
     "toolz>=0.12.1",
+    "typing_extensions>=4.0.0,<=4.15.0"
```
ajjackson: Why 4.15 rather than `typing_extensions~=4.0`?

ajjackson:
Coverage action still not working from remote branch 😞

pyproject.toml
oerc0122: Because I still don't really understand python version pinning? Will fix.

ajjackson: It's not like the existing ones show a consistent practice either!

It's a bit annoying that `pint` is still on zero-based version numbers and can have breaking changes; we don't want to hard-pin and create conflicts so have to keep an eye on this.

`toolz` now has a 1.0 release so I think we can add a 2.0 upper bound now.

ajjackson:
I think all this needs now is a note in the CHANGELOG that `typing_extensions` was officially added to dependencies (but already implicitly included via Pint's dependencies.)

ajjackson (approved):
I think this is ready to go! Is there a reason it was still marked draft?

oerc0122:
> I think this is ready to go! Is there a reason it was still marked draft?

It was originally a bit experimental and I didn't want to harass you for review.

ajjackson:
Thank you! I'll sort out some kind of merge/rebase to release-2.0.0 once the other PR is merged

PR #497, opened 2026-05-08
Migrate to dependency groups

oerc0122:
- Move dev deps to dependency-groups (install with `--group`)
- Update tests respectively
- Unifies all dependencies into `pyproject.toml` and allows us to build the min reqs from another.  

Failure due to coverage.


build_utils/pypi_release_tox.ini
```diff
@@ -30,7 +31,10 @@ commands_pre = {[testenv]inst_str} --only-binary 'euphonic'
 
 [testenv:py310-min]
 platform = (linux)|(win32)
+depedency_groups =
```
ajjackson:
```suggestion
dependency_groups =
```

build_utils/pypi_release_tox.ini
```diff
@@ -30,7 +31,10 @@ commands_pre = {[testenv]inst_str} --only-binary 'euphonic'
...
+    {[testenv]dependency_groups}
+    min_reqs
 deps =
     numpy==1.24
     setuptools==60.*
```
ajjackson: I expect this `setuptools` can also be dropped since we moved to meson-python

CHANGELOG.rst
```diff
@@ -3,6 +3,8 @@
 
 - Requirements
 
+  - Migrate dev dependencies to dependency groups.
```
ajjackson:
```suggestion
  - Migrate dev dependencies from "optional dependencies" to `dependency groups <https://packaging.python.org/en/latest/specifications/dependency-groups/>`__. 
    This decouples them from the actual package and changes their installation process
    from e.g. `pip install .[test]` to `pip install --group test`.
```

pyproject.toml
```diff
@@ -52,13 +52,43 @@ Repository = "https://github.com/pace-neutrons/Euphonic.git"
...
+    "pint==0.22",
+    "matplotlib==3.8",
+    "h5py==3.6; sys_platform != 'darwin'",
+    "h5py==3.7; sys_platform == 'darwin'",
+    "PyYAML==6.0",
+    "threadpoolctl==3.0.0",
+    "toolz==0.12.1"
+]
+test = ["mock", "pytest>=9.0.3,<10.0", "pytest-cov", "pytest-mock", "pytest-lazy-fixtures", "pytest-xdist", "pytest-xvfb", "python-slugify"]
+ci = ["tox==4.30.3"]
```
ajjackson, line 70: I think this means *tests_and_analysis/ci_requirements* can also be deleted.
There is also a "tox_requirements" file which, as far as I can tell, is already no longer used anywhere.

pyproject.toml
```diff
@@ -52,13 +52,43 @@ Repository = "https://github.com/pace-neutrons/Euphonic.git"
...
+    "sphinx~=5.3.0",
+    "sphinx-autodoc-typehints==1.19.5",
+    "sphinx-argparse==0.3.2",
+    "sphinx-rtd-theme==1.1.1",
+    "wheel==0.46.2",
+]
+dev = [
+    {include-group = "test"},
+    {include-group = "ci"},
+    {include-group = "docs"},
```
ajjackson: Do we actually want to have `docs` in dev when `docs` includes so many hard pins? (And sphinx is a slightly chunky requirement.) In the `uv` workflow (if/when other issues with that are resolved...) I'd think I'd rather `uv run --group docs make html` just for docs builds.

(I get that there is a case for hard pins in `dev`, but even if we go along with that it's weird for `docs` to be the driver. Lockfile commits are the standard way to do that.)

tox.ini
ajjackson, line 115: I guess this means we no longer need separate tox environment specs for minrequirements on Mac and Linux; we can just have a global minrequirements env and the dependency group magic should take care of things.

ajjackson (commented):
Looking like a nice modernisation and cleanup! A few comments so far.

build_utils/pypi_release_tox.ini
oerc0122: Good catch! I keep doing this, it's a really bad habit.

build_utils/pypi_release_tox.ini
oerc0122:
```suggestion
```

pyproject.toml
oerc0122:
```suggestion
```

tox.ini
oerc0122: Do you think that refactor should be part of this?

pyproject.toml
oerc0122: Done.

tox.ini
ajjackson: It can be a follow-on (very straightforward) PR if you like. Would let the review focus more on scrutinising the test outputs.

oerc0122: Opened #504

ajjackson (approved):
Looks great 🚀

PR #501, opened 2026-05-10
Fix some typing by using generic spectrum mixin

oerc0122:
- Make `SpectrumCollectionMixin` generic w.r.t. `Spectrum` type.
- Fix some erroneous `cls: Self` in `@classmethods`

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -53,6 +53,10 @@
     **breaking change** as somebody's code could depend on this
     IndexError. At this stage it seems an acceptable risk.
 
+- Improvements
+
+  - Fix several typing issues with SpectrumNDCollection.
+
 - Other changes
 
   - Error messages have been overhauled and now follow a consistent format::
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -55,7 +55,7 @@
 
 - Improvements
 
-  - Fix several typing issues with SpectrumNDCollection.
+  - Fix several typing issues with Spectrum collections.
 
 - Other changes
 
--- a/euphonic/spectra/collections.py
+++ b/euphonic/spectra/collections.py
@@ -12,6 +12,7 @@ from operator import contains
 from pathlib import Path
 from typing import (
     Any,
+    Generic,
     Literal,
     TypeVar,
     overload,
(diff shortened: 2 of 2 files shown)


euphonic/spectra/collections.py
```diff
@@ -276,8 +278,8 @@ def copy(self) -> Self:
 
     def __add__(self, other: Self) -> Self:
         """
-        Appends the y_data of 2 Spectrum1DCollection objects,
-        creating a single Spectrum1DCollection that contains
+        Appends the y_data of 2 SpectrumNDCollection objects,
+        creating a single SpectrumNDCollection that contains
```
ajjackson: As this gives the strong impression that a "SpectrumNDCollection" class exists, maybe we should use something like "spectrum collection" when we want to ambiguously refer to the current class.

It's still a bit unsatisfying, I wonder if it's possible to magically interpolate the class type in there somehow. I think it could be done with `__init_subclass__` modifying the `__doc__` attributes of methods. Definitely more complicated than using vague text!

oerc0122:
```suggestion
        Appends the y_data of 2 spectrum collection objects,
        creating a single spectrum collection that contains
```

euphonic/spectra/collections.py
```diff
@@ -348,7 +350,7 @@ def select(
         Returns
         -------
         selected_spectra
-           A Spectrum1DCollection containing the selected spectra
+           A SpectrumNDCollection containing the selected spectra
```
oerc0122:
```suggestion
           A spectrum collection containing the selected spectra
```

euphonic/spectra/collections.py
```diff
@@ -463,7 +465,7 @@ def group_by(self, *line_data_keys: str) -> Self:
         Returns
         -------
         grouped_spectrum
-            A new Spectrum1DCollection with one line for each group. Any
+            A new SpectrumNDCollection with one line for each group. Any
```
oerc0122:
```suggestion
            A new spectrum collection with one line for each group. Any
```

CHANGELOG.rst
```diff
@@ -53,6 +53,10 @@
     **breaking change** as somebody's code could depend on this
     IndexError. At this stage it seems an acceptable risk.
 
+- Improvements
+
+  - Fix several typing issues with SpectrumNDCollection.
```
oerc0122:
```suggestion
  - Fix several typing issues with Spectrum collections.
```

ajjackson (approved):
Great! This should help type-checkers understand what is going on, without much risk of API change/breakage.

PR #502, opened 2026-05-10
Allow empty collections

oerc0122:
Fixes #465 

Permit empty spectrum collections.

N.B. needs more testing.

### DIFF
--- a/euphonic/spectra/collections.py
+++ b/euphonic/spectra/collections.py
@@ -401,8 +401,11 @@ class SpectrumCollectionMixin(ABC, Generic[Spec]):
 
         # Combine key-value pairs common to *all* metadata lines into new dict
         common_metadata = dict(
-            reduce(set.intersection,
-                   (set(metadata.items()) for metadata in all_metadata)))
+            reduce(
+                set.intersection,
+                (set(metadata.items()) for metadata in all_metadata),
+                set(),
+            ))
 
         # Put all other per-spectrum metadata in line_data
         is_common = partial(contains, common_metadata)
@@ -603,7 +606,7 @@ class Spectrum1DCollection(SpectrumCollectionMixin[Spectrum1D],
             X_TICK_CHECK._replace(value=x_tick_labels),
             METADATA_CHECK._replace(value=metadata),
         )
-        ny = len(y_data[0])
+        ny = y_data.shape[1]
         _check_constructor_inputs(
             InputCheck(x_data, (Quantity,), {(ny,), (ny+1,)}, 'x_data'))
 
@@ -665,8 +668,9 @@ class Spectrum1DCollection(SpectrumCollectionMixin[Spectrum1D],
 
         """
         if len(spectra) < 1:
-            msg = 'At least one spectrum is needed for collection'
-            raise IndexError(msg)
+            x_data = ureg.Quantity(np.empty((0,)))
+            y_data = ureg.Quantity(np.empty((0,0)))
+            return cls(x_data, y_data, [], {})
 
         cls._item_type_check(spectra[0])
         x_data = spectra[0].x_data
@@ -762,7 +766,7 @@ class Spectrum1DCollection(SpectrumCollectionMixin[Spectrum1D],
                 width_lower_limit: Quantity | None = None,
                 width_convention: Literal['fwhm', 'std'] = 'fwhm',
                 width_interpolation_error: float = 0.01,
-                ) -> Self: ...
+                ) -> T: ...
 
     def broaden(self,
                 x_width,
@@ -771,7 +775,7 @@ class Spectrum1DCollection(SpectrumCollectionMixin[Spectrum1D],
                 width_lower_limit=None,
                 width_convention='fwhm',
                 width_interpolation_error=0.01,
-                ) -> Self:
+                ) -> T:
         """
         Individually broaden each line in y_data, returning a new
         Spectrum1DCollection
@@ -1030,11 +1034,10 @@ class Spectrum2DCollection(SpectrumCollectionMixin[Spectrum2D],
         """
 
         if len(spectra) < 1:
-            msg = format_error(
-                'No spectra provided.',
-                fix='Must have at least one spectrum for collection.',
-            )
-            raise IndexError(msg)
+            x_data = ureg.Quantity(np.empty((0,)))
+            y_data = ureg.Quantity(np.empty((0,)))
+            z_data = ureg.Quantity(np.empty((0,0,0)))
+            return cls(x_data, y_data, z_data, [], {})
 
         cls._item_type_check(spectra[0])
         bins_data = {
--- a/euphonic/spectra/collections.py
+++ b/euphonic/spectra/collections.py
@@ -13,6 +13,7 @@ from typing import (
     Any,
     Generic,
     Literal,
+    TypedDict,
     TypeVar,
     overload,
 )
@@ -129,9 +130,13 @@ class SpectrumCollectionMixin(ABC, Generic[Spec]):
     def _get_item_data_unit(cls, item: Spec) -> str:
         return getattr(item, f'{cls._spectrum_axis}_data_unit')
 
+    ## 3.15+ frozendict
+    class _BinDict(TypedDict):
+        _x_bins: Quantity
+
     @property
     @abstractmethod
-    def _bin_data(self) -> dict[str, Quantity]: ...
+    def _bin_data(self) -> _BinDict: ...
 
     def sum(self) -> Spec:
         """
@@ -164,12 +169,10 @@ class SpectrumCollectionMixin(ABC, Generic[Spec]):
 
     # Required methods
     @classmethod
-    @abstractmethod
     def from_spectra(
             cls,
             spectra: Sequence[Spec],
             *,
-            _x_bins: Quantity | None = None,
             unsafe: bool = False,
     ) -> Self:
         """Construct spectrum collection from a sequence of components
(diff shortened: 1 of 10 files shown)


ajjackson:
There's a design choice to make here regarding x-values. It looks like in this prototype an empty SpectrumCollection is typically constructed with a zero-size array of x-values. But this means that when we try to append another Spectrum or join to another collection it will fail due to non-matching x.

I think the options, from simplest to most-complex, are:
a) Require that all SpectrumCollection have x-values, even if y is empty. `.from_spectra()` will still fail on empty input.
b) As a), but provide an optional `x=` parameter to `.from_spectra()` so it is possible to instantiate an empty collection with `from_list([], x=x_bins)`.
c) Allow empty collections as a special-case. So whenever a collection with size-0 x is used, (or some other sentinel like `None`) the incoming x-values will automatically be adopted when combining/appending with another SpectrumCollection or Spectrum.

I think a) is actually fine; the main use-case for empty spectra is that they can be _produced_ by select() etc. in which case x-values are available. b) would just be sugar for something like `SpectrumCollection(x, np.zeros_like(x)[:, None])`. c) has the problem that either we allow more types into x (boooo) or create ambiguity about whether the empty data was created "naturally" (e.g. by split()) or represents a special case.

euphonic/spectra/collections.py
```diff
@@ -155,7 +158,11 @@ def sum(self) -> Spec:
     @classmethod
     @abstractmethod
     def from_spectra(
-            cls, spectra: Sequence[Spec], *, unsafe: bool = False,
+            cls,
+            spectra: Sequence[Spec],
+            *,
+            _x_bins: Quantity | None = None,
+            unsafe: bool = False,
```
ajjackson, line 177: Am I right in thinking that the undocumented `_x_bins` parameter will only be used by internal methods and is not recommended as part of public API?

Perhaps it would be more idiomatic to have a private `_from_spectra` which exposes this parameter and wrap it with public `from_spectra` which doesn't.

But then we have an abstract private method which isn't great either 🤔

euphonic/spectra/collections.py
```diff
@@ -399,9 +396,17 @@ def _combine_metadata(all_metadata: LineData) -> Metadata:
...
-                   (set(metadata.items()) for metadata in all_metadata)))
+        try:
+            common_metadata: Metadata = dict(
+                reduce(
+                    set.intersection,
+                    (set(metadata.items()) for metadata in all_metadata),
+                ))
+        except TypeError as err:
+            if 'empty iterable' in err.args[0]:
+                common_metadata = {}
```
ajjackson: Could this be covered with 
```python
if not all_metadata:
    common_metadata = {}
```
or are there some other cases to worry about?

euphonic/spectra/collections.py
```diff
@@ -664,8 +678,17 @@ def from_spectra(
...
+                            'providing _x_bins'),
+                    fix='Provide _x_bins.',
+                )
+                raise ValueError(msg)
+
+            y_data = ureg.Quantity(np.empty((0, len(_x_bins))))
+            return cls(_x_bins, y_data, [], {})
 
         cls._item_type_check(spectra[0])
         x_data = spectra[0].x_data
```
ajjackson, line 717: I note that `_x_bins` is not used if spectra are provided. In principle that's a validation opportunity missed, but... probably not worth the complexity for a case we might never use.

tests_and_analysis/test/euphonic_test/test_spectrum1dcollection.py
```diff
@@ -709,16 +712,6 @@ def test_select_with_not_all_matching_combinations_doesnt_raise_key_error(
...
-            [('quartz_666_pdos.json', fake_metadata,
-             {'inst': ['LET', 'TOSCA'], 'index': [4, 6]})])
-    def test_select_with_no_matches_raises_value_error(
-            self, spectrum_file, metadata, select_kwargs):
-        spec_col = get_spectrum1dcollection(spectrum_file)
-        spec_col.metadata = metadata
-        with pytest.raises(ValueError):
-            spec_col.select(**select_kwargs)
-
     @pytest.mark.parametrize(
```
ajjackson, line 716: I see this is replaced by `create_from_empty_select`, below.

I don't think we have a test that `select()` can be called on an empty spectrum, is that worth checking?

tests_and_analysis/test/euphonic_test/test_spectrum2dcollection.py
```diff
@@ -296,3 +296,72 @@ def test_collection_methods(self, quartz_fuzzy_collection):
...
         check_spectrum2d(selection.sum(), ref_item_2)
+
+
+class TestEmptyCollection:
+    @pytest.fixture
+    def empty_spectrum(self) -> Spectrum2DCollection:
+        return Spectrum2DCollection.from_spectra(
+            [],
+            _x_bins=ureg.Quantity(np.empty((10,))),
+            _y_bins=ureg.Quantity(np.empty((10,))),
```
ajjackson, line 320: I suggest making these different sizes so we know they don't get transposed somewhere.

tests_and_analysis/test/euphonic_test/test_spectrum2dcollection.py
```diff
@@ -296,3 +296,72 @@ def test_collection_methods(self, quartz_fuzzy_collection):
...
+        empty = real_spectrum[()]
+        assert len(empty) == 0
+
+    def test_add_empty(self, empty_spectrum):
+        assert len(empty_spectrum + empty_spectrum) == 0
+
+    def test_add_non_empty(self, empty_sized, real_spectrum):
+        spec = empty_sized + real_spectrum
+        assert len(spec) == len(real_spectrum)
+        assert np.all(spec.x_data == real_spectrum.x_data)
```
ajjackson: `np.allclose` gives better feedback when there are differences

pyproject.toml
```diff
@@ -44,7 +44,8 @@ dependencies = [
     "toolz>=0.12.1",
     "typing_extensions~=4.0"
 ]
-dynamic = ["version"]
+#dynamic = ["version"]
+version = "2.2.0"
```
ajjackson:
```suggestion
dynamic = ["version"]
```

ajjackson (commented):
A few comments here and there, but basically this looks brilliant.

I think it is worth reviewing alternatives to an `_x_bins` parameter, but it may well be the best option in the end.

euphonic/spectra/collections.py
oerc0122: Yeah, I did consider that, or migrating them to a `**kwargs` block so they aren't publicly known. Thoughts?

euphonic/spectra/collections.py
oerc0122: I'm not sure, is there such a thing as empty metadata for an extant line?

euphonic/spectra/collections.py
oerc0122: I did try that, but it was a lot of extra logic for a condition that should never happen. Here, I think simplicity won.

euphonic/spectra/collections.py
ajjackson: It looks like (informally) abstract private methods are used in [Scipy](https://docs.scipy.org/doc/scipy/reference/generated/scipy.integrate.OdeSolver.html) for solvers, and [Pandas](https://stijnvanhoey.github.io/pandas-getting-started-tutorials/reference/api/pandas.api.extensions.ExtensionArray.html) for custom array types.

It doesn't seem to be common but I don't see a big issue with it either.

euphonic/spectra/collections.py
```diff
@@ -664,8 +695,17 @@ def from_spectra(
 
         """
         if len(spectra) < 1:
-            msg = 'At least one spectrum is needed for collection'
-            raise IndexError(msg)
+            if _x_bins is None:
```
oerc0122, line 704: Does making this private mean we no longer want this error phrased like this since it may be confusing for users rather than devs?

euphonic/spectra/collections.py
ajjackson: If the collection just has empty metadata (fairly common) then iterating over lines will give a series of `{}`.

euphonic/spectra/collections.py
ajjackson: Yeah, it would be correct to say that `from_spectra` requires at least one spectrum. (But `_from_spectra` does not.)

oerc0122: I might move the error into `from_spectra` and leave this here, since if you're calling `_from_spectra` that error is more helpful.

tests_and_analysis/test/euphonic_test/test_spectrum2dcollection.py
ajjackson: (This ended up being in a different place but I think the coverage is sufficient.)

oerc0122: I thought there was more suitable since it was explicit. Could do both pretty trivially.

ajjackson (approved):
Great, I think this is ready to go! Just needs a rebase as the copy() branch got there first.

Shall I let you do the rebase? I don't trust the github UI for this any more... If you're busy I can pull and sort it out locally.

euphonic/spectra/collections.py
```diff
@@ -77,6 +77,7 @@ class SpectrumCollectionMixin(ABC, Generic[Spec]):
     _bin_axes = ('x',)
     _spectrum_axis = 'y'
     _item_type: type[Spec]
+    metadata: Metadata
```
ajjackson, line 80: I removed this in the other branch as it doesn't seem to do anything and potentially causes harm. The metadata is logically _very much_ an instance attribute rather than a class attribute. Any thoughts?

ajjackson: We can address in a separate PR, it doesn't break anything.

PR #505, opened 2026-05-12
Remove Spectrum.copy() and implement shallower Spectrum.__copy__

### LINKED ISSUE #453: `__copy__` methods deeper than necessary
I wonder if these `copy` are "deeper" in places than one might expect, but this does keep the backward-compatibility simple. Seems unlikely to lead to complaints.

_Originally posted by @ajjackson in https://github.com/pace-neutrons/Euphonic/pull/452#pullrequestreview-3693243475_

ajjackson:
Closes #453 

Dropping copy() is a tidy-up but pretty marginal. The main benefit is to reduce confusion... but does it, when users may find shallow copies confusing in their own right?

It's not entirely clear how _useful_ the shallow copies are, but they should be cheap to build!
We hope to have immutable objects in v3, for which `copy()` and `deepcopy()` will be redundant and the general workflow will be to work with the original object or spawn a modified copy with `__replace__`.

### DIFF
--- a/euphonic/spectra/base.py
+++ b/euphonic/spectra/base.py
@@ -636,8 +636,8 @@ class Spectrum1D(Spectrum):
 
     def __deepcopy__(self, memo: dict) -> Self:
         """Get a completely independent copy of spectrum"""
-        return type(self)(np.copy(self.x_data),
-                          np.copy(self.y_data),
+        return type(self)(self.x_data,
+                          self.y_data,
                           x_tick_labels=copy.deepcopy(
                               self.x_tick_labels, memo),
                           metadata=copy.deepcopy(self.metadata, memo))
@@ -1022,8 +1022,8 @@ class Spectrum2D(Spectrum):
                 width_convention=width_convention)
 
             spectrum = Spectrum2D(
-                np.copy(self.x_data),
-                np.copy(self.y_data),
+                self.x_data,
+                self.y_data,
                 ureg.Quantity(z_broadened, units=self.z_data_unit),
                 copy.copy(self.x_tick_labels),
                 copy.deepcopy(self.metadata),
@@ -1092,11 +1092,11 @@ class Spectrum2D(Spectrum):
         if axis == 'x':
             z_broadened = z_broadened.T
 
-        return Spectrum2D(np.copy(spectrum.x_data),
-                          np.copy(spectrum.y_data),
+        return Spectrum2D(spectrum.x_data,
+                          spectrum.y_data,
                           z_broadened,
-                          copy.copy(spectrum.x_tick_labels),
-                          copy.copy(spectrum.metadata))
+                          copy.deepcopy(spectrum.x_tick_labels),
+                          copy.deepcopy(spectrum.metadata))
 
     def __copy__(self) -> Self:
         """Get a shallow copy of spectrum
@@ -1116,9 +1116,9 @@ class Spectrum2D(Spectrum):
 
     def __deepcopy__(self, memo: dict) -> Self:
         """Get an fully-independent copy of spectrum"""
-        return type(self)(np.copy(self.x_data),
-                          np.copy(self.y_data),
-                          np.copy(self.z_data),
+        return type(self)(self.x_data,
+                          self.y_data,
+                          self.z_data,
                           copy.deepcopy(self.x_tick_labels, memo),
                           copy.deepcopy(self.metadata, memo))
 
--- a/euphonic/spectra/base.py
+++ b/euphonic/spectra/base.py
@@ -96,7 +96,7 @@ class Spectrum(ABC):
 
     def __mul__(self, other: Real) -> Self:
         """Get a new spectrum with scaled data"""
-        new_spec = self.copy()
+        new_spec = copy.deepcopy(self)
         new_spec *= other
         return new_spec
 
@@ -629,10 +629,14 @@ class Spectrum1D(Spectrum):
         Note that this is a a 'shallow' copy with the same underlying data:
         if you intend to mutate them, deepcopy is a better choice.
         """
-        return type(self)(self.x_data,
-                          self.y_data,
-                          x_tick_labels=copy.copy(self.x_tick_labels),
-                          metadata=copy.copy(self.metadata))
+        new = self.__new__(type(self))
+        for attr in ('_x_data', '_internal_x_data_unit', 'x_data_unit',
+                     '_y_data', '_internal_y_data_unit', 'y_data_unit',
+                     '_x_tick_labels', 'metadata',
+                     ):
+            setattr(new, attr, getattr(self, attr))
+
+        return new
 
     def __deepcopy__(self, memo: dict) -> Self:
         """Get a completely independent copy of spectrum"""
@@ -826,7 +830,7 @@ class Spectrum1D(Spectrum):
             msg = 'x_width must be a Quantity or Callable'
             raise TypeError(msg)
 
-        new_spectrum = self.copy()
+        new_spectrum = copy.deepcopy(self)
         new_spectrum.y_data = y_broadened
         return new_spectrum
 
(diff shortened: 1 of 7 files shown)


euphonic/spectra/base.py
```diff
@@ -624,16 +620,24 @@ def _split_by_indices(self,
...
+            setattr(new, attr, getattr(self, attr))
+
+        return new
 
     def __deepcopy__(self, memo: dict) -> Self:
         """Get a completely independent copy of spectrum"""
-        return type(self)(np.copy(self.x_data),
-                          np.copy(self.y_data),
+        return type(self)(self.x_data,
+                          self.y_data,
```
oerc0122, line 661: Does this run a risk of views modifying things?

euphonic/spectra/base.py
```diff
@@ -1088,25 +1092,33 @@ def _broaden_spectrum2d_with_function(
...
+            setattr(new, attr, getattr(self, attr))
+
+        return new
 
     def __deepcopy__(self, memo: dict) -> Self:
-        """Get an independent copy of spectrum"""
-        return type(self)(np.copy(self.x_data),
-                          np.copy(self.y_data),
-                          np.copy(self.z_data),
+        """Get an fully-independent copy of spectrum"""
```
oerc0122:
```suggestion
        """Get a fully-independent copy of spectrum"""
```
Or to comply with the other

```suggestion
        """Get a completely independent copy of spectrum"""
```

euphonic/spectra/collections.py
```diff
@@ -266,16 +266,13 @@ def _get_item_metadata(self, item):
         return self._combine_metadata([metadata_lines[i] for i in item])
 
     def __copy__(self) -> Self:
-        return type(self).from_spectra(list(self), unsafe=True)
+        # Borrow implementation from child tyoe: the data fields are the same
```
oerc0122:
```suggestion
        # Borrow implementation from child type: the data fields are the same
```

euphonic/spectra/collections.py
```diff
@@ -266,16 +266,13 @@ def _get_item_metadata(self, item):
...
+        return self._item_type.__copy__(self)
```
oerc0122: Is this guaranteed to work? I fear this runs the risk of separation if extra data are ever added to these.

euphonic/spectra/base.py
```diff
@@ -1088,25 +1092,33 @@ def _broaden_spectrum2d_with_function(
...
+
+        Note that this is a a 'shallow' copy with the same underlying data:
+        if you intend to mutate them, deepcopy is a better choice.
+        """
+        new = self.__new__(type(self))
+        for attr in ('_x_data', '_internal_x_data_unit', 'x_data_unit',
+                     '_y_data', '_internal_y_data_unit', 'y_data_unit',
+                     '_z_data', '_internal_z_data_unit', 'z_data_unit',
+                     '_x_tick_labels', 'metadata',
+                     ):
```
oerc0122: Maybe this should be a ClassVar of key components. This may also help sync XCollection vs X

oerc0122 (commented):
Looks quite neat, just a concern about np views

euphonic/spectra/base.py
ajjackson: There is a unit-test showing that modifications to elements of the copy create a difference in results to the original for `deepcopy` but not for `copy`. Are there some other cases we could/should check?

ajjackson: Perhaps bad things can happen if you set up the spectrum with a View in the first place 🤔

euphonic/spectra/collections.py
ajjackson: I guess it depends how comfortable we are with SpectrumNDCollection depending on implementation details of SpectrumND. They are quite closely connected and in practice when features of one are changed it leads to changes in the other.

It could be factored out to some generic copy_fields function for more safety. On th other hand I don't want to encourage further re-use than this 😆

euphonic/spectra/base.py
ajjackson: Would this be explicitly for `copy` and so named something like `_copy_attrs`?

I'm not sure that having SpectrumNDCollection rely on `SpectrumND._copy_attrs` is any less of a flaky implementation-detail-dependence than having it rely on `SpectrumND.__copy__`.

If SpectrumNDCollection has its own ClassVar of `_copy_attrs` this isn't practically different from including the list in its own `__copy__` method.

ajjackson: (I do think there might be a better structure around here somewhere, I just haven't quite spotted it yet.)

oerc0122: It could also be used in `__eq__` or associated methods.

euphonic/spectra/base.py
ajjackson: Things seem to be fine with deepcopy
```
In [34]: x = np.array([1, 2, 3])

In [35]: x_view = x.view()

In [36]: spec = Spectrum1D(Quantity(x_view, "meV"), Quantity([2, 3, 4], "W"))

In [37]: spec2 = copy.deepcopy(spec)

In [38]: spec2.x_data[2] = Quantity(4., 'meV')

In [39]: spec2.x_data  # is updated
Out[39]: <Quantity([1. 2. 4.], 'millielectron_volt')>

In [40]: spec.x_data  # is unaffected
Out[40]: <Quantity([1. 2. 3.], 'millielectron_volt')>

In [41]: spec.x_data[2] = Quantity(5., 'meV')

In [42]: spec.x_data  # is updated
Out[42]: <Quantity([1. 2. 5.], 'millielectron_volt')>

In [43]: x  # the array we viewed from was not updated
Out[43]: array([1, 2, 3])

In [44]: Quantity(x_view, "meV")[2] = Quantity(5, "meV")

In [45]: x  # but it is possible to mutate via Quantity of view
Out[45]: array([1, 2, 5])
```

ajjackson: shallow copy will change the array via a view but that's not a big surprise
```
In [46]: x = np.array([1, 2, 3])

In [47]: x_view = x.view()

In [48]: spec = Spectrum1D(Quantity(x_view, "meV"), Quantity([2, 3, 4], "W"))

In [49]: spec2 = copy.copy(spec)

In [50]: spec2.x_data[2] = Quantity(4., 'meV')

In [51]: spec.x_data
Out[51]: <Quantity([1. 2. 4.], 'millielectron_volt')>
```

oerc0122: That sounds like the behaviour we want!

euphonic/spectra/collections.py
oerc0122: I just don't like a dunder being applied to a different class, I think that's a recipe for desynchronisation between the classes even if non-critical.

euphonic/spectra/base.py
ajjackson: That's a nice idea!

ajjackson: Have implemented something like this with an `__eq__` method as well. Went with `_core_attrs` to indicate that it isn't only for `copy`.

euphonic/spectra/collections.py
ajjackson: Ok, with refactor this is now inherited from Spectrum

euphonic/spectra/base.py
```diff
@@ -100,14 +101,44 @@ def __mul__(self, other: Real) -> Self:
         new_spec *= other
         return new_spec
 
+    @property
     @abstractmethod
+    def _core_attrs(self) -> set[str]:
+        """Attribute names of underlying data; used for __copy__ and __eq__"""
```
oerc0122, line 107: I thought you'd strictly violated this spec with the other ClassVars not being properties, apparently I was wrong

https://discuss.python.org/t/provide-a-canonical-way-to-declare-an-abstract-class-variable/69416/2

tests_and_analysis/test/euphonic_test/test_spectrum1d.py
```diff
@@ -574,6 +574,19 @@ def test_mul(self):
         with pytest.raises(AssertionError):
             check_spectrum1d(spec, spec * 2.)
 
+    def test_compare_equal(self):
```
oerc0122, line 577: Just for completeness you might want to have a spectrum2d

oerc0122 (approved):
Looking good. I think this is probably ready

euphonic/spectra/base.py
ajjackson: Yeah, it does feel like this implementation is mixing them together more than usual, but... Declaring class variables is weird.

tests_and_analysis/test/euphonic_test/test_spectrum1d.py
ajjackson: I figured Spectrum1D and Spectrum2DCollection would cover everything. But they're small and fast tests so no harm in expanding it a bit.

ajjackson: Done!

oerc0122: Oh, I meant compare a Spectrum1D to a Spectrum2D with `==` and check it's false.

ajjackson: Good point! Will pop one in.

oerc0122 (approved):

PR #507, opened 2026-05-14
Remove metadata class variable

ajjackson:
Apologies for pushing this to release-2.0 by mistake, have reverted there.

I'm still a bit unsure about why this line is there and whether it is a problem.


oerc0122:
> Apologies for pushing this to release-2.0 by mistake, have reverted there.
> 
> I'm still a bit unsure about why this line is there and whether it is a problem.

It is a type hint that if there is a `metadata` attribute on `self`, it will be of type `Metadata`, it should not do anything at runtime.

ajjackson:
Cool, should be innocuous then 😄

PR #509, opened 2026-05-29
Implement a more complete Sears-1992 dataset using CSV reference

### LINKED ISSUE #508: Implement more complete isotope dataset
Following #494 : use this infrastructure to create Abins-equivalent capability to select appropriate data using masses to distinguish between isotopes.

It would be good to have this ready for 2.0 so it can be made the default dataset.
But not the end of the world if it goes into 2.1; downstream packages can require it and ask explicitly.

ajjackson:
Closes #508 

The original plan was to use a dump of data from Mantid, but numerous minor inconsistencies were found compared with the Sears reference document. ([These have been reported.](https://github.com/mantidproject/mantid/issues/41462)) This implementation includes a new digitization of that 1992 dataset; other more up-to-date sources could be added to follow.

This branch also implements a cleaner solution to import loops between euphonic.data.isotopes and euphonic.util: the offending function is moved to its own file and is now the only place imports have to be deferred.


ajjackson:
There is a bit more to do in terms of linking the new dataset to CLI and replacing the equivalent JSON implementation in places (as long as we validate the numbers match!)

But this is already a lot more code than I'd hoped it would take, and seems a good point for reviewer input. I've prioritised clarity over conciseness, but it would be nice to shrink things a little. I'm also not entirely happy with where the code is all placed: it would be good to split out the CSV IsotopeData from the others into its own file. But already I'm a bit uneasy about having `euphonic/data/isotopes.py` when previously this was just a directory for data files. That's all happened on the 2.0 branch so there is still a chance to change it.

euphonic/data/_legacy.py
```diff
@@ -0,0 +1,85 @@
...
+    data = _get_all_dicts_from_json(collection)
+
+    if physical_property not in data:
+        msg = format_error(
+            (
+                f'No such collection {collection!r} '
+                f'with property {physical_property!r}.'
+            ),
+            fix=(
+                f'Available properties for this collection:{comma_join(data)}.'
```
oerc0122:
```suggestion
                f'Available properties for this collection: {comma_join(data)}.'
```

euphonic/data/isotopes.py
```diff
@@ -169,94 +191,411 @@ class LegacyJsonData(AtomTypeDictData):
...
+    @staticmethod
+    def _split_unit(col_header: str) -> tuple[str, str | None]:
+        """Split unit from column name if present
+
+        e.g.::
+            'name  '       -> 'name', None
+            'name (unit)'  -> 'name', '(unit)'
+        """
+        if match := re.match(
+            r"""(.+?)      # Mandatory NAME; any characters, non-greedy
```
oerc0122: Do you want an all-capture here? or a `\w`? You may find you need less documentation if you used named captures:
```python
r"""(?P<name>\w+?)\s*(?:\((?P<unit>.+)\))?"""

File: euphonic/data/isotopes.py [hunk 3]
```diff
@@ -169,94 +191,411 @@ class LegacyJsonData(AtomTypeDictData):
...
+                               #
+                   \s*         # Any amount of whitespace between NAME and UNIT
+                               #
+                   (           # Begin unit group
+                               #
+                   \(.+\)    # At least one character surrounded by parens
```
oerc0122: This also allows whitespace.

euphonic/data/isotopes.py
```diff
@@ -169,94 +191,411 @@ class LegacyJsonData(AtomTypeDictData):
...
+                               #
+                   )?$         # Optional group must complete the input string;
+                               # this ensures whole line is used despite
+                               # non-greedy NAME.  Otherwise we can get
+                               # 'NAME (UNIT)' -> ('N', None)
+                """,
+            col_header,
+            re.VERBOSE,
+        ):
+            name, unit = match.groups()
```
oerc0122: This would become:

```suggestion
            name, unit = match.groups("name", "unit")
```

euphonic/data/isotopes.py
```diff
@@ -169,94 +191,411 @@ class LegacyJsonData(AtomTypeDictData):
...
+        stored in th MISSING attribute.
+
+        Parameters
+        ----------
+
+        csv_file: reference data set, such as sears-1992.csv
+        property_columns: mapping of property names to column headers in CSV.
+            The property names *must* include symbol, mass and a_number; these
+            are used to select and organise the data.
+
```
oerc0122, line 353: Halfway between google and numpy?

euphonic/data/isotopes.py
```diff
@@ -169,94 +191,411 @@ class LegacyJsonData(AtomTypeDictData):
...
+            msg = format_error(
+                f'No data found for {symbol!r} in {self._csv_file}.',
+                fix='Ensure symbol corresponds to an element in the table.',
+            )
+            raise ValueError(msg)
+        return symbol_rows
+
+    def _get_nearest_row(self, symbol: str, mass: float) -> np.recarray:
+        symbol_rows = self._get_symbol_rows(symbol)
+        return symbol_rows[np.argmin(np.abs(symbol_rows['mass'] - mass))]
```
oerc0122: Does this actually want to yield from rows sorted by difference so you can find the matches in case data are missing?

euphonic/data/isotopes.py
```diff
@@ -169,94 +191,411 @@ class LegacyJsonData(AtomTypeDictData):
...
+        ----------
+        symbol:
+            element symbol e.g. 'Hg'
+        mass:
+            mass in a.m.u. e.g. 200.6. This is used to identify the isotope or
+            standard isotope mixture (i.e. row) from reference data table.
+        key:
+            property name e.g. 'coherent_cross_section'
+        """
+        item = self._get_item(symbol, mass)
```
oerc0122, line 455: Would `item` be better named `row`?

euphonic/data/isotopes.py
```diff
@@ -169,94 +191,411 @@ class LegacyJsonData(AtomTypeDictData):
...
 
+        if isinstance(value, str):
+            msg = format_error(
+                'This method cannot return str values.',
+                fix=(
+                    'Choose another property, or read '
+                    f'{self._csv_file} by other means.'
+                ),
+            )
+            raise NotQuantityError(msg)
```
oerc0122, line 476: Do you want a `get_raw` to handle this use-case just in case?

euphonic/data/isotopes.py
```diff
@@ -169,94 +191,411 @@ class LegacyJsonData(AtomTypeDictData):
...
+        value_table = {
+            (symbol, mass): self.get_value(symbol, mass, key).magnitude
+            for (symbol, mass) in targets
+        }
 
-    data = _get_all_dicts_from_json(collection)
+        result_raw = np.empty_like(
+            structure.atom_mass,
+            dtype=type(  # int, real or complex
+                next(iter(value_table.values()))
```
oerc0122:
```suggestion
                toolz.itertoolz.first(value_table.values())
```

euphonic/data/isotopes.py
```diff
@@ -169,94 +191,411 @@ class LegacyJsonData(AtomTypeDictData):
...
+
+    @classmethod
+    def _normalise_record(cls, line: str, types: list[type]) -> tuple:
+        record = line.strip().split(',')
+
+        # We can't deal with uncertain values, these should become NaN
+        bad_complex = re.compile(r'\(?[±<>].*$')
+        bad_float = re.compile(r'[±<>].*$')
+
+        def _cast(item: str, item_type: type) -> int | float | complex | str:
```
oerc0122, line 561: Bit messy, because of the `object` situation. Is that ever used?
```suggestion
        T = TypeVar("T", int, float, complex, str)
        @overload
        def _cast(item: str, item_type: type[object]) -> str: ...
        @overload
        def _cast(item: str, item_type: type[T]) -> T: ...
        def _cast(item, item_type):
```
If that weren't the case:

```suggestion
        T = TypeVar("T", int, float, complex, str)
        def _cast(item: str, item_type: type[T]) -> T:
```

And in 3.12+

```suggestion
        def _cast[T: (int, float, complex, str)](item: str, item_type: type[T]) -> T:
```

euphonic/data/isotopes.py
```diff
@@ -169,94 +191,411 @@ class LegacyJsonData(AtomTypeDictData):
...
+    def _table_and_headers(
+        self,
+    ) -> tuple[np.recarray, dict[str, CsvColumnInfo]]:
+        # Map of CSV columns to rename: inverse of user-provided map
+        name_map = {value: key for key, value in self._property_map.items()}
+
+        with self._csv_file.open('rt', encoding='utf-8') as fd:
+            col_names = next(fd).strip().split(',')
+            col_types = next(fd).strip().split(',')
+            records = fd.readlines()
```
oerc0122: To preprocess the csv, it may(?) be better to use
`csv.reader` and iterate over the data rather than processing it manually? 

It saves little beyond manual stripping and splitting, but should be more resilient to CSV dialects in principle.

euphonic/data/_legacy.py
```diff
@@ -0,0 +1,85 @@
+# ruff: noqa: PLC0415
```
oerc0122: Does this also want to move into the pyproject.toml?

euphonic/data/isotopes.py
ajjackson: It wouldn't be unreasonable to have something like

`speed (km / s)`.

oerc0122: Would it be unreasonable to have
`count (  )`
?

euphonic/data/isotopes.py
ajjackson: Yes the `object` situation will _always_ come up because of the `symbol` field.

euphonic/data/_legacy.py
ajjackson: We generally _do_ want this rule, which is import-outside-top-level.

This is a slightly naughty file which will be removed in v3, so is allowed to be an exception.

It would indeed be nicer to declare the exception in pyproject.toml so we can easily get a handle on how many special cases are kicking around.

oerc0122: Sorry, I meant as another file-specific exception.

ajjackson: I followed, just talking through the justification for posterity 😄

euphonic/data/isotopes.py
ajjackson: Not overly worried; error or Dimensionless would both be reasonable outcomes. I think it would give a Pint error at the moment because `Unit("( )")` isn't allowed. But `Unit(" ")` resolves to Dimensionless which is not a bad interpretation; we could strip the parens to get that fallback behaviour.

euphonic/data/isotopes.py
ajjackson: How about documentation _and_ named captures?

I don't think it hurts to assume that the average Euphonic contributor is less fluent in line noise than the average castep-outputs contributor 😆

oerc0122: Not a *bad* compromise, I just feel that the less documented and illegible an re, the closer we come to the perfection of Perl. 

(Note that here, by the way, I only capture the inner contents of the parens, rather than the whole chunk)

ajjackson: Ok, have gone for
- \w+ on name, that should cover reasonable characters (no emoji 😞 , but greek letters and accented letters seem ok)
- named captures for extra clarity
- still a massive long commented block to help newcomers

euphonic/data/isotopes.py
ajjackson: I think the updated regex with your grouping arrangement does this very trick. Seems harmless?

euphonic/data/isotopes.py
ajjackson: Fixed. But I don't love it.

euphonic/data/isotopes.py
ajjackson: I think what it actually wants to do is
- raise an error if the distance was too large
- otherwise, print some information about the match that was found to STDOUT so users have a chance to see if there are surprises

In this implementation I seem to have accidentally solved the "mass matches rare isotope when you probably meant standard mixture" problem in a different way to Abins: I just removed isotopes that have no data. There don't seem to be any remaining cases where it would be reasonable to overlook the nearest isotope and go to standard mixture... perhaps one will turn up and we deal with it then.

euphonic/data/isotopes.py
ajjackson: > (Note that here, by the way, I only capture the inner contents of the parens, rather than the whole chunk)

Already noted, and I think this helps robustness w.r.t. whitespace and empty values.

euphonic/data/isotopes.py
ajjackson: Ok have implemented that and pushed. I'll leave this conversation open until you have a look!

euphonic/data/isotopes.py
```diff
@@ -393,9 +400,30 @@ def _get_symbol_rows(self, symbol: str) -> np.recarray:
             raise ValueError(msg)
         return symbol_rows
 
+    @classmethod
+    def _format_row_name(cls, row: np.record) -> str:
+        if row.a_number == cls.MISSING[int]:
```
oerc0122: I would explicitly use `_is_missing` just for consistency.

euphonic/data/isotopes.py
```diff
@@ -393,9 +400,30 @@ def _get_symbol_rows(self, symbol: str) -> np.recarray:
...
+            return f'{row.symbol}'
+        return f'{row.symbol}-{row.a_number}'
+
     def _get_nearest_row(self, symbol: str, mass: float) -> np.recarray:
         symbol_rows = self._get_symbol_rows(symbol)
-        return symbol_rows[np.argmin(np.abs(symbol_rows['mass'] - mass))]
+        nearest = symbol_rows[np.argmin(np.abs(symbol_rows['mass'] - mass))]
+
+        if np.abs(nearest['mass'] - mass) > self.MASS_MATCH_THRESHOLD:
```
oerc0122: I think the meaning is better expressed as:
```python
if not np.isclose(nearest['mass'], mass, atol=self.MASS_MATCH_THRESHOLD):
```

euphonic/data/isotopes.py
ajjackson: I'm generally using `row` to mean a np.record. `._get_item` produces quite a similar object to the dict output of `get_item()`, it just hasn't had the missing values stripped out.

euphonic/data/isotopes.py
ajjackson: I don't currently feel confident enough in that use-case to decide the API for it. We can backward-compatibly add methods or options to get string values at a later stage?

euphonic/data/isotopes.py
ajjackson: Good shout, I initially tried to do this with pattern-matching so needed a value.

euphonic/data/isotopes.py
ajjackson: Much better

euphonic/data/isotopes.py
```diff
@@ -169,94 +195,452 @@ class LegacyJsonData(AtomTypeDictData):
...
+        """Split unit from column name if present
+
+        e.g.::
+            'name  '       -> 'name', None
+            'name (unit)'  -> 'name', '(unit)'
+        """
+
+        if match := re.match(
+            r"""
+                (?P<name>\w+?)  # Mandatory NAME; any characters, non-greedy
```
oerc0122:
```suggestion
                (?P<name>\w+?)  # Mandatory NAME; any word characters, non-greedy
```

euphonic/data/isotopes.py
ajjackson: Done!

ajjackson:
Ok, the new data should now be in use for command-line tools.

I think the remaining thing is to update https://euphonic.readthedocs.io/en/stable/utils.html#reference-data and this requires some decisiveness around file names and locations 😅

Some options:
## Mixed data and code
- euphonic/data contains .json, .csv files
- euphonic/data/isotopes.py, euphonic/data/_legacy.py contain code

## No code in data/, move up a level
- euphonic/data contains .json, .csv
- euphonic/data/isotopes.py -> euphonic/isotopes.py is the current euphonic/data/isotope.py
- euphonic/data/_legacy.py -> euphonic/_legacy_data_loader.py

## Add another level of structure
- euphonic/data contains .json, .csv
- euphonic/data/isotopes.py -> euphonic/isotopes/core.py + euphonic/data/isotopes/csv.py
- euphonic/data/_legacy.py -> euphonic/isotopes/_legacy.py
- euphonic/isotopes/\_\_init\_\_.py could contain some exported useful/more-public bits.

euphonic/data/_legacy.py
```diff
@@ -0,0 +1,85 @@
...
+euphonic.data.isotopes while avoiding import loops with euphonic.util.
+
+"""
+
+from euphonic.ureg import Quantity
+
+
+def get_reference_data(
+    collection: str = 'Sears1992',
+    physical_property: str = 'coherent_scattering_length',
```
oerc0122, line 16: Are you sure you want a default here?

euphonic/data/isotopes.py
```diff
@@ -169,94 +196,452 @@ class LegacyJsonData(AtomTypeDictData):
...
+        """Split unit from column name if present
+
+        e.g.::
+            'name  '       -> 'name', None
+            'name (unit)'  -> 'name', '(unit)'
+        """
+
+        if match := re.match(
+            r"""
+              (?P<name>\w+?)  # Mandatory NAME; any word characters, non-greedy
```
oerc0122: Strictly this doesn't need to be non-greedy anymore since `\w` won't match `\s` or `\(`.

euphonic/data/isotopes.py
```diff
@@ -169,94 +196,452 @@ class LegacyJsonData(AtomTypeDictData):
...
+    @staticmethod
+    def _split_unit(col_header: str) -> tuple[str, str | None]:
+        """Split unit from column name if present
+
+        e.g.::
+            'name  '       -> 'name', None
+            'name (unit)'  -> 'name', '(unit)'
+        """
+
+        if match := re.match(
```
oerc0122: Just remembered that what you want here is [`re.fullmatch`](https://docs.python.org/3/library/re.html#re.fullmatch)

euphonic/data/isotopes.py
```diff
@@ -169,94 +196,452 @@ class LegacyJsonData(AtomTypeDictData):
...
+
+        Standard element mixtures should have no a_number. Monoisotopic
+        elements (or elements that lack data for other isotopes) should be
+        treated as isotopes, including all three properties.  Elements without
+        a standard isotopic mixture (such as Tc) should be expressed as a set
+        of isotopes.
+
+        Symbols are permitted to include comments/variations after a colon
+        character, e.g. {symbol='H:2', mass=2.0} might conveniently indicate
+        a deuterium nucleus.
```
oerc0122, line 333:
```suggestion
        a deuterium atom.
```
? Not sure if electrons matter here, but I would think so?

euphonic/data/isotopes.py
```diff
@@ -169,94 +196,452 @@ class LegacyJsonData(AtomTypeDictData):
...
+    def from_raw(
+        cls, raw_name: str, raw_dtype: str, name_map: dict[str, str]
+    ) -> Self:
+        """Convert items from zipped CSV headers
+
+        Parameters
+        ----------
+
+        raw_name:
+          e.g. 'b_inc (fermi)'
```
oerc0122: Only 2-space indent?

euphonic/data/isotopes.py
```diff
@@ -169,94 +196,452 @@ class LegacyJsonData(AtomTypeDictData):
...
+        """
+        self._csv_file = csv_file
+        self._property_map = property_map
+
+    MISSING: ClassVar[dict[str, Any]] = {
+        int: -(2**31),
+        float: float('-Inf'),
+        complex: complex(float('-Inf'), float('-Inf')),
+    }
+    INVALID_FLOAT = float('NaN')
```
oerc0122: Does this also want to be `ClassVar`? (or `Final`)

euphonic/data/isotopes.py
```diff
@@ -169,94 +196,452 @@ class LegacyJsonData(AtomTypeDictData):
...
+                )
+            msg = format_error(
+                summary,
+                fix='Check isotope is correct, or use another data source.',
+            )
+            raise ValueError(msg)
+
+        return value
+
+    def get_array(self, structure: Structure, key: str) -> Quantity:
```
oerc0122, line 504: Docstring from parent?

euphonic/data/isotopes.py
```diff
@@ -169,94 +196,452 @@ class LegacyJsonData(AtomTypeDictData):
...
+
+        This may include invalid (NaN) values, but *missing* values will be
+        omitted from the dict.
+        """
+        item = self._get_item(symbol, mass)
+
+        def _allowed_value(value: str | Quantity) -> bool:
+            return not (
+                isinstance(value, str) or self._is_missing(value.magnitude)
+            )
```
oerc0122: Wonder if this wants to become a private classmethod

euphonic/data/isotopes.py
```diff
@@ -169,94 +196,452 @@ class LegacyJsonData(AtomTypeDictData):
...
+        nearest_row = self._get_nearest_row(symbol, mass)
+
+        values = (
+            (col_info.name, col_info._apply_unit(item))
+            for col_info, item in zips(
+                self._column_headers.values(), nearest_row
+            )
+        )
+
+        return dict(values)
```
oerc0122:
```suggestion
        return {
            col_info.name: col_info._apply_unit(item)
            for col_info, item in zips(
                self._column_headers.values(), nearest_row
            )
        }
```

euphonic/data/isotopes.py
```diff
@@ -169,94 +196,452 @@ class LegacyJsonData(AtomTypeDictData):
...
+
+    # We can't deal with uncertain values, these should become NaN
+    _bad_complex = re.compile(r'\(?[±<>].*$')
+    _bad_float = re.compile(r'[±<>].*$')
```
oerc0122: While they're used here, I would probably position these with the other class-level constants.

oerc0122 (commented):
This all looks good, no functional suggestions, just a couple of oddities.

euphonic/data/isotopes.py
ajjackson: I think it's ok for a subclass to change them, so not Final

euphonic/data/_legacy.py
ajjackson: This is the deprecated backward-compatibility function, so yeah might as well keep it for now.

euphonic/data/isotopes.py
ajjackson: These are nuclear properties rather than atomic properties. The atom has the same properties, but "nucleus" seems more precise.

euphonic/data/isotopes.py
ajjackson: sure why not

euphonic/data/isotopes.py
ajjackson: Surprising that the linter allowed this, it complained about similar things before!

PR #511, opened 2026-06-05
Move files around

ajjackson:
Remove data module, move content to ureg/ and isotopes/.
This gives a generally cleaner structure: data files live in /data/ subdirectories of their relevant packages.

Annoyingly it does bring back some of the import-loop headache for legacy data loader; isotopes/_legacy.py can't be loaded independently of isotopes/__init__.py which creates a circular dependence on euphonic.util.  The long-term fix is clear (drop isotopes._legacy) and the real sin was having functionality of util depend on another module.

The workaround here is to move that problematic import to the end of *util.py* so that the required functions for isotopes/__init__.py have already been loaded.


.gitignore
```diff
@@ -1,3 +1,4 @@
+.DS_Store
```
oerc0122, line 1: What is this file?

oerc0122 (approved):
All looks good to me

.gitignore
ajjackson: It's used by MacOS to track where icons were dragged in the file browser.

oerc0122: Why would you leave the terminal? What's a file browser?

PR #512, opened 2026-06-07
Enable FBT ruff rule

ajjackson:
Most of these are straightforward.

I made exceptions for a couple of cases where the functions is clearly internal/private and has a large number of arguments, such that calling them as kwargs would actively harm readability while risk of API desync is low.

Note that I have made *all* the parameters keyword-only to many "import phonopy" functions. I'm not loving the current path + filename arrangement and would prefer to do something more Pathlib-native - but don't have a precise proposal for replacement, yet. It seems reasonably likely then that this will involve additional keyword(s) while current one(s) are deprecated; mandatory kwarg should make that transitions safer.

### DIFF
--- a/euphonic/force_constants.py
+++ b/euphonic/force_constants.py
@@ -447,20 +447,21 @@ class ForceConstants:
             return qpt_freqs, grads
         return qpt_freqs
 
+
     def _calculate_phonons_at_qpts(  # noqa: C901
             self,
             qpts: np.ndarray,
             weights: np.ndarray | None,
             asr: Literal['realspace', 'reciprocal'] | None,
-            dipole: bool,
+            dipole: bool,                # noqa: FBT001
             dipole_parameter: float,
-            splitting: bool,
-            insert_gamma: bool,
-            reduce_qpts: bool,
-            use_c: bool | None,
+            splitting: bool,             # noqa: FBT001
+            insert_gamma: bool,          # noqa: FBT001
+            reduce_qpts: bool,           # noqa: FBT001
+            use_c: bool | None,          # noqa: FBT001
             n_threads: int | None,
-            return_mode_gradients: bool,
-            return_eigenvectors: bool) -> tuple[
+            return_mode_gradients: bool, # noqa: FBT001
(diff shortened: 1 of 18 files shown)


euphonic/force_constants.py
```diff
@@ -447,20 +447,21 @@ def calculate_qpoint_frequencies(
...
         return qpt_freqs
 
+
     def _calculate_phonons_at_qpts(  # noqa: C901
             self,
             qpts: np.ndarray,
             weights: np.ndarray | None,
             asr: Literal['realspace', 'reciprocal'] | None,
-            dipole: bool,
+            dipole: bool,                # noqa: FBT001
```
oerc0122, line 456: > The following is currently only available in [preview mode](https://docs.astral.sh/ruff/linter/preview.md).
> 
> To cover an entire "logical" line (a multi-line statement or suite header), an "ignore" comment may be placed above the first line:
> 
> ```
> # ruff: ignore[ARG001]  # Covers the entire function signature
> def foo(
>     arg1,
>     arg2,
> ):
>     pass
> 
> # ruff: ignore[E501]  # Covers the entire list literal
> things = [
>     "really long string literal ...",
>     "really long string literal ...",
> ]
> ```

...

> ```
> # ruff: disable[E501]
> VALUE_1 = "Lorem ipsum dolor sit amet ..."
> VALUE_2 = "Lorem ipsum dolor sit amet ..."
> VALUE_3 = "Lorem ipsum dolor sit amet ..."
> # ruff: enable[E501]
> ```

Good to know for future

oerc0122 (approved):
Looks good. Sorry for all the extra work, but I think it's important we get this in now and establish it before we have to wait for 3.0.0

euphonic/force_constants.py
ajjackson: Yeah, I tried enable/disable and it didn't work properly with the current version of ruff used with Euphonic CI. I intend to bump that *after* release. Logical handling of a function def would be even better.

ajjackson:
Not at all, glad you brought it up!

Looks like we'll be clear to release tomorrow :)

PR #513, opened 2026-06-08
Release 1.6.2

ajjackson:
Collect some outstanding things from master branch before dealing with v2 release.


oerc0122 (approved):

ajjackson:
Tempted to squash this as it's mostly nonsense. But would mean the release tag is not on `master` history.

oerc0122:
> Tempted to squash this as it's mostly nonsense. But would mean the release tag is not on `master` history.

It's the final (non-emergency bugfix) release? But I don't think it's that essential to squash if it's going to cause desyncs.

PR #519, opened 2026-06-09
Add an untested function, fiddle with test config

ajjackson:
Following success with abins-lib, see if we can get orgoro coverage action fully working here as well!

The main problem was related to name mismatch between tested files in coverage.xml and in git diff. There are a few parameters to help with the mapping.


ajjackson:
After a lot of fiddling and some Github copiloting things still seem to be complicated and broken. Think this might need a do-over 😮‍💨

PR #520, opened 2026-07-24
Add Ubuntu ARM support

ajjackson:
I'm currently spending a lot of time in Linux containers on my ARM Macbook and it's pretty annoying that Euphonic doesn't have a pre-built wheel for this platform 😅


ajjackson:
Having more success in #521, can close this attempt.

PR #521, opened 2026-07-25
Support Linux on ARM, fix Windows builds, streamline tox setup.

### LINKED ISSUE #504: Unify min_requirements tox envs
I guess this means we no longer need separate tox environment specs for minrequirements on Mac and Linux; we can just have a global minrequirements env and the dependency group magic should take care of things.

_Originally posted by @ajjackson in https://github.com/pace-neutrons/Euphonic/pull/497#discussion_r3215444542_

ajjackson:
I'm increasingly working in Linux containers on my ARM Mac laptop, so lack of Euphonic wheels is annoying.

It turns out there are issues with the C extension on this platform, and with brille building/interface.

- Tweak typing in C code, it looks some complex numbers were only allocated correct space by good fortune and this fails on ARM Linux
- Streamline tox.ini in general; we don't actually need these Mac special-cases

### DIFF
--- a/.github/workflows/build_wheels.yml
+++ b/.github/workflows/build_wheels.yml
@@ -73,23 +73,13 @@ jobs:
           echo LDFLAGS="-L${LLVM_DIR}/lib $LDFLAGS" >> $GITHUB_ENV
           echo CPPFLAGS="-I${LLVM_DIR}/include $CPPFLAGS" >> $GITHUB_ENV
 
-      - name: Windows - find MSVC and set environment variables
-        if: startsWith(matrix.os, 'windows')
-        shell: bash -l {0}
-        env:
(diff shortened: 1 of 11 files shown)


ajjackson:
Wheels all building!

image

ajjackson:
Have taken a deeper human dive into the out-of-bounds error identified by LLMs; it seems reasonable and actually a bit surprising it never caused a problem before. Hopefully the updated C code is also slightly clearer.

ajjackson:
Something funny is going on in the build/test wheels action, these wheels have the wrong version number.

ajjackson:
Resolved: because 2.0.0 is in a release branch, building wheels from here sees the wrong tag and builds the wrong version number. The workaround is to add a tag; in this case `2.0.1.dev1`.

This problem doesn't affect regular release builds as they will always be done on top of a fresh tag. I think the workaround is ok (we don't really use dev builds for anything else) but should be documented somehow; I created an Issue in the pace developer docs.

PR #522, opened 2026-08-04
Vasp hdf5 import

ajjackson:
Recent (relatively-speaking) versions of VASP have improved phonon calculation features and a machine-friendly .h5 file output. This makes it a lot more attractive to include some built-in support. A tricky point is that various calculation types will produce an identically-named .h5 file so a bit of intelligence is needed to identify if relevant information is available:

- An IBRION=[5-8] calculation will produce a Hessian for the whole simulation cell. VASP might identify the supercell nature and use this to reduce the number of displacements, but we still get a 3Nx3N set of force constants. For now we do not look for a primitive cell in this case, and treat the system as though it is primitive. That will give correct DOS and INS simulations for a supercell calculation, but at increased cost compared with a more sophisticated scheme. (Band structures will include folding and incorrect symmetry labels.)

- If QPOINTS are provided and PHON_DOS is set, VASP will do q-point interpolation and produce a dataset with defined primitive cell and sufficient data to construct QpointPhononModes. The force constants _may_ be present, but can also be missing if the file was produced in a "post-processing" run on an existing .h5 file.

- If both force constants and primitive cell are defined, we can _attempt_ to map the force constants to primitive as we do for Phonopy. Less metadata is available so for now we naively try to solve for the supercell matrix. This might go wrong in non-trivial cases, but so do some Phonopy imports at the moment; probably they can be fixed together in future. In (fairly common) trivial diagonal supercell cases we should be fine.

https://vasp.at/wiki/Computing_the_phonon_dispersion_and_DOS

h5py is an optional dependency included in the existing phonopy-reader dependency group, so we re-use that in packaging/CI infrastructure. (phonopy-reader _also_ includes YAML, but VASP doesn't need it.)

At the moment I've put the tests in a new vasp_reader group. But perhaps life would be simpler if they shared phonopy_reader.


ajjackson:
I note that coverage reporting seems to be totally broken. This isn't even a fork and it's still failing to pick up the changed files correctly 😞

euphonic/readers/vasp.py
```diff
@@ -0,0 +1,577 @@
...
+    ------
+    ValueError
+        If atomic masses (POMASS) cannot be found in INCAR or POTCAR datasets.
+    """
+    filename = Path(h5_file.filename)
+
+    # 1. Try active input/incar/POMASS
+    if 'input/incar/POMASS' in h5_file:
+        val = h5_file['input/incar/POMASS'].asstr()[()]
+        raw_vals = re.findall(r'[0-9.]+', str(val))
```
oerc0122: Does this need casting to `str` again? 

Also, would it be "better" to use `finditer`, build the list from that, then check whether the list has elements before returning?

euphonic/readers/vasp.py
```diff
@@ -0,0 +1,577 @@
...
+            raise ValueError(msg)
+        return [float(mass) for mass in raw_vals]
+
+    # 2. Try original/incar/content
+    if 'original/incar/content' in h5_file:
+        incar_content = h5_file['original/incar/content'].asstr()[()]
+        match = re.search(
+            r'POMASS\s*=\s*(?P<masses>[0-9.\s,]+)', incar_content
+        )
+        if match:
```
oerc0122:
```suggestion
        if not (match := re.search(
            r'POMASS\s*=\s*(?P<masses>[0-9.\s,]+)', incar_content
        )):
```

ajjackson:
Woohoo, rebased and the new coverage reporter seems to be working 🎉

euphonic/readers/vasp.py
```diff
@@ -0,0 +1,577 @@
...
+                    fix='Ensure POMASS contains valid numeric values.',
+                )
+                raise ValueError(msg)
+            return [float(mass) for mass in raw_vals]
+
+    # 3. Try POTCAR content stored inside the HDF5 file
+    if 'input/potcar/content' in h5_file:
+        potcar_content = h5_file['input/potcar/content'].asstr()[()]
+        matches = re.findall(r'POMASS\s*=\s*(?P<mass>[0-9.]+)', potcar_content)
+        if matches:
```
oerc0122:
```suggestion
        if matches := re.findall(r'POMASS\s*=\s*(?P<mass>[0-9.]+)', potcar_content):
```

euphonic/readers/vasp.py
```diff
@@ -0,0 +1,577 @@
...
+        fix=(
+            'Ensure POMASS is set in INCAR, or that POTCAR contains '
+            'POMASS data.'
+        ),
+    )
+    raise ValueError(msg)
+
+
+def _read_cell_from_group(
+    h5_file: 'h5py.File', group_path: str
```
oerc0122: Do you need forward declaration if it's in a `TYPE_CHECKING` block?

euphonic/readers/vasp.py
ajjackson: Nope, the `str` is redundant. I've refactored this function to be a bit flatter, DRYer with more walrus; take a look!

euphonic/readers/vasp.py
ajjackson: Did some other stuff

euphonic/readers/vasp.py
ajjackson: ibid.

euphonic/readers/vasp.py
ajjackson: Making safe forward declarations is `__future__` annotations's job, and usually required to do useful things with TYPE_CHECKING. Implemented and removed the string quotes.

euphonic/readers/vasp.py
```diff
@@ -113,7 +119,25 @@ def _open_vasp_h5(filename: Path):
         yield h5_file
 
 
-def _extract_pomass(h5_file: 'h5py.File') -> list[float]:
+def _parse_masses(content: str, pattern: str) -> list[float]:
```
oerc0122, line 122: *very* slight "optimisation" might be to accept `re.Pattern` as well, and pre-`re.compile` the RE patterns.

oerc0122 (approved):
Looks good.

euphonic/readers/vasp.py
ajjackson: It might even be worse.
- They're literally going to be called once in 95% of applications
- Python caches a compiled regex the first time you use them: `compile()` makes sure it isn't lost from cache if a lot of other regex is flying around.

So performance implications are going to be quite low-level subtle bytecode stuff.

I think in practice the usual reason to `compile()` is to get something you can call methods from. In this situation the `str` look pretty clean so I think we'll leave them.

oerc0122: > Note The compiled versions of the most recent patterns passed to [re.compile()](https://docs.python.org/3/library/re.html#re.compile) and the module-level matching functions are cached, so programs that use only a few regular expressions at a time needn’t worry about compiling regular expressions.

Huh. Never knew that, always assumed it was better.

ajjackson: castep-outputs probably qualifies as a project with so many regex flying around it makes sense to explictly preserve some of them. (I think it would still usually be under the [512-pattern cache limit](https://zeitbach.com/blog/2024/01/24/python-rematch-caches-compiled-patterns), but maybe when combined with other libraries...)

PR #524, opened 2026-08-05
Another attempt to fix coverage CI

ajjackson:
Github Copilot kindly offered to create a branch for me; it seemed worth a try!

The coverage reporter action has been failing to pick up changed/added files, which makes the report in PR chat pretty useless.

- The main insight is that a missing checkout means the Coverage action lacks context for the coverage information. That seems plausible.
- The secondary suggestion was to make sure we are _definitely_ finding the coverage .xml and not getting the artifact paths wrong. This seems less likely but doesn't hurt to check at the same time.

### DIFF
--- a/.github/workflows/run_tests.yml
+++ b/.github/workflows/run_tests.yml
@@ -119,11 +119,19 @@ jobs:
           path: artifacts
           name: 'Coverage XML (ubuntu-latest)'
 
+      - name: Checkout repository
+        uses: actions/checkout@v6
+        with:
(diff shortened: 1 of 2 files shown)


ajjackson:
This is still a nightmare. I think the answer is to come up with a _simpler_, _locally-runnable_, workflow to produce a merged coverage.xml with appropriate root path. Preferably orchestrated with `tox`, but some other standard tool would be acceptable.

PR #525, opened 2026-08-10
Coverage workflow N+1

ajjackson:
Yet another attempt to get a reliable low-fuss coverage setup. This time the strategy was to implement as much as possible _locally_ and orchestrate with tox.

- Move most config to pyproject.toml
- Ditch dotnet-based tools so it is easier to reproduce things locally, managing requirements and commands with `tox`
- Use tox markers to define useful test groups; coverage-local will give a quick run with coverage and generate a badge to ensure the XML was propertly readable.
- For coverage runs on ubuntu, we select a marker which substitutes the default py314 environment for a py314-cov environment. We _don't_ need to run coverage on every python version!
- Keeping the coverage contained to one tox env and using the native .coverage data dump seems to make it easier for pytest-cov to merge results. We convert to xml as a separate step.

Unfortunately a bit of Github Actions magic is still needed to deal with external fork permissions. They aren't allowed to have write permissions because that could be abused. Instead, the run_tests workflow creates a blob of coverage information and when it finishes another workflow running in this repo is launched and writes to the PR. We can't test the second part at all until _something_ is on `master`, and we can't fiddle with the "true path" except by pushing things to master. So let's hope it works first time 🤞 

**TODO**
- [x] Restore phonopy_reader extra, my agents keep deleting it because they use uv
- [x] Restore HTML generation

---
**Notes on AI usage**

This work was LLM-assited; I take responsibility for the output.

`qwen3-coder` was used with a lot of iteration and intervention. It feels like it does well with a precise brief but was not so proactive about troubleshooting as some models. If we're not careful to limit the task size it can also waste time on hallucinations!

Results were a bit better using gemini-3.6-flash to generate an openspec and then delegating to qwen3-coder or gemini-3.5-flash-lite for implementation.


.github/workflows/test_checkout_one_os.yml
```diff
@@ -85,39 +85,38 @@ jobs:
...
 
-      - name: Merge coverage reports
+      - name: Debug coverage files
         if: inputs.coverage
-        shell: bash -l {0}
         run: |
-          dotnet tool install -g dotnet-coverage
-          dotnet coverage merge --reports "tests_and_analysis/test/reports/coverage*.xml" -f cobertura -o "tests_and_analysis/test/reports/coverage-combined.xml"
+          ls -la tests_and_analysis/test
+          find . -name ".coverage*"
```
oerc0122: Is this testing info only?

.github/workflows/test_checkout_one_os.yml
```diff
@@ -76,7 +76,7 @@ jobs:
 
       - name: Run tests
         shell: bash -l {0}
-        run: python -m tox run -- ${{ inputs.coverage && '--cov' || '' }} --parallel
+        run: python -m tox run -m ${{ inputs.coverage && 'ci-coverage' || 'ci-base' }} -- --parallel
```
oerc0122, line 79: Just so I follow, this is tox running `ci-coverage` iff `inputs.coverage`?

tests_and_analysis/test/run_tests.py
```diff
@@ -148,12 +147,12 @@ def run_tests(
...
-        coveragerc_filepath: Path = test_dir / '.coveragerc'
-
         pytest_options = [
-            '--cov',
-            f'--cov-report=xml:{reports_dir}/coverage_{uuid4()}.xml',
-            f'--cov-config={coveragerc_filepath}',
+            '--cov=euphonic',
+            '--cov-config=../../pyproject.toml',
+            '--cov-append',
+            '--cov-branch',
```
oerc0122, line 154: This is doing the work of the dotnet stuff previously?

CONTRIBUTING.rst
```diff
@@ -29,6 +29,36 @@ e.g. to run ``tox`` for available Python versions with parallel
...
 
+Coverage Testing
+~~~~~~~~~~~~~~~~
+
+Test coverage is checked automatically for Github pull requests.
+To examine coverage locally, there are two options:
+
+1. A ``tox`` marker has been created for this purpose: it creates a
+   coverage badge and cleans up temporary files, but relies on a
+   specific Python version. Run with::
```
oerc0122, line 40: I realise this would need updating with the toxfile, is it worth noting which?

pyproject.toml
```diff
@@ -222,3 +222,31 @@ conflicts = [
...
+source = ["euphonic"]
+parallel = true
+data_file = ".coverage"
+relative_files = true
+
+[tool.coverage.paths]
+source = [
+    "euphonic",
+    "*/site-packages/euphonic",
+    "*\\site-packages\\euphonic",
```
oerc0122: Why are some unix style, others windows, does coverage/toml not respect and convert these?

.github/workflows/test_checkout_one_os.yml
ajjackson: Hah, I see git has mashed together one thing being deleted with another thing being added. Yes, this is a debug step that could be deleted now.

.github/workflows/test_checkout_one_os.yml
ajjackson: Yes. Note also that it's a "marker" (i.e. multiple environments) selected with `-m` rather than an environment selected with `-e`.

tests_and_analysis/test/run_tests.py
ajjackson: `--cov-append` replaces the dotnet coverage merging. A key trick is that it works nicely for .coverage files but not so much for coverage.xml . It also helps that we now only run coverage on one Python version, so a single env is running pytest twice in succession and we don't need to merge with files from other envs.

`--cov-branch` adds a calculation of branch coverage, I think this was missing before.

CONTRIBUTING.rst
ajjackson: That's exactly why I didn't mention it. People should use the persistent marker name, and can infer from the error message if they were missing the right python.

pyproject.toml
ajjackson: I'm not 100% confident it does, and don't have a convenient windows environment to check with.

ajjackson: I guess I could temporarily enable coverage with Windows instead of ubuntu in the test matrix to see if it works without this?

ajjackson: Testing now

.github/workflows/test_checkout_one_os.yml
ajjackson: Done

oerc0122:
Looks like it worked except the final workflow stage pointed at the wrong stage.

ajjackson:
Hmm, the coverage printed while testing is good but the generated HTML is not:

image

ajjackson:
Now trying with the windows path restored and final step pointing at windows artifact.

oerc0122:
> Hmm, the coverage printed while testing is good but the generated HTML is not:
> 
> img

At least your 0 line file got 100% coverage

ajjackson:
That's odd, it now produces a good text artifact from Windows but the HTML is still broken 🤔

ajjackson:
Ok, as things seem to be slightly broken for Windows CI coverage (which we don't use) and look fine for the actual pytest run on Windows (which someone might need) regardless of that line... Let's drop the Windows-normalised pyproject line, switch back to Ubuntu and the real-world cases all seem fine.

ajjackson:
Thanks for review @oerc0122 , I will try auto-merge now and see what happens when it hits master!

PR #533, opened 2026-09-03
Release v2.1.0

ajjackson:


ajjackson:
Despite rebase, github failed to place these commits with the release tag onto the `master` linear history. This seems to be related to timestamps. It _might_ not happen if I close the release PR on same day it is made?

Fixed with force-push but I don't like doing that.

PR #535, opened 2026-09-15
Fix failing tests: Pint unicode change

ajjackson:
Pint has starting using a slightly different unicode "dot" in its formatting, and this is breaking a brille script test that checks for an exact plot label string. Here we automatically convert these to the old style so that either version will pass the unit test.


ajjackson:
@oerc0122 See what you make of this match/case 😉 

It's not 100% robust against every possible test scenario: I tried some things and then decided YAGNI/KSS as the test content shouldn't change much.

